PR #9, opened 2016-12-01
Reserve a bit for RVE ABI in e_flags

kito-cheng:
I think it's time to reserve e_flags for RVE even we don't implement yet.


sorear:
I agree

PR #10, opened 2016-12-03
First draft of calling convention

aswaterman:


riscv-elf.md
```diff
@@ -42,6 +42,106 @@
...
+The base integer calling convention provides eight argument registers,
+a0-a7, the first two of which are also used to return values.
+
+Scalars that are 2✕XLEN bits wide are passed in an aligned pair of argument
+registers.  Wider scalars are passed by reference and are replaced in the
+argument list with the address.
+
+Aggregates whose total size is no more than XLEN bits are passed in
+a register, with the fields laid out as though they were passed in memory.
+Aggregates whose total size is no more than 2✕XLEN bits are passed in a pair
```
sorear: Is it correct that scalars always take an aligned pair and aggregates are never aligned, regardless of types?  What about a struct containing a single long long?  (I'd like the rule that structs with exactly one field after flattening are passed as if the field itself were passed to be general, not just for floats.)

riscv-elf.md
```diff
@@ -42,6 +42,106 @@
     object is taken to use the soft-float ABI.
   * EF_RISCV_RVE (0x0008): This bit is set when the binary targets the E ABI.
 
+# Register Map
+
+Integer Registers
+--------------------------------
+Name    | ABI Mnemonic | Meaning
+--------|--------------|--------
+x0      | x0           | Zero
```
sorear: mnemonic is "zero"

riscv-elf.md
```diff
@@ -42,6 +42,106 @@
...
+
+## Hardware floating-point calling convention
+
+For the purposes of this section, FLEN refers to the width of a
+floating-point register in the ABI.  The ISA might have wider
+floating-point registers than than ABI.
+
+For the purposes of this section, "struct" refers to a C struct with
+its hierarchy flattened.  That is, struct { struct { float f; } g; int i; }
+and struct { float f; int i; } are equivalent.
```
sorear: Maybe clarify that this applies to arrays as well?

riscv-elf.md
```diff
@@ -42,6 +42,106 @@
...
+f10-f17 | fa0-fa7      | Argument registers
+f18-f27 | fs2-fs11     | Callee-saved registers
+f28-f31 | ft8-ft11     | Temporary registers
+
+# Procedure Calling Convention
+
+The base integer calling convention provides eight argument registers,
+a0-a7, the first two of which are also used to return values.
+
+Scalars that are 2✕XLEN bits wide are passed in an aligned pair of argument
```
sorear: The base case for ≤XLEN scalars appears to be missing, as is any discussion of sign extension.  Is `struct { short; }` sign extended as a result of the one-element struct rule?  What about `struct { float; short; }` ?

riscv-elf.md
```diff
@@ -42,6 +42,106 @@
...
+registers.  Wider scalars are passed by reference and are replaced in the
+argument list with the address.
```
sorear: Also those with (C++ only) nontrivial copy constructors, destructors, or vtables.

riscv-elf.md
sorear: Need a clause for putting arguments on the stack.

riscv-elf.md
```diff
@@ -42,6 +42,106 @@
...
+
+Aggregates whose total size is no more than XLEN bits are passed in
+a register, with the fields laid out as though they were passed in memory.
+Aggregates whose total size is no more than 2✕XLEN bits are passed in a pair
+of registers; if only one register is available, the first half is passed in
+a7 and the second half is passed on the stack.  Aggregates larger than 2✕XLEN
+bits are passed by reference and are replaced in the argument list with the
```
sorear: Is the called function permitted to modify the referenced memory?

riscv-elf.md
aswaterman: ack

riscv-elf.md
aswaterman: ack.  re: struct { short; } etc., see one-element struct rule discussion below.

riscv-elf.md
aswaterman: Aggregates should follow the doubleword alignment rule like scalars.  Will fix.

The rule that one-element structs are passed just like their constituent type seems nice, but I can't figure out how to implement it.  Are you aware of other ABIs that impose this rule?

sorear: Not that I'm directly aware of.  The primary effect of the rule concerns sign-extension; if we can find a good enough description of the sign-extension behavior of `struct { short; }` etc that should be enough?

aswaterman: I'm specifically concerned about entering uncharted territory with GCC etc... all other properties of this proposal, AFAIK, are shared by at least one other ABI.

riscv-elf.md
aswaterman: Seems best to follow the majority on this one, and give the memory to the callee.

riscv-elf.md
aswaterman: For now, I'm writing that if aggregates don't completely fill a register, the MSBs are undefined.  We can revisit this.

PR #14, opened 2016-12-18
Add relocation and thread local storage tables for review

michaeljclark:
- Add initial RISC-V ELF relocations table
- Add initial RISC-V TLS (thread local storage) table


aswaterman:
Thanks.  We should also add the numeric values, and flesh out the descriptions, but this is a helpful start.

michaeljclark:
Yes. It's rather condensed. It could likely be split into two tables (one with the symbol values) or one table and a descriptive sections with a subhead for each relocation. A prompt to think about it... Cheers

aswaterman:
I'd keep the tables together if the formatting is sensible.

I'm also starting to wonder if we should move this into LaTeX before it metastasizes, but this is again an @sorear question.

sorear:
AsciiDoc or reST would also be options; IIRC O'Reilly has used asciidoc for actual books.  What problem are we trying to solve again?

aswaterman:
Not trying to solve an immediate problem.  Markdown is sufficient for a sketch of this document.  But every time I've embarked on this kind of effort beginning with something like Markdown or Google Docs, I've ultimately had to switch to TeX for more sophistication, and regretted the effort to port the document.

I'm not familiar with reST or AsciiDoc.  Whatever you think is best (including sticking with MD) is fine with me--just putting in my 2c.

michaeljclark:
Agree. Markdown is sufficient for information collection.

I like the idea of having TeX as an output. We could potentially find a Markdown or AsciiDoc to TeX converter in the documentation tool pipeline.

http://tex.stackexchange.com/questions/101717/converting-markdown-to-latex-in-latex

There is mention of [Pandoc](http://pandoc.org/) which is a Haskell tool. There is also [Sphinx](http://www.sphinx-doc.org/en/1.5.1/) which is used by the LLVM project. [PGF TikZ](http://www.texample.net/media/pgf/builds/pgfmanualCVS2012-11-04.pdf) is quite powerful for generating diagrams.

michaeljclark:
Currently checking the instruction types for the various relocations. When I prepared the table my understanding was less than it is now. %hi(symbol) should only be usable with lui?

```
R_RISCV_PCREL_HI20   | PC-relative reference      | %pcrel_hi(symbol)   | U-Type (auipc)
R_RISCV_HI20         | Absolute address           | %hi(symbol)         | U-Type (lui,auipc)
```

should be:

```
R_RISCV_PCREL_HI20   | PC-relative reference      | %pcrel_hi(symbol)   | U-Type (auipc)
R_RISCV_HI20         | Absolute address           | %hi(symbol)         | U-Type (lui)
```

sorear:
It's only intended for use with lui, but we should define it in a way which can be unambiguously interpreted for any instruction, because undefined behavior in a linker is bad.

michaeljclark:
Yap. Any of the %xxx_hi functions technically should work on any U-Type instructions due to the common immediate encoding. There are separate relocation symbols for the %xxx_lo functions depending on whether the function is used on an I-Type or S-Type instruction (`LO12_I` for addi, loads, and `LO12_S` for stores). The assembler is selective as to which functions can be used on which instructions.

This is what happens if one tries to use a U-Type function on an I-Type instruction:

```
foo.S:13: Error: illegal operands `addi t0,t0,%hi(msg)'
```

%lo also does not work on any I-Type or S-Type instruction, only addi, addi, loads and stores.

```
src/test/hello-world-abs.S:13: Error: illegal operands `slli t0,t0,%lo(msg)'
```

The functions appear to be restricted only at the Instruction-type level which makes sense as the immediate coding is shared for all instructions of one type, and this is why we have separate relocation symbols for `LO12_I` (addi, loads) and `LO12_S` (store).

michaeljclark:
The %xxx_hi functions could both be labeled (lui,auipc) as the assembler accepts them on either however I can't see the use for an absolute on auipc and PC-relative of lui. I think how it is in 0cf3bbfbba5397c59cdc0aeafb670c22221f4fd4 (need to check the remainder, and some are lacking details) ...

sorear:
%hi: All I'm saying is that the linker's job is to push bits around, not to worry if they make sense afterward.  Assembler can care if it wants though.  (And we should document the assembler rules.)  BTW you can edit comments here if you don't need to preserve the exact timeline.

michaeljclark:
I guess the linker could overwrite bits if it has the wrong reloc for an instruction but if the linker is performing relaxation (target specific optimizations) then it could potentially verify. The linker relaxation code must be ISA-aware to the degree that is required to turn far jumps into near jumps, etc. We should document linker relaxations somewhere too.

michaeljclark:
Is the linker relaxation presently just AUIPC/JALR into JAL?

I am trying to figure out how %tprel_add(symbol) is used and why it is needed in addition to %tprel_lo(symbol). I gave it a vague description "internal assembler expansion". It seems to be used on the add instruction for TLS local exec. I should make the description I-Type (addi)

michaeljclark:
Found %tprel_add(symbol) reference and, yes, it is specific to the add instruction: https://raw.githubusercontent.com/riscv/riscv-gcc/riscv-gcc-6.1.0/gcc/config/riscv/riscv.md

michaeljclark:
Are auipc,(load or store) relaxed into GP-Relative in the linker?

sorear:
auipc+load could be relaxed but isn't currently.  lui+load, lui+store, and lui+addi are relaxed to gp-relative.

michaeljclark:
Guess that means we don't get gp-relative for PIC code.

sorear:
Not now, but we will, you just need to add it.  Makes a great deal of sense for PIEs.

aswaterman:
Yep.  My plan is to support it for PIE (but not shared libraries, to avoid gp save/restore complexity).

michaeljclark:
I have been thinking about how to explain R_RISCV_PCREL_HI20, and R_RISCV_PCREL_LO12_I or R_RISCV_PCREL_LO12_S pairs.

The R_RISCV_PCREL_HI20 relocation associates a symbol with an AUIPC instruction where the high 20-bits are relocated to point to a symbol relative to the PC of the instruction.

The AUIPC instruction and relocation is followed by an I-Type instruction (add immediate, load) and R_RISCV_PCREL_LO12_I relocation or S-Type instruction (store) and R_RISCV_PCREL_LO12_S relocation.

The R_RISCV_PCREL_LO12_I or R_RISCV_PCREL_LO12_S relocations contains a (typically private) local label symbol that points to the AUIPC instruction which has the R_RISCV_PCREL_HI20 relocation entry that then points to the actual symbol. e.g. LO12 reloc -> label -> labeled instruction HI20 reloc -> symbol

To get the symbol address to perform the calculation to fill the 12-bit immediate on the add, load or store instruction we have to find the R_RISCV_PCREL_HI20 relocation entry associated with the AUIPC instruction offset pointed to by the R_RISCV_PCREL_LO12_I or R_RISCV_PCREL_LO12_S symbol.
 
As expected, if there is no %pcrel_hi(msg) on the instruction labelled in the LO12 relocation, e.g. %pcrel_lo(1b), the linker fails, as it fails on the second part of the double indirect lookup.

```
(.text+0x8): relocation truncated to fit: R_RISCV_PCREL_LO12_I against `.L11'
```

In any case I learned the mechanism, as I've been hammering out a quick and dirty assembler to test a JIT API and this relocation mechanism became clear to me today:

  https://hastebin.com/uhuquhefoh.bash

aswaterman:
Do you want me to merge this now?

michaeljclark:
I think the data in the table is okay for information collection purposes but not great for layout. I'm having a look at how other ELF PS ABI docs have described their relocations.

I want to add the symbol value into the left hand column and perhaps change the right hand column to a simpler description and move some of the text (assembler macros and instructions) into a section following the table describing the relocations in more detail. Perhaps the address calculation can be in one column.

I need to objdump some of the thread local relocations so I can explain their address calculation. I've figured out the address calculations for most of the typical relocations bar thread local storage.

Value | Name                | Description
:-----  | :---------------           | :-------------
26 | R_RISCV_HI20         | Absolute address (HI20 U-Type instruction)
27 | R_RISCV_LO12_I       | Absolute address (LO12 I-Type instruction)
28 | R_RISCV_LO12_S       | Absolute address (LO12 S-Type instruction)

It's fine to merge as we go but if you give me a day or two I can make some more updates (since I've figured more out regarding relocs since I compiled the table)...

aswaterman:
Cool, no rush, just ping me after your next update.

michaeljclark:
I'm also testing the assembler functions that I somehow associated with the relocs in the table.

```
$ cat a.s 
.section .text
.globl _start
_start:
        lbu a0, %gprel(msg)(gp)

.section .rodata
_gp:
msg:
        .string "Hello World\n"
$ riscv64-unknown-elf-as a.s 
a.s: Assembler messages:
a.s:4: Error: bad expression
a.s:4: Error: bad expression
a.s:4: Error: illegal operands `lbu a0,%gprel(msg)(gp)'
```

Is GP-Relative only emitted during relaxation in the program linkage stage? i.e. is never emitted by the compiler backend? The expression looks valid. `%gprel(symbol)` appears on other arches.

aswaterman:
Linker only at this point.

michaeljclark:
OK

michaeljclark:
I'm looking at `R_RISCV_CALL` and `R_RISCV_CALL_PLT` and learning how call relaxation works.

Is it correct to state that these relocs apply to an instruction sequence instead of one instruction? (an exception to the rule as most RISC-V relocations typically apply to one instruction).

The assembler is emitting a two instruction sequence (`auipc+jalr`) but there is only one reloc, either `R_RISCV_CALL` (static) or `R_RISCV_CALL_PLT` (dynamic PIC) and it is attached to the `auipc` instruction.

In other cases, the two instruction sequences have two relocs e.g. (`auipc+load` or `auipc+store`) (`PCREL_HI20` plus `PCREL_LO12_I` or `PCREL_LO12_S`).

I've actually implemented this reloc as updating "two instructions" in my test assembler (no relaxation logic), and this seems to be an exception to the rule of one reloc per updated instruction:

https://github.com/michaeljclark/riscv-meta/blob/945c6f59b04c374ba70130c27f36d22bc1555826/src/asm/assembler.cc#L398-L419

It seems that `ld` link time optimises the (`auipc`+`jalr`) to a single (`jal`) to the PLT entry.

Example:

```
$ cat a.c 
#include <stdio.h>

int main()
{
	printf("Hello\n");
}
```

objdump of the intermediate object:

```
$ riscv64-unknown-elf-objdump -x -d a.o

a.o:     file format elf64-littleriscv
a.o
architecture: riscv, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x0000000000000000

Sections:
Idx Name          Size      VMA               LMA               File off  Algn
  0 .text         00000038  0000000000000000  0000000000000000  00000040  2**2
                  CONTENTS, ALLOC, LOAD, RELOC, READONLY, CODE
  1 .data         00000000  0000000000000000  0000000000000000  00000078  2**0
                  CONTENTS, ALLOC, LOAD, DATA
  2 .bss          00000000  0000000000000000  0000000000000000  00000078  2**0
                  ALLOC
  3 .rodata       00000006  0000000000000000  0000000000000000  00000078  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  4 .comment      00000012  0000000000000000  0000000000000000  0000007e  2**0
                  CONTENTS, READONLY
SYMBOL TABLE:
0000000000000000 l    df *ABS*	0000000000000000 a.c
0000000000000000 l    d  .text	0000000000000000 .text
0000000000000000 l    d  .data	0000000000000000 .data
0000000000000000 l    d  .bss	0000000000000000 .bss
0000000000000000 l    d  .rodata	0000000000000000 .rodata
0000000000000000 l       .rodata	0000000000000000 .LC0
0000000000000010 l       .text	0000000000000000 .L0 
0000000000000000 l    d  .comment	0000000000000000 .comment
0000000000000000 g     F .text	0000000000000038 main
0000000000000000         *UND*	0000000000000000 puts



Disassembly of section .text:

0000000000000000 <main>:
   0:	ff010113          	addi	sp,sp,-16
   4:	00113423          	sd	ra,8(sp)
   8:	00813023          	sd	s0,0(sp)
   c:	01010413          	addi	s0,sp,16

0000000000000010 <.L0 >:
  10:	00000517          	auipc	a0,0x0
			10: R_RISCV_PCREL_HI20	.LC0
  14:	00050513          	mv	a0,a0
			14: R_RISCV_PCREL_LO12_I	.L0 
  18:	00000317          	auipc	t1,0x0
			18: R_RISCV_CALL_PLT	puts
			18: R_RISCV_RELAX	*ABS*
  1c:	000300e7          	jalr	t1
  20:	00000793          	li	a5,0
  24:	00078513          	mv	a0,a5
  28:	00813083          	ld	ra,8(sp)
  2c:	00013403          	ld	s0,0(sp)
  30:	01010113          	addi	sp,sp,16
  34:	00008067          	ret
```

link dynamically

```
riscv64-unknown-linux-gnu-gcc  -fPIC -c a.c -o a.o
riscv64-unknown-linux-gnu-gcc  a.o -o a
riscv64-unknown-elf-objdump -x -d a
```

R_RISCV_CALL_PLT (`auipc`+`jalr`) gets rewritten as (`jal`):

```
0000000000010480 <main>:
   10480:	ff010113          	addi	sp,sp,-16
   10484:	00113423          	sd	ra,8(sp)
   10488:	00813023          	sd	s0,0(sp)
   1048c:	01010413          	addi	s0,sp,16
   10490:	00000517          	auipc	a0,0x0
   10494:	0b850513          	addi	a0,a0,184 # 10548 <__libc_csu_fini+0x4>
   10498:	eb9ff0ef          	jal	ra,10350 <puts@plt>
   1049c:	00000793          	li	a5,0
   104a0:	00078513          	mv	a0,a5
   104a4:	00813083          	ld	ra,8(sp)
   104a8:	00013403          	ld	s0,0(sp)
   104ac:	01010113          	addi	sp,sp,16
   104b0:	00008067          	ret
```

In any case. It seems like `R_RISCV_CALL` and `R_RISCV_CALL_PLT` are not simple relocs that apply just to the instruction they label.

aswaterman:
In general, the relocs apply to bit sequences, rather than instructions.  The CALL relocs operate on the 64 bits they point to, not just the instruction.  They are not the only exceptions, though; COPY, ADD*, SUB*, and JUMP_SLOT generally don't refer to instructions at all.

michaeljclark:
Now it makes sense why it was defined as a 64-bit reloc

michaeljclark:
It's hard to describe some of the relocations in an abstract sense (as bit sequences) when some of the relocations are instruction type specific such as the pairs emitted by `la`. Below is my attempt at describing the label indirection required for the `R_RISCV_PCREL_HI20 + R_RISCV_PCREL_LO12_{I,S}` relocs. Is this readable? _(PS: the x86 PS ABI doc has some very wordy sections describing the relocations)_.

### Position Independent Code

PC-relative relocations for symbol addresses (such as those emitted by the `la` pseudo-instruction) in position independent code are typically comprised of a pair of relocations: `R_RISCV_PCREL_HI20` plus `R_RISCV_PCREL_LO12_I` or `R_RISCV_PCREL_LO12_S`.

The `R_RISCV_PCREL_HI20` relocation refers to an `AUIPC` instruction containing the high 20-bits to be relocated to a symbol relative to the program counter address of the `AUIPC` instruction. The `AUIPC` instruction is followed by an I-Type instruction (add immediate or load) with an `R_RISCV_PCREL_LO12_I` relocation or an S-Type instruction (store) and an `R_RISCV_PCREL_LO12_S` relocation. The successive instruction has a signed 12-bit immediate so the value of the preceding `AUIPC` instruction may have 1 added to it.

The `R_RISCV_PCREL_LO12_I` or `R_RISCV_PCREL_LO12_S` relocations contain a label pointing to the `AUIPC` instruction with the `R_RISCV_PCREL_HI20` relocation entry that then points to the target symbol:

 - `R_RISCV_PCREL_LO12_I` relocation entry ⟶ `label`
 - `label` ⟶ `AUIPC` instruction with `R_RISCV_PCREL_HI20` reloc
 - `AUIPC` instruction `R_RISCV_PCREL_HI20` reloc ⟶ `symbol`

To get the symbol address to perform the calculation to fill the 12-bit immediate on the add, load or store instruction the linker finds the `R_RISCV_PCREL_HI20` relocation entry associated with the AUIPC instruction. The addresses for pair of relocations are calculated like this:

 - `hi20 = ((symbol_address - hi20_reloc_offset + 0x800) >> 12) << 12;`
 - `lo12 = symbol_address - hi20_reloc_offset - hi20;`

michaeljclark:
We could substitute instruction with bit sequence. The `auipc+(add,load,store)` relocations are tricky ones to explain.

aswaterman:
The PCREL_LO ones still act on bit sequences--it's just that the symbol they refer to is indirect.  I'd remove the AUIPC instruction from the description; the only requirement is that there is a PCREL_HI20 reloc at the PCREL_LO target.

 - At `label`: `R_RISCV_PCREL_HI20` reloc -> `symbol`
 - `R_RISCV_PCREL_LO12_I` relocation entry ⟶ `label`

michaeljclark:
OK

michaeljclark:
I've changed the text to use the term "bit sequence" for relocation but have made (parenthetical) reference to the instruction pairs that generate said "bit sequences". I have kept the instruction type in the text as it is suffixed in the relocation name. It's useful to document the relationship between the relocation and the instructions they are generated for even if the linker only deals with bit sequences. I've also removed the right shift (<< 12). This should be a small improvement...

### Position Independent Code

PC-relative relocations on bit sequences for symbol addresses (such as the bit sequence emitted by the `auipc+addi` instruction pair expanded from the `la` pseudo-instruction) in position independent code are typically comprised of a pair of relocations: `R_RISCV_PCREL_HI20` plus `R_RISCV_PCREL_LO12_I` or `R_RISCV_PCREL_LO12_S`.

The `R_RISCV_PCREL_HI20` relocation refers to a bit sequence where the high 20-bits are relocated to the symbol relative to the program counter. The `R_RISCV_PCREL_HI20` relocation is generally followed by an I-Type instruction (add immediate or load) with an `R_RISCV_PCREL_LO12_I` relocation or an S-Type instruction (store) and an `R_RISCV_PCREL_LO12_S` relocation. 

The `R_RISCV_PCREL_LO12_I` or `R_RISCV_PCREL_LO12_S` relocations contain a label pointing to a bit-sequence with a `R_RISCV_PCREL_HI20` relocation entry that points to the target symbol:

 - At label: `R_RISCV_PCREL_HI20` relocation entry ⟶ symbol
 - `R_RISCV_PCREL_LO12_I` relocation entry ⟶ label

To get the symbol address to perform the calculation to fill the 12-bit immediate, the linker finds the symbol in the `R_RISCV_PCREL_HI20` relocation entry at the label. The addresses for pair of relocations are calculated like this:

 - `hi20 = ((symbol_address - hi20_reloc_offset + 0x800) >> 12);`
 - `lo12 = symbol_address - hi20_reloc_offset - hi20;`

The successive bit sequence has a signed 12-bit immediate so the value of the preceding high 20-bit relocation may have 1 added to it.

michaeljclark:
I might leave that text as is for now and move on to Procedure Linkage Table and GOT...

aswaterman:
Mmm, the text really does make more sense when you write "instruction" instead of "bit sequence".

How about, at the start of the section on relocs, mention something like:  While relocations can be made on arbitrary memory locations, many relocations are designed for use with specific instructions or instruction sequences.  For clarity, the description of those relocations assumes they are used in the intended context.

Then, keep using "instruction" in the text.

michaeljclark:
OK. I have the previous version... It is dependent on the particular relocation type. Some are more instruction specific than others...

aswaterman:
Right.  Whatever you think gives the clearest description.

michaeljclark:
The `R_RISCV_PCREL_HI20` relocation can (at present) only ever be on an `AUIPC` instruction. It makes sense to refer to the `AUIPC` instruction as the `auipc+<add|load|store>` instruction pair address calculation seems to come up on the mailing list occasionally which is why I wanted to spend the time to try to describe it clearly.

I'll spin another version... probably later tonight or tomorrow...

aswaterman:
Makes sense.

michaeljclark:
Thinking about appropriate subtitles to group descriptions of the various relocations. Added empty titles to describe findings on `R_RISCV_CALL` and `R_RISCV_CALL_PLT` that we were discussing earlier ... can add Absolute Addresses (for `R_RISCV_{HI20,LO12_{I,S}}`) and Thread Local Storage as titles, etc. We can't use the relocs themselves as titles as there are logical groupings of associated relocations.

# Relocations

...

While the linker can make relocations on arbitrary memory locations, many
relocations are designed for use with specific instructions or instruction
sequences. For clarity, the description of those relocations assumes they
are used in the intended context.

## Position Independent Code

### PC-Relative Procedure Calls

...

### PC-Relative Jumps and Branches

...

### PC-Relative Symbol Addresses

PC-relative relocations for symbol addresses on pairs of instructions
(such as the `auipc+addi` instruction pair expanded from the `la`
pseudo-instruction) in position independent code typically have an
associated pair of relocations: `R_RISCV_PCREL_HI20` plus
`R_RISCV_PCREL_LO12_I` or `R_RISCV_PCREL_LO12_S`.

The `R_RISCV_PCREL_HI20` relocation refers to an `AUIPC` instruction
containing the high 20-bits to be relocated to a symbol relative to the
program counter address of the `AUIPC` instruction. The `AUIPC`
instruction is followed by an I-Type instruction (add immediate or load)
with an `R_RISCV_PCREL_LO12_I` relocation or an S-Type instruction (store)
and an `R_RISCV_PCREL_LO12_S` relocation.

The `R_RISCV_PCREL_LO12_I` or `R_RISCV_PCREL_LO12_S` relocations contain
a label pointing to an instruction with a `R_RISCV_PCREL_HI20` relocation
entry that points to the target symbol:

 - At label: `R_RISCV_PCREL_HI20` relocation entry ⟶ symbol
 - `R_RISCV_PCREL_LO12_I` relocation entry ⟶ label

To get the symbol address to perform the calculation to fill the 12-bit
immediate on the add, load or store instruction the linker finds the
`R_RISCV_PCREL_HI20` relocation entry associated with the `AUIPC`
instruction. The addresses for pair of relocations are calculated like this:

 - `hi20 = ((symbol_address - hi20_reloc_offset + 0x800) >> 12);`
 - `lo12 = symbol_address - hi20_reloc_offset - hi20;`

The successive instruction has a signed 12-bit immediate so the value of the
preceding high 20-bit relocation may have 1 added to it.

aswaterman:
Subdividing them makes sense to me.

sorear:
Might be good to point out that there's no relocation that truly works on an `auipc+load/store/add` 64-bit pair, the 32-bit relocations are independent and a compiler can order them freely.  This is legal:

```c
extern int foo;
foo += 0x1000;
```

&dArr;

```asm
1: auipc t0, %pcrel_hi(foo)
   lui t1, 1
   lw t2, t0, %pcrel_lo(foo)(1b)
   add t2, t2, t1
   sw t2, t0, %pcrel_lo(foo)(1b)
```

aswaterman:
Right, CALL and CALL_PLT are the only ones that work on contiguous 64 bits.

michaeljclark:
Yes. I'll add a sentence to the section on "PC-Relative Symbol Addresses" to mention that the compiler emitted instructions are not necessarily sequential and the constraint is that the labeled instruction in the RISCV_R_PCREL_LO12_{I,S} relocation has a R_RISCV_PCREL_HI20 relocation pointing to the symbol.

Do you think I should include asm fragments in the docs on the relocations?

aswaterman:
As in, an additional column in the table with an asm example?

michaeljclark:
The table is quite packed. I'm in the text section on "PC-Relative Symbol Addresses". I'll commit an example and you can take a look...

michaeljclark:
### PC-Relative Symbol Addresses

32-bit PC-relative relocations for symbol addresses on pairs of instructions
such as the `AUIPC+ADDI` instruction pair expanded from the `la`
pseudo-instruction, in position independent code typically have an
associated pair of relocations: `R_RISCV_PCREL_HI20` plus
`R_RISCV_PCREL_LO12_I` or `R_RISCV_PCREL_LO12_S`.

The `R_RISCV_PCREL_HI20` relocation refers to an `AUIPC` instruction
containing the high 20-bits to be relocated to a symbol relative to the
program counter address of the `AUIPC` instruction. The `AUIPC`
instruction is followed by an I-Type instruction (add immediate or load)
with an `R_RISCV_PCREL_LO12_I` relocation or an S-Type instruction (store)
and an `R_RISCV_PCREL_LO12_S` relocation.

The `R_RISCV_PCREL_LO12_I` or `R_RISCV_PCREL_LO12_S` relocations contain
a label pointing to an instruction with a `R_RISCV_PCREL_HI20` relocation
entry that points to the target symbol:

 - At label: `R_RISCV_PCREL_HI20` relocation entry ⟶ symbol
 - `R_RISCV_PCREL_LO12_I` relocation entry ⟶ label

To get the symbol address to perform the calculation to fill the 12-bit
immediate on the add, load or store instruction the linker finds the
`R_RISCV_PCREL_HI20` relocation entry associated with the `AUIPC`
instruction. The addresses for pair of relocations are calculated like this:

 - `hi20 = ((symbol_address - hi20_reloc_offset + 0x800) >> 12);`
 - `lo12 = symbol_address - hi20_reloc_offset - hi20;`

The successive instruction has a signed 12-bit immediate so the value of the
preceding high 20-bit relocation may have 1 added to it.

Note the compiler emitted instructions for PC-relative symbol addresses are
not necessarily sequential or in pairs. There is a constraint is that the
instruction with the `RISCV_R_PCREL_LO12_I` or `RISCV_R_PCREL_LO12_S`
relocation label points to a `R_RISCV_PCREL_HI20` relocation pointing to the
symbol. e.g.

```
0000000000000000 <.L11>:
   0: 00000297            auipc t0,0x0
      0: R_RISCV_PCREL_HI20 foo
   4: 0002a383            lw  t2,0(t0)
      4: R_RISCV_PCREL_LO12_I .L11
   8: 006383b3            add t2,t2,t1
   c: 0072a023            sw  t2,0(t0)
      c: R_RISCV_PCREL_LO12_S .L11
```

michaeljclark:
Using objdump format to show the relocations might be helpful.

michaeljclark:
also, I'll change the text to use "sequence of instructions" instead of "pair of instructions".

aswaterman:
I think assembly [pseudo]code will get the point across more clearly, e.g.

```
label:
  auipc t0, %pcrel_hi(sym)      # t0 := label + (sym - label + 0x800)[31:12]
  <unrelated instructions...>
  addi t0, t0, %pcrel_lo(label)    # t0 := t0 + (sym - label)[11:0]
```

michaeljclark:
OK. I think Stefan's load store example is good. I'll use the asm with comments style with the load store as the example.

michaeljclark:
While there are distinct relocations for `R_RISCV_CALL` and `R_RISCV_CALL_PLT` (for which `-fno-pic` and default chooses the former and `-fpie`, `-fpic` and `-fPIC` choose the later), it seems the compiler must treat them interchangeably as if the call is intermodule it won't go via the `PLT`.

I'm wondering what case actually requires the distinction in an "object file" as it seems whether the call goes via the `PLT` is actually decided at program link time and is not dependent on the relocation but where the target symbol is. i.e. if it is in a shared object then it relaxes the sequence to a `JAL` to the `PLT`, but if the symbol is in a module being linked then it relaxes to a direct call.

Is there any known case where the distinction of `R_RISCV_CALL` and `R_RISCV_CALL_PLT` is required? (will play with the linker a bit - I've spent less time with `-shared`)

aswaterman:
R_RISCV_CALL means "if the symbol is defined, call it, else call the PLT."  R_RISCV_CALL_PLT means "call the PLT regardless."

The reason both are necessary is the linker doesn't have enough information to make the correct decision, because it varies by call site, not by symbol.  In some places, the source code may indicate that the symbol binds locally, and in others, it may indicate it doesn't.  This shows up in ld.so and/or libc (forget which).

Note x86 does this, too.

michaeljclark:
No worries. But `R_RISCV_CALL_PLT` should only be relaxed to a `JAL foo@plt` assuming it means "call the PLT regardless" but the linker still makes a direct call. It must be used as a hint or it could have something to do with using the GOT when getting the address of a function. The module must be compiled with `-fPIC` to be used in a shared library, so it affects how the address of a function is taken as this needs to work between modules. Here you can see the linker has turned it into a direct call not via the PLT.

```
$ cat 1.c 
extern void foo();

int main()
{
	foo();
}
$ cat 2.c 
#include <stdio.h>

void foo()
{
	printf("foo\n");
}
$ riscv64-unknown-linux-gnu-gcc -c -fPIC 1.c -o 1.o
$ riscv64-unknown-linux-gnu-gcc -c -fPIC 2.c -o 2.o
$ riscv64-unknown-linux-gnu-gcc 1.o 2.o -o a
$ riscv64-unknown-elf-objdump -x -d 1.o
...
0000000000000000 <main>:
   0:	ff010113          	addi	sp,sp,-16
   4:	00113423          	sd	ra,8(sp)
   8:	00813023          	sd	s0,0(sp)
   c:	01010413          	addi	s0,sp,16
  10:	00000317          	auipc	t1,0x0
			10: R_RISCV_CALL_PLT	foo
			10: R_RISCV_RELAX	*ABS*
  14:	000300e7          	jalr	t1
  18:	00000793          	li	a5,0
  1c:	00078513          	mv	a0,a5
  20:	00813083          	ld	ra,8(sp)
  24:	00013403          	ld	s0,0(sp)
  28:	01010113          	addi	sp,sp,16
  2c:	00008067          	ret
$ riscv64-unknown-elf-objdump -x -d a
...
0000000000010480 <main>:
   10480:	ff010113          	addi	sp,sp,-16
   10484:	00113423          	sd	ra,8(sp)
   10488:	00813023          	sd	s0,0(sp)
   1048c:	01010413          	addi	s0,sp,16
   10490:	01c000ef          	jal	ra,104ac <foo>
   10494:	00000793          	li	a5,0
   10498:	00078513          	mv	a0,a5
   1049c:	00813083          	ld	ra,8(sp)
   104a0:	00013403          	ld	s0,0(sp)
   104a4:	01010113          	addi	sp,sp,16
   104a8:	00008067          	ret

00000000000104ac <foo>:
   104ac:	ff010113          	addi	sp,sp,-16
   104b0:	00113423          	sd	ra,8(sp)
   104b4:	00813023          	sd	s0,0(sp)
   104b8:	01010413          	addi	s0,sp,16
   104bc:	00000517          	auipc	a0,0x0
   104c0:	0b450513          	addi	a0,a0,180 # 10570 <__libc_csu_fini+0x4>
   104c4:	e8dff0ef          	jal	ra,10350 <puts@plt>
   104c8:	00000013          	nop
   104cc:	00813083          	ld	ra,8(sp)
   104d0:	00013403          	ld	s0,0(sp)
   104d4:	01010113          	addi	sp,sp,16
   104d8:	00008067          	ret
...
```

michaeljclark:
On reflection, I think it must be used for taking the address of a function as it needs to return something from the GOT table whether the function was called directly or via the PLT (as the functions address is in the GOT when compiled as shared?)

michaeljclark:
```
$ cat 1.c 
extern void foo();

void (*moo)(void) = &foo;

int main()
{
	moo();
}
$ cat 2.c 
#include <stdio.h>

void foo()
{
	printf("foo\n");
}
$ riscv64-unknown-linux-gnu-gcc -c -fPIC 1.c -o 1.o
$ riscv64-unknown-linux-gnu-gcc -c -fPIC 2.c -o 2.o
$ riscv64-unknown-linux-gnu-gcc 1.o 2.o -o a
```

aswaterman:
Let me try again :)

R_RISCV_CALL means "if the symbol is defined, call it, else call the PLT." R_RISCV_CALL_PLT means "if generating a shared library, call the PLT regardless; otherwise, same as R_RISCV_CALL"

michaeljclark:
Yes. I think the linker can make a direct call but taking the address of the function has to be equal no matter what type of module tries to take the address. It's making a GOT reference for the address instead of taking the address of the function directly. I just objdumped that example.

aswaterman:
Taking the address of a function is a separate issue.  It's OK if CALL uses the non-canonical address, because the program is none the wiser.

In general, if you take the address of a function, it will generate a GOT access.  The GOT entry will contain the real address of the function if no PLT stub exists, or the address of the PLT stub otherwise.

michaeljclark:
Yap. Taking the address of the function always has to return the same result but the linker can choose to make a direct call during program linking.

michaeljclark:
This is the `R_RISCV_PCREL_LO12_I` relocation used in a different manner than I have seen. It slightly invalidates a sentence I made about R_RISCV_PCREL_LO12_{I,S}. The label needs to point to an instruction with a valid "PC-relative relocation". Less specific.

```
Disassembly of section .text:

0000000000000000 <main>:
   0:	ff010113          	addi	sp,sp,-16
   4:	00113423          	sd	ra,8(sp)
   8:	00813023          	sd	s0,0(sp)
   c:	01010413          	addi	s0,sp,16

0000000000000010 <.L0 >:
  10:	00000797          	auipc	a5,0x0
			10: R_RISCV_GOT_HI20	moo
  14:	0007b783          	ld	a5,0(a5) # 10 <.L0 >
			14: R_RISCV_PCREL_LO12_I	.L0 
  18:	0007b783          	ld	a5,0(a5)
  1c:	000780e7          	jalr	a5
  20:	00000793          	li	a5,0
  24:	00078513          	mv	a0,a5
  28:	00813083          	ld	ra,8(sp)
  2c:	00013403          	ld	s0,0(sp)
  30:	01010113          	addi	sp,sp,16
  34:	00008067          	ret
```

aswaterman:
Oh, right.  Forgot about that.  A PCREL_LO must refer to any of the PC-relative HI relocs:

R_RISCV_GOT_HI20
R_RISCV_TLS_GOT_HI20
R_RISCV_TLS_GD_HI20
R_RISCV_PCREL_HI20

michaeljclark:
Looking at thread local storage now and writing up some adhoc notes on my findings...

I had found earlier, that RISC-V appears to implement 3 models at the assembly level although 4 models are exposed via the `-ftls=` compiler flag: (`local-exec`, `initial-exec`, `local-dynamic` and `global-dynamic`). `local-dynamic` and `global-dynamic` generate the same code.

I found this elucidating comment in `riscv-gnu-toolchain/src/newlib-gcc/gcc/config/riscv/riscv.c`:

```
    case TLS_MODEL_LOCAL_DYNAMIC:
      /* Rely on section anchors for the optimization that LDM TLS
         provides.  The anchor's address is loaded with GD TLS. */
    case TLS_MODEL_GLOBAL_DYNAMIC:
```

What I didn't realise is the `%tprel_add(symbol)` does not expand to a value but instead appears to be present to generate a relocation as it is used as a fourth operand to the `add` instruction that adds the tls offset to the `tp` register.

Still need to analyse the TLS pseudo instruction expansions for `la.tls.ie` and `la.tls.gd`; and how TLS relaxations work by running the sample code through the program linker and analysing the output.

In Ulrich's [Thead Local Storage Paper](http://it.bmc.uu.se/andlov/dev/manual/Ulrich%20Drepper%20-%20TLS%20-%20ELF%20Handling%20For%20Thread-Local%20Storage%20(tls.pdf).pdf) he describes two variants names "Varant I" and "Variant II". It appears Variant I TLS dynamic thread vector offsets grows upwards and Variant II offsets grows downwards

> Variant II reminds me of `nosegneg` on i686 Xen as TLS on i686 relied on segment wrap for negative references which conflicted with Xen virtualisation which used segmentation for protecting the Hypervisor (http://blog.ispeakdeutsch.com/2011/02/xen-hwcap-0-nosegneg-and-mno-tls-direct.html).

Is it correct to state that RISC-V is "Variant I"?

Also trying to find the definition of `__tls_get_addr` to see if this is the signature. OK. Found it in `riscv-gnu-toolchain/riscv-glibc/sysdeps/riscv/dl-tls.h`

```
extern void *__tls_get_addr (tls_index *ti);
```

where the type tls index is defined as:

```
typedef struct
{
unsigned long int ti_module;
unsigned long int ti_offset;
} tls_index;
```

FYI sample RISC-V Thread Local Storage assembly output for the 3 models:

```
$ cat 3.c 
extern __thread int i;

int moo()
{
	return i++;
}
```

```
$ riscv64-unknown-linux-gnu-gcc -fPIC -ftls-model=local-exec -c -S 3.c -o 3.s
$ cat 3.s 
	.file	"3.c"
	.option pic
	.text
	.align	2
	.globl	moo
	.type	moo, @function
moo:
	add	sp,sp,-16
	sd	s0,8(sp)
	add	s0,sp,16
	lui	a5,%tprel_hi(i)
	add	a5,a5,tp,%tprel_add(i)
	lw	a5,%tprel_lo(i)(a5)
	addw	a3,a5,1
	lui	a4,%tprel_hi(i)
	add	a4,a4,tp,%tprel_add(i)
	sw	a3,%tprel_lo(i)(a4)
	mv	a0,a5
	ld	s0,8(sp)
	add	sp,sp,16
	jr	ra
	.size	moo, .-moo
	.ident	"GCC: (GNU) 6.1.0"
```

```
$ riscv64-unknown-linux-gnu-gcc -fPIC -ftls-model=initial-exec -c -S 3.c -o 3.s
$ cat 3.s 
	.file	"3.c"
	.option pic
	.text
	.align	2
	.globl	moo
	.type	moo, @function
moo:
	add	sp,sp,-16
	sd	s0,8(sp)
	add	s0,sp,16
	la.tls.ie	a5,i
	add	a5,a5,tp
	lw	a5,0(a5)
	addw	a3,a5,1
	la.tls.ie	a4,i
	add	a4,a4,tp
	sw	a3,0(a4)
	mv	a0,a5
	ld	s0,8(sp)
	add	sp,sp,16
	jr	ra
	.size	moo, .-moo
	.ident	"GCC: (GNU) 6.1.0"
```

```
$ riscv64-unknown-linux-gnu-gcc -fPIC -ftls-model=global-dynamic -c -S 3.c -o 3.s
$ cat 3.s 
	.file	"3.c"
	.option pic
	.text
	.align	2
	.globl	moo
	.type	moo, @function
moo:
	add	sp,sp,-32
	sd	ra,24(sp)
	sd	s0,16(sp)
	sd	s1,8(sp)
	sd	s2,0(sp)
	add	s0,sp,32
	la.tls.gd	a0,i
	call	__tls_get_addr@plt
	mv	a5,a0
	lw	s1,0(a5)
	addw	s2,s1,1
	la.tls.gd	a0,i
	call	__tls_get_addr@plt
	mv	a5,a0
	sw	s2,0(a5)
	mv	a5,s1
	mv	a0,a5
	ld	ra,24(sp)
	ld	s0,16(sp)
	ld	s1,8(sp)
	ld	s2,0(sp)
	add	sp,sp,32
	jr	ra
	.size	moo, .-moo
	.ident	"GCC: (GNU) 6.1.0"
```

aswaterman:
Do you think this is PR is in a mergeable state now?  It's fine to add more/tweak later in a separate PR.

michaeljclark:
Yes I think it can be merged.

michaeljclark:
I just went through trying to make the comments that show the relocation types use a consistent format. It could perhaps have some wider review now.

riscv-elf.md
```diff
@@ -167,10 +167,438 @@ type would be passed.
...
+- Compiler flag `-ftls-model=local-exec`
+- Variable attribute: `__thread int i __attribute__((tls_model("local-exec")));`
+
+Example assembler load and store of a thread local variable `i` using the
+`%tprel_hi`, `%tprel_add` and `%tprel_lo` assembler functions. The emitted
+relocations are in comments.
+
+```
+  lui  a5,%tprel_hi(i)           # R_RISCV_TPREL_HI20 (symbol)
+  add  a5,a5,tp,%tprel_add(i)    # R_RISCV_TPREL_ADD (symbol)
```
sorear: Is this not relaxable?

riscv-elf.md
```diff
@@ -167,10 +167,438 @@ type would be passed.
...
+- ELF flags: DF_STATIC_TLS
+
+Example assembler load and store of a thread local variable `i` using the
+`la.tls.ie` psuedo-instruction, with the emitted TLS relocations in comments:
+
+```
+   la.tls.ie a5,i
+   add  a5,a5,tp
+   lw   t0,0(a5)
+   addi t0,a0,1
```
sorear: t0,t0 I think

michaeljclark: Did I miss a R_RISCV_RELAX relocation?

I can go through and double check objdump output of the TLS examples tomorrow... 

I wanted this level of detail as I was thinking about the LLVM integrated assembler and LLD... and the TLS pseudo instructions are otherwise not documented anywhere else.

michaeljclark: Oh right. t0,t0

michaeljclark: I've fixed the addi.

riscv-elf.md
michaeljclark: I think we need to add an entire section on Linker Relaxations. I've just read through. I think this is good enough for an incremental merge.

michaeljclark: I don't really have much on GP-Relative other than listing it in the table

aswaterman:
Thanks @michaeljclark!!

PR #23, opened 2017-03-14
Provide a clearer FP calling convention example (fixes #22)

### LINKED ISSUE #22: Misleading example in floating point calling convention section
The document currently states: "`struct { struct { float f[1]; } g[2]; int h; }` and `struct { float f; float g; int h; }` are treated the same"

There are multiple possible fixes depending on the point that is being made. Perhaps `{ struct { float f[1]; } float g[1]; int h; }` is what was intended? However regardless of whether you flatten the struct fields, this is (by my reading) always passed identically to the integer calling convention. `struct { struct { float f[1]; } int h[1]; }` would perhaps be a better example?

asb:
As discussed, the example struct introduced in the floating point calling
convention example will be passed as in the integer calling convention, which
isn't ideal. This PR removes a field so it is covered by the rule for a struct
containing two fp reals.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -134,8 +134,8 @@ floating-point register in the ABI.  The ISA might have wider
 floating-point registers than the ABI.
 
 For the purposes of this section, "struct" refers to a C struct with its
-hierarchy flattened, including any array fields.  That is, struct { struct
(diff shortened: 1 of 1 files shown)


aswaterman:
Thanks, Alex.

PR #25, opened 2017-03-15
Clarify rules for passing small aggregates when no registers are available

asb:
I believe the intent is that if no registers are available, then small aggregates are passed on the stack. This would seem consistent with the behaviour for scalars and with the fact that a 2*xlen is passed half on the stack, half in a register if only one register is available.

Is my understanding correct?


asb:
See also #26 regarding the under-specification for the alignment of values passed in this way.

asb:
Ping - if anyone has a moment to review I'd really appreciate it.

aswaterman:
sorry, this slipped through the cracks.

PR #30, opened 2017-03-22
fix typos about branch/jump range and integer register use with floats

DSHorner:
thnx to sorear for noting the more substantial range fix.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -124,10 +124,11 @@ address as an implicit first parameter.
 
 The hardware floating-point calling convention adds eight floating-point
 argument registers, fa0-fa7, the first two of which are also used to return
-values.  Values are passed in floating-point registers whenever possible, even
-if all integer registers have been exhausted.  The remainder of this section
-applies only to named arguments.  Variadic arguments are passed according to
-the integer calling convention.
+values.  Values are passed in floating-point registers whenever possible,
+whether or not the integer registers have been exhausted.
+
+The remainder of this section applies only to named arguments.  Variadic
+arguments are passed according to the integer calling convention.
 
 For the purposes of this section, FLEN refers to the width of a
 floating-point register in the ABI.  The ISA might have wider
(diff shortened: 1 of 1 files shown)


riscv-elf.md
```diff
@@ -125,7 +125,7 @@ address as an implicit first parameter.
 The hardware floating-point calling convention adds eight floating-point
 argument registers, fa0-fa7, the first two of which are also used to return
 values.  Values are passed in floating-point registers whenever possible, even
-if all integer registers have been exhausted.  The remainder of this section
+if all integer registers have not been exhausted.  The remainder of this section
```
aswaterman: As I mentioned in the other comment, this change destroys the intended meaning.  To capture both of our points, might I suggest:

> whether or not the integer registers have been exhausted

DSHorner: agreed. much better than what I originally considered - "irrespective"

As usual I'm struggling with github and how to revise pull request, but I intend to work it through for this one, but it will be later today.

aswaterman: If you push additional commits to your typos1 branch, they should show up in this pull request.  (It's worth suffering though figuring it out, as github and RISC-V seem inextricably tied...)

DSHorner:
There is a revised script.

riscv-elf.md
```diff
@@ -138,6 +139,8 @@ hierarchy flattened, including any array fields.  That is, struct `{ struct
 { float f[1]; } g[2]; }` and `struct { float f; float g; }` are
 treated the same.  Empty structs are ignored, even in C++, unless they
 have nontrivial copy constructors or destructors.
+A struct containing just one floating-point real is passed as though it were
```
DSHorner: This got picked up somehow.

aswaterman:
Merged manually https://github.com/riscv/riscv-elf-psabi-doc/commit/1c69ccf4098f46109341ce39153c61638841afaf

PR #35, opened 2017-07-19
Fix PLT header assembly code

PkmX:
`auipc` should be paired with a `%pcrel_lo` for `addi`, not absolute addressing `%lo`.


aswaterman:
Thanks, this looks like an editing error in the comments.

It looks like it's wrong in https://github.com/riscv/riscv-binutils-gdb/blob/riscv-next/bfd/elfnn-riscv.c, too.

PR #39, opened 2017-09-04
Document that char is unsigned

asb:
This pull request builds on #38.


palmer-dabbelt:
I merged this manually.

PR #40, opened 2017-09-07
State stack alignment requirements

aswaterman:


asb:
Thus looks good to me from the perspective that it accurately and clearly describes the status quo for rv64, and describes a sensible alternative to the current rv32 stack alignment convention. Probably need more feedback on the rv32 change before merging.

aswaterman:
I've modified the PR to reflect the present requirement (128 bits for RV32I and RV64I, or 32 bits for RV32E), and I'm going to merge it.  The debate about RV32I's stack alignment can continue separately.

PR #41, opened 2017-09-09
Describe callee-saved scheme

aswaterman:
If you can think of a clearer way to explain the floating-point scheme, let me know...

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -146,6 +146,9 @@ The stack pointer shall be aligned to a 128-bit boundary upon procedure
 entry, except for the RV32E ABI, where it need only be aligned to 32 bits.
 The stack pointer need not remain aligned throughout procedure execution.
 
+Registers s0-s11 shall be preserved across procedure calls.
+No floating-point registers, if present, are preserved across calls.
+
 ## Hardware floating-point calling convention
 
 The hardware floating-point calling convention adds eight floating-point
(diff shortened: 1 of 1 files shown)


riscv-elf.md
```diff
@@ -193,6 +196,14 @@ Otherwise, it is passed according to the integer calling convention.
...
 type would be passed.
 
+Floating-point registers fs0-fs11 shall be preserved across procedure calls,
+provided they hold values no more than FLEN bits wide.  If the floating-point
+registers are wider than FLEN bits, a callee may either preserve the entire
+register, or preserve the lower FLEN bits and fill the remaining bits with 1s.
+(Thus, it suffices to spill registers with a floating-point store of at
+least FLEN bits and to refill the register with a floating-point load of
+the same width.)
+
```
asb: I'm either misunderstanding, or this paragraph isn't consistent. The second sentence seems to imply that when compiling for FLEN (ABI FP register size), you might expect the compiler to generate a wider-than-FP store. Isn't the most obvious option for an FLEN-sized store to be generated? As I understand it, we're talking about a situation like -march=rv32imafd -mabi=rv32if or -march=rv64imafdq -mabi=lp64d

aswaterman: Yeah, it's just written too confusingly. I meant to say that using an at-least-FLEN-bit store is required. However, requiring an exactly FLEN-bit store be used implies that the MSBs must be destroyed, which clearly isn't the goal. I'll rewrite it later today.

PR #44, opened 2017-09-09
Document alignment of _Complex

### LINKED ISSUE #28: Size of standard C types should be described
There is a table in the v2.1 spec which should be moved here. It might also be worth expanding to include a mapping of `_Complex` and `_Bool`.

The table in the v2.1 specification lists long double as 16 bytes in RV32. Is that correct given that RV32IMFDQ is specifically disallowed in the spec? Or is the idea that it's a handy way to access 128-bit fp implemented via software emulation routines?

aswaterman:
Closes #28

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -243,6 +243,11 @@ default ABIs:
 
 `char` is unsigned.
 
+`_Complex` types have the alignment and layout of a struct containing two
+fields of the corresponding real type (`float`, `double`, or `long double`),
+with the first field holding the real part and the second field holding
+the imaginary part.
+
 A future version of this specification may define an ILP32 ABI for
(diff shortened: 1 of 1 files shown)


asb:
The alignment and size of _Complex is specified in C99 (http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf p46). There's some value in saving people effort of double-checking whether _Complex needs special handling - if we're doing that, we might as well clarify the field order too.

How about "`_Complex` types have the alignment and layout of a struct containing two fields of the corresponding real type (`float`, `double`, or `long double`). The first field contain. the real part and the second field contains the imaginary part. e.g. `struct { float re; float imm; };`?

aswaterman:
Ack

PR #47, opened 2017-09-27
Add RVX bit in elf flag

kito-cheng:


kito-cheng:
Yeah, standard extensions have no flags to indicate, that I think it's an issue too. For floating point, we have ABI to identify that, but for other other extension which don't effect ABI it seems a problem, user will got illegal instruction exception when execute if there is some extension not supported in target, I hope there is some way to check before execute, e.g. check that in dynamic linker or kernel when load elf?

Or maybe we could create section to record that? and check that during link time, given warning when linking different arch?

aswaterman:
I see.  I think we need a separate section for ISA string (e.g. rv64imafdcxandes), because saying only "X" is not enough information.  You need to know which X it is, or else you can't determine if an Xandes processor can run an Xsifive binary.

Similarly, if we eventually have objdump support for NSE, the separate section can tell it how to disassemble.

kito-cheng:
Yes, we are now using the form like `rv64imafdcxandes`, you are right that not enough info for single elf flag bit.

So, I just go create section and write concept-of-proof on binutils, and then write detail into this spec.

aswaterman:
sgtm

PR #48, opened 2017-09-27
Document __int128 alignment on RV64

asb:
This documents the current behaviour of GCC. Although not part of the C
standard, the __int128/__uint128 types are mentioned in both the AArch64
and x86-64 ABI docs.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -225,6 +225,7 @@ default ABIs:
     int         |  4            |  4
     long        |  8            |  8
     long long   |  8            |  8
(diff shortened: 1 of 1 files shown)


aswaterman:
I believe the current behavior of GCC is that __int128 is unsupported on RV32.  But this definition would make sense if it were supported.

asb:
Yes, __int128 is not supported on RV32 in gcc. This patch adds the entry only for RV64.

aswaterman:
Sorry, I misread the patch.

PR #49, opened 2017-09-28
Explicitly state which registers are preserved across calls

aswaterman:


asb:
Are you leaving out gp and tp because you consider them unallocatable?

aswaterman:
Yeah, describing them as preserved across calls isn't sufficient.  Signal handlers rely on them, so they shouldn't be modified at all.

Can you think of a pithy way to express that in this table?

asb:
Maybe just `-- (Unallocatable)` or `-- (Reserved)`. Then a comment below saying something like "In the standard ABI, the tp and gp registers should not be considered eligible for spilling - signal handlers may rely upon their value.`

aswaterman:
Thanks - addressed.

PR #56, opened 2017-11-07
Factor RV32E calling convention into separate section

aswaterman:
I also am proposing that x16-x31 be reserved for interrupt handlers when running RV32E ABI code on RV32I.  I realize this is different than the convention for floating-point registers when using an integer-only calling convention, but I think it will be more useful in these systems than treating x16-x31 as temporaries.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -7,6 +7,7 @@
 2. [Procedure Calling Convention](#procedure-calling-convention)
 	* [Integer Calling Convention](#integer-calling-convention)
 	* [Hardware Floating-point Calling Convention](#hardware-floating-point-calling-convention)
+	* [RV32E Calling Convention](#rv32e-calling-convention)
 	* [Default ABIs and C type sizes](#default-abis-and-c-type-sizes)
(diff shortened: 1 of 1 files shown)


asb:
I can understand that people using the RV32E ABI on an RV32I target might want to use the extra registers for ISRs, but why don't people targeting those systems just compile with `-march=rv32e -mabi=ilp32e'? If people are confident that their runtime environment will allow them to use x16-x31 as temporaries, they can compile with `-march=rv32i -mabi=ilp32e'. With the proposed behaviour, we might have to dream up another way of using the RV32E ABI on an RV32I target and using x16-x31 as temporaries.

It seems advantageous to keep uniformity here. In most cases the appropriate flags will be set by the user's toolchain anyway, which weakens the argument for trying to guess common use-cases and match them with default behaviour.

aswaterman:
That's a good point.

aswaterman:
I'm going to punt this by requiring the RV32E calling convention be used with the RV32E ISA, which we can relax later.

PR #57, opened 2017-11-08
Define wchar_t to be signed

palmer-dabbelt:
Our char is unsigned, but our wchar_t is unspecified.  GCC currently
emits a signed wchar_t, but I don't think that makes any sense.  Other
ports define their Linux ABI to have unsigned wchar_t in GCC, which I
think we should do as well.  I have a patch prepared for GCC.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -217,6 +217,7 @@ default ABIs:
     char        |  1            |  1
     short       |  2            |  2
     int         |  4            |  4
+    wchar_t     |  4            |  4
     long        |  8            |  8
     long long   |  8            |  8
     __int128    | 16            | 16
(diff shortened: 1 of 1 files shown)


palmer-dabbelt:
@michaeljclark informs me that musl will have to change <https://github.com/rv8-io/musl-riscv/commit/ec91b72cf15c040f441ccbf0ddf5ccd30797ea97>.  I haven't checked glibc yet.

Neither is upstream, so I think we're safe.

aswaterman:
(Also, if wchar is signed, then all unsigned chars and signed chars are representable with the wchar type.  Not sure how important this property is, but it's certainly not a bad thing.)

michaeljclark:
Perhaps it is best to keep wchar_t as a signed int then.

It would be a little painful to change as libgcc has to be in sync with the C library headers. Noticed this when building musl and comparing with aarch64 alltypes.h.in which has unsigned wchar_t.

aswaterman:
OK, let's keep it as-is

palmer-dabbelt:
@aswaterman I flipped the polarity of the PR

PR #58, opened 2017-11-12
Document signedness, size, and alignment of wint_t

asb:
As per current RISC-V GCC, wint_t is unsigned (unlike wchar_t which is signed). This seems a little odd, but isn't unprecedented. It seems {x86_64,i386}-linux has the same, as does {ppc,ppc64}-linux.


aswaterman:
It is odd, but given x86-64 is doing the same thing, I'm not worried about it causing any gratuitous porting headaches.

Issue #29, opened 2017-03-19
Vararg details are underspecified


asb:
At the bare minimum we should specify the `va_list` struct. The document should probably go much further and give a detailed run-down, much like the AArch64 and x86-64 psABI docs.

PR #60, opened 2017-11-15
Start to document the details of vararg handling

asb:
See also issue #29. I expect the description in this PR is not yet clear or complete enough - feedback very welcome.


aswaterman:
Thanks, it's a good start

PR #61, opened 2017-11-15
Clarify that the vararg aligned register pair rule may leave a7 unused

asb:
I think it's worth being explicit here, even though doing anything else would be impossible (as `va_list` couldn't work).


aswaterman:
agreed

PR #63, opened 2017-11-30
Add EF_RISCV_TSO flag

aswaterman:


asb:
Also looks good to me.

asb:
This does raise a question about versioning. RVWMO/RVTSO seems to have consensus, but hasn't really been adopted yet. In this case it probably makes sense to go ahead and assign the bit. In general, we might need to differentiate between "pending" assignments that are awaiting the relevant spec to be ratified.

aswaterman:
We generally can't wait for ratification, but I agree we shouldn't speculatively assign bits, either.

PR #65, opened 2017-12-12
Nonstandard relocs

### LINKED ISSUE #64: Customized relocation type
In our current CPU design, we extend RISC-V ISA by adding several instructions, such as GP-implied load/store. To handle these instructions, we also extend relocation type. However, there does not have any rule or guideline to add customized relocation type currently. To avid conflict, in our current design, the new relocation id is assigned from 255 and in decreasing order. Could anyone have any idea or propose to deal with this issue?

aswaterman:
Resolves #64

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -407,6 +407,8 @@ Enum | ELF Reloc Type       | Description                     | Details
 57-191  | *Reserved*        | Reserved for future standard use |
 192-255 | *Reserved*        | Reserved for nonstandard ABI extensions |
 
+Nonstandard extensions are free to use relocation numbers 192-255 for any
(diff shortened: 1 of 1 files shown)


asb:
Is the intention that the 192-255 space is a free-for-all, or that non-standard relocation numbers will be managed in some way? Or is this patch purposely avoiding making such a determination?

aswaterman:
My plan was the former: chaos, just like the ISA opcode space for nonstandard extensions.  I guess I should add a sentence to that effect?

asb:
Yes, chaos was my assumption but just leaving it marked as "reserved" without further explanation might lead someone to think there's something more managed going on. An extra sentence would be a good idea I think.

aswaterman:
Yep.

PR #67, opened 2018-01-12
Document that unions are always passed according to the integer calling convention

### LINKED ISSUE #24: FP calling convention: no mention of unions in struct flattening description
Consider the following declarations:

    struct s1_ty { float f; int i;  };
    struct s2_ty { float f; char c;  };
    union u_ty { struct s1_ty s1; struct s2_ty s2; };
    struct s3_ty { union u_ty u; };

As it stands, the FP calling convention makes no mention of unions. So if I had a function taking `u_ty` as an argument I might reasonably pass this according to the integer calling convention. If a function took `s3_ty`, there's a question of whether a union can ever be "flattened". For instance in this particular case, the first field of the union will always be a float. Presumably even if you did "flatten" the fp field, you'd still consider the second field of the flattened result to be a union of `int` and `char`, which would be an "aggregate" rather than an "integer". The degenerate case would be a union of two identical structs, containing a real and an integer.

I assume the intention is that unions are never "flattened"?

asb:
Closes #24.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -31,7 +31,7 @@ This RISC-V ELF psABI specification document is
  &copy; 2016 Kito Cheng <kito.cheng@gmail.com>
  &copy; 2016-2017 Andrew Waterman <aswaterman@gmail.com>
  &copy; 2016-2017 Michael Clark <michaeljclark@mac.com>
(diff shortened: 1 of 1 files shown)


aswaterman:
thanks

PR #68, opened 2018-01-12
Empty structs or unions are always ignored

asb:
The document previously documented that empty structs are ignored, but this note was only present under the floating point calling convention. Empty structs are ignored for all calling conventions, so document it under the base integer calling convention.

Additionally, document that empty unions are ignored.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -107,6 +107,10 @@ Aggregates larger than 2✕XLEN bits are passed by reference and are replaced in
 the argument list with the address, as are C++ aggregates with nontrivial copy
 constructors, destructors, or vtables.
 
+Empty structs or union arguments or return values are ignored by C compilers
+which support them as a non-standard extension.  This is not the case for C++,
+which requires them to be sized types.
+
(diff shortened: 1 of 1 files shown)


aswaterman:
I only meant that empty structs are ignored in figuring out how to pass floating-point structs in registers.  In the general case, empty structs are _not_ ignored for C++, where they're considered to have size 1 rather than size 0.  Consider this code:

    struct s {};
    int foo(struct s a, int c) { return c; }

If compiled as C code, ```c``` is passed in ```a0```.  If compiled as C++ code, ```c``` is passed in ```a1```.  (ARM and AArch64 work the same way.)

asb:
That's a good point - I was only considering C. I've pushed an updated version of the patch which

1) is explicit about empty unions and structs in C (yes it's a non-standard extension, but it's supported by clang and gcc and we might as well clarify it)
2) clarifies that the discussion of 'ignoring' empty structs or unions in the floating point calling convention section is referring to struct flattening

aswaterman:
Thanks, looks good to me.

PR #69, opened 2018-01-13
Describe bitfield layout

### LINKED ISSUE #20: Bit fields are unspecified
The AArch64 AAPCS and x86-64 psABI both make an effort to standardise bit field representation. I don't know how much code is out there that exposes bit fields in its public ABI, but it may be worth specifying something sensible here.

aswaterman:
Hopefully resolves #20.


aswaterman:
OK, I'll merge it as-is, but let me know if you later come up with a clearer way to phrase it.

PR #71, opened 2018-03-27
Document ELF attribute section

kito-cheng:
Hi all:

It's the draft of the ELF attribute.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -1007,14 +1007,14 @@ can hold an integer encoded in the uleb128 format, a null-terminated byte
 string (NTBS), or a combination of an integer and a string.
 
 ### List of attributes
-Tag                    | Value | Parameter type | Description
-:--------------------- | :---- | :------------- | :---------------------
-Tag_RISCV_arch               |     4 | NTBS           | Indicates the target architecture of this object.
-Tag_RISCV_priv_spec          |     5 | uleb128        | Indicates the major version of the privileged specification.
+Tag                          | Value | Parameter type | Description
+:--------------------------- | :---- | :------------- | :---------------------
+Tag_RISCV_priv_spec          |     4 | uleb128        | Indicates the major version of the privileged specification.
+Tag_RISCV_arch               |     5 | NTBS           | Indicates the target architecture of this object.
 Tag_RISCV_priv_spec_minor    |     6 | uleb128        | Indicates the minor version of the privileged specification.
-Tag_RISCV_priv_spec_revision |     7 | uleb128        | Indicates the revision version of the privileged specification.
-Tag_RISCV_unaligned_access   |     8 | uleb128        | Indicates whether to impose unaligned memory accesses in code generation.
-Tag_RISCV_stack_align        |     9 | uleb128        | Indicates the stack alignment requirement in bytes.
+Tag_RISCV_priv_spec_revision |     8 | uleb128        | Indicates the revision version of the privileged specification.
+Tag_RISCV_unaligned_access   |    10 | uleb128        | Indicates whether to impose unaligned memory accesses in code generation.
+Tag_RISCV_stack_align        |    12 | uleb128        | Indicates the stack alignment requirement in bytes.
 
 ### Detailed attribute description
 
@@ -1023,7 +1023,7 @@ Tag_RISCV_stack_align        |     9 | uleb128        | Indicates the stack alig
 Each attribute is described in the following structure:
 ```<Tag name>, <Value>, <Parameter type 1>=<Parameter name 1>[, <Parameter type 2>=<Parameter name 2>]```
 
-#### Tag_RISCV_arch, 4, NTBS=subarch
+#### Tag_RISCV_arch, 5, NTBS=subarch
 Tag_RISCV_arch contains the string about the target architecture from the option
 `-march`. Different architectures will be integrated into a superset when object
 files are merged.
@@ -1034,25 +1034,25 @@ information, if not given by `-march`, must be accordance with the default
 specified by the tool. For example, the architecture “RV32I” had to be recorded
 in the attribute as `RV32I2P0` in which `2P0` stands for the default version of
 its based ISA. On the other hand, the architecture `RV32G` has to be presented
-as `RV32I2P0M2P0A2P0F2P0D2P0` in which the abbreviation `G` is expanded
+as `RV32I2P0_M2P0_A2P0_F2P0_D2P0` in which the abbreviation `G` is expanded
 to the IMAFD combination with default versions of the standard extensions.
 
 
-#### Tag_RISCV_priv_spec, 5, uleb128=version
+#### Tag_RISCV_priv_spec, 4, uleb128=version
 #### Tag_RISCV_priv_spec_minor, 6, uleb128=version
-#### Tag_RISCV_priv_spec_revision, 7, uleb128=version
+#### Tag_RISCV_priv_spec_revision, 8, uleb128=version
 
 Tag_RISCV_priv_spec contains the major/minor/revision version information about
 the privileged specification. It will report errors if object files of different
 privileged specification versions are merged.
 
-#### Tag_RISCV_unaligned_access, 8, uleb128=value
+#### Tag_RISCV_unaligned_access, 10, uleb128=value
 Tag_RISCV_unaligned_access denotes the code generation policy for this object file.
 Its values are defined as follows:
  - 0: This object does not allow any unaligned memory accesses.
  - 1: This object allows unaligned memory accesses.
 
-#### Tag_RISCV_stack_align, 9, uleb128=value
+#### Tag_RISCV_stack_align, 12, uleb128=value
 Tag_RISCV_strict_align records the N-byte stack alignment for this object. The default
 value is 16 for RV32I or RV64I, and 4 for RV32E.
 
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -995,9 +995,9 @@ The following table lists the special sections defined by this ABI.
 
 Name                       | Type                 | Attributes
 :------------------------- | :------------------- | :-------------
-.riscv.attributes.         | SHT_RISCV_ATTRIBUTES | none
+.riscv.attributes          | SHT_RISCV_ATTRIBUTES | none
 
-.riscv.attributes. names a section that contains RISC-V ELF attributes.
+.riscv.attributes names a section that contains RISC-V ELF attributes.
 
 ## <a name=program-header-table></a>Program Header Table
 
@@ -1017,22 +1017,25 @@ There are no RISC-V specific definitions relating to ELF hash tables.
 
 ## <a name=Attributes></a>Attributes
 
-Attributes are used to record the information about an object file/binary that a
-linker or runtime loader needs to check the compatibility.
-Attributes are encoded in an vendor-specific section of type
-SHT_RISCV_ATTRIBUTES and name .riscv.attributes. The value of an attribute
-can hold an integer encoded in the uleb128 format, a null-terminated byte
-string (NTBS), or a combination of an integer and a string.
+Attributes are used to record information about an object file/binary that a
+linker or runtime loader needs to check compatibility.
+
+Attributes are encoded in a vendor-specific section of type SHT_RISCV_ATTRIBUTES
+and name .riscv.attributes. The value of an attribute can hold an integer
+encoded in the uleb128 format or a null-terminated byte string (NTBS).
+
+RISC-V attributes have a string value if the tag number is odd and an integer
+value if the tag number is even.
 
 ### List of attributes
-Tag                          | Value | Parameter type | Description
-:--------------------------- | :---- | :------------- | :---------------------
-Tag_RISCV_stack_align        |     4 | uleb128        | Indicates the stack alignment requirement in bytes.
-Tag_RISCV_arch               |     5 | NTBS           | Indicates the target architecture of this object.
-Tag_RISCV_unaligned_access   |     6 | uleb128        | Indicates whether to impose unaligned memory accesses in code generation.
-Tag_RISCV_priv_spec          |     8 | uleb128        | Indicates the major version of the privileged specification.
-Tag_RISCV_priv_spec_minor    |    10 | uleb128        | Indicates the minor version of the privileged specification.
-Tag_RISCV_priv_spec_revision |    12 | uleb128        | Indicates the revision version of the privileged specification.
+Tag                                 | Value    | Parameter type | Description
+:---------------------------------- | :------- | :------------- | :---------------------
+Tag_RISCV_stack_align               |        4 | uleb128        | Indicates the stack alignment requirement in bytes.
+Tag_RISCV_arch                      |        5 | NTBS           | Indicates the target architecture of this object.
+Tag_RISCV_unaligned_access          |        6 | uleb128        | Indicates whether to impose unaligned memory accesses in code generation.
+Tag_RISCV_priv_spec                 |        8 | uleb128        | Indicates the major version of the privileged specification.
+Tag_RISCV_priv_spec_minor           |       10 | uleb128        | Indicates the minor version of the privileged specification.
+Tag_RISCV_priv_spec_revision        |       12 | uleb128        | Indicates the revision version of the privileged specification.
 
 ### Detailed attribute description
 
@@ -1042,29 +1045,29 @@ Each attribute is described in the following structure:
 ```<Tag name>, <Value>, <Parameter type 1>=<Parameter name 1>[, <Parameter type 2>=<Parameter name 2>]```
 
 #### Tag_RISCV_stack_align, 4, uleb128=value
-Tag_RISCV_strict_align records the N-byte stack alignment for this object. The default
-value is 16 for RV32I or RV64I, and 4 for RV32E.
+Tag_RISCV_strict_align records the N-byte stack alignment for this object. The
+default value is 16 for RV32I or RV64I, and 4 for RV32E.
 
 The smallest value will be used if object files with different Tag_RISCV_stack_align
 values are merged.
 
 #### Tag_RISCV_arch, 5, NTBS=subarch
-Tag_RISCV_arch contains the string about the target architecture from the option
-`-march`. Different architectures will be integrated into a superset when object
-files are merged.
+Tag_RISCV_arch contains a string for the target architecture taken from
+the option `-march`. Different architectures will be integrated into a superset
+when object files are merged.
 
-Note that the version information of target architecture must be presented
-explicitly in the attribute and abbreviation must be expanded. The version
-information, if not given by `-march`, must be accordance with the default
-specified by the tool. For example, the architecture “RV32I” had to be recorded
+Note that the version information for target architecture must be presented
+explicitly in the attribute and abbreviations must be expanded. The version
+information, if not given by `-march`, must agree with the default
+specified by the tool. For example, the architecture `RV32I` has to be recorded
 in the attribute as `RV32I2P0` in which `2P0` stands for the default version of
 its based ISA. On the other hand, the architecture `RV32G` has to be presented
 as `RV32I2P0_M2P0_A2P0_F2P0_D2P0` in which the abbreviation `G` is expanded
 to the IMAFD combination with default versions of the standard extensions.
 
 #### Tag_RISCV_unaligned_access, 6, uleb128=value
-Tag_RISCV_unaligned_access denotes the code generation policy for this object file.
-Its values are defined as follows:
+Tag_RISCV_unaligned_access denotes the code generation policy for this object
+file. Its values are defined as follows:
  - 0: This object does not allow any unaligned memory accesses.
  - 1: This object allows unaligned memory accesses.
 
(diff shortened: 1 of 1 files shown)


riscv-elf.md
```diff
@@ -788,3 +789,105 @@ typedef struct
...
+specification. It will report errors if object files of different privileged
+specification versions are merged.
+
+The value of version information is computed by the following formula:
+```
+Major version * 10000 + Minor version * 100 + Revision version
+```
+
+For example, the value for version 2.1.3 of the privileged specification will be
+20103.
```
palmer-dabbelt: Is it a problem to encode these as separate fields, so there isn't a problem with running out of space?  I don't want to end up in a "100 is enough for anyone" situation in 20 years :).

riscv-elf.md
```diff
@@ -788,3 +789,105 @@ typedef struct
...
+
+Each attribute is described in the following structure:
+```<Tag name>, <Value>, <Parameter type 1>=<Parameter name 1>[, <Parameter type 2>=<Parameter name 2>]```
+
+#### Tag_arch, 4, NTBS=subarch
+Tag_arch contains the raw string about the target architecture from the option
+`-march`. Different architectures will be integrated into a superset when object
+files are merged.
+
+#### Tag_priv_spec, 5, uleb128=version
```
palmer-dabbelt: Why is the priv ISA version specified in the ELF but not the user ISA?  The user ISA seems like the more important one.

riscv-elf.md
```diff
@@ -788,3 +789,105 @@ typedef struct
...
+
+
+### Detailed attribute description
+
+#### How does this specification describe public attributes?
+
+Each attribute is described in the following structure:
+```<Tag name>, <Value>, <Parameter type 1>=<Parameter name 1>[, <Parameter type 2>=<Parameter name 2>]```
+
+#### Tag_arch, 4, NTBS=subarch
```
palmer-dabbelt: If we're including "-march" then we should include "-mabi" as well.

riscv-elf.md
```diff
@@ -788,3 +789,105 @@ typedef struct
...
+ - sub-architecture: The format denoting sub-architecture is undefined
+   now. The string should be left empty.
+
+#### Tag_X_ext, 88, NTBS=arch
+
+This attribute indicates if this object contains any non-standard extensions.
+The arch string denotes the name of a non-standard extension with a prefix of
+x. If there are multiple non-standard extensions in the string, each extension
+is separated from another with an underline. For example, when there are two
+non-standard extensions foo and bar, the string is encoded as `xfoo_xbar`.
```
palmer-dabbelt: @aswaterman IIRC there was some ambiguous wording in the ISA spec about this, did it ever get fixed?

riscv-elf.md
```diff
@@ -788,3 +789,105 @@ typedef struct
...
+and sub-architecture. Instructions of the x extension are only permitted in an
+object/binary if Tag_x _ ext is presented in the attributes section.
+
+The attributes of standard extensions are all encoded in the same format that
+consists of the following two parts:
+
+ - Version information
+   The value of version information is computed by the following formula:
+
+   ```Major version * 10000 + Minor version * 100 + Revision version```
```
palmer-dabbelt: The same versioning question, though it's probably less important for these.

riscv-elf.md
aswaterman: Yes, and this description is correct.

riscv-elf.md
kito-cheng: Hmm, you are right, maximal number of attribute is almost unlimited, we should separate that.

riscv-elf.md
kito-cheng: I thought version info of user ISA should separated in different `Tag_*_ext` entry? Does here any info not included in a standard extension or base integer instruction?

riscv-elf.md
kito-cheng: `-mabi` are all included in `e_flags`, so I guess we don't need repeat again in attribute?

riscv-elf.md
kito-cheng: BTW, I am not mean every extension should using only one entry, I expect there would be more than one entry for `V` extension to recording type and vector length, but the version info using same rule for `V` extension just like all other extension.

asb:
For anyone who is only following the discussion here, note that there is additional discussion on the sw-dev mailing list:
https://groups.google.com/a/groups.riscv.org/d/msgid/sw-dev/CA%2ByXCZCr%3DGPJWYHcc-NMrfYpy4CfQNu8WCMjUwiRVHwLsGEnnQ%40mail.gmail.com

I think it would be worth considering just emitting a ISA string of the form specified in Chapter 22 of the v2.2 user spec. That section only anticipates specs having major and minor version numbers (i.e. no 'revision version').

palmer-dabbelt:
I agree with @asb here: there's already a standard for naming RISC-V extensions so let's just use that.

kito-cheng:
Changes:
- Update the description of Tag_arch: Need to expand abbreviation and version explicitly.
- Update the define of Tag_priv_spec: it's record major version of priv spec only now.
- Add new attributes: Tag_priv_spec_minor and Tag_priv_spec_revision.
- Remove Tag_*_ext entries.

kito-cheng:
Changes:
- Rebase to master
- Change `Tag_strict_align` to `Tag_unaligned_access`
  - It also getting opposite meaning, because value 0 also mean not present in implementation.

kito-cheng:
Changes:
- Add new section: `Section Types and Special Sections`.
- Update Tag number to fit even/odd rule.
- Change prefix to Tag_RISCV_

aswaterman:
@jim-wilson is the version in FSF Binutils consistent with this pull request?  If so, I'm gonna click the merge button.

jim-wilson:
I haven't double checked that.  I can do that tomorrow.  I do see some changes related to my binutils patch review, so it is likely to be up to date.

asb:
I think RISC-V needs a standard like this for specifying the target ISA using ELF attributes and I'm incredibly grateful for Kito's work on it. However, I'm unhappy at the idea that this be blessed as a standard based on a binutils decision to merge patches. That's not the way standards development should work in an ecosystem with so many interested parties.

I think thought needs to be given to attributes that have scope narrower than an object (as supported by ARM BuildAttributes). This is likely to be a common requirements if you consider e.g. a library with multiple versions of the same function compiled for different extensions. With appropriately scoped attributes, objdump can correctly disassemble these functions even if there was an overlap in encodings.

As commented [in March](https://github.com/riscv/riscv-elf-psabi-doc/pull/71#issuecomment-376559831), official specs don't seem to have a "revision" just major+minor (so maybe Tag_RISCV_priv_spec_revision should be removed). If it's meant to indicate draft/non-draft, some description would be useful. It's a weakness of the ISA naming string description that there's not a clear way to specify the version of the privileged spec and no names are defined to specify e.g. a system that supports U+M privilege levels. I think it should be fixed there, and we should use the ISA naming string throughout.

riscv-elf.md
```diff
@@ -777,6 +778,24 @@ typedef struct
...
+:-------------------- | :---------- | :-------------
+SHT_RISCV_ATTRIBUTES  | 0x70000003  | none
+
+### Special Sections
+
+The following table lists the special sections defined by this ABI.
+
+Name                       | Type                 | Attributes
+:------------------------- | :------------------- | :-------------
+.riscv.attributes.         | SHT_RISCV_ATTRIBUTES | none
```
jim-wilson: The section name is ".riscv.attributes" not ".riscv.attributes.".

riscv-elf.md
```diff
@@ -777,6 +778,24 @@ typedef struct
...
+
+.riscv.attributes. names a section that contains RISC-V ELF attributes.
```
jim-wilson: Likewise ".riscv.attributes" not ".riscv.attributes.".

riscv-elf.md
```diff
@@ -785,3 +804,68 @@ typedef struct
 
 ## <a name=hash-table></a>Hash Table
 
+## <a name=Attributes></a>Attributes
+
+
+Attributes are used to record the information about an object file/binary that a
```
jim-wilson: "the information" -> "information"

riscv-elf.md
```diff
@@ -785,3 +804,68 @@ typedef struct
...
+linker or runtime loader needs to check the compatibility.
```
jim-wilson: "the compatibility" -> "compatibility"

riscv-elf.md
```diff
@@ -785,3 +804,68 @@ typedef struct
...
+Attributes are encoded in an vendor-specific section of type
```
jim-wilson: "an vendor" -> "a vendor"

riscv-elf.md
```diff
@@ -785,3 +804,68 @@ typedef struct
...
+SHT_RISCV_ATTRIBUTES and name .riscv.attributes. The value of an attribute
+can hold an integer encoded in the uleb128 format, a null-terminated byte
+string (NTBS), or a combination of an integer and a string.
```
jim-wilson: Should mention that even numbered attributes are assumed to be uleb128, and odd numbered ones are assumed to be NTBS.  This allows for some forward/backward compatibility.

riscv-elf.md
```diff
@@ -785,3 +804,68 @@ typedef struct
...
+
+#### Tag_RISCV_stack_align, 4, uleb128=value
+Tag_RISCV_strict_align records the N-byte stack alignment for this object. The default
+value is 16 for RV32I or RV64I, and 4 for RV32E.
+
+The smallest value will be used if object files with different Tag_RISCV_stack_align
+values are merged.
+
+#### Tag_RISCV_arch, 5, NTBS=subarch
+Tag_RISCV_arch contains the string about the target architecture from the option
```
jim-wilson: Tag_RISCV_arch contains a string for the target architecture taken from the option `-march`.

riscv-elf.md
```diff
@@ -785,3 +804,68 @@ typedef struct
...
+`-march`. Different architectures will be integrated into a superset when object
+files are merged.
+
+Note that the version information of target architecture must be presented
```
jim-wilson: "of" -> "for"

riscv-elf.md
```diff
@@ -785,3 +804,68 @@ typedef struct
...
+explicitly in the attribute and abbreviation must be expanded. The version
```
jim-wilson: "abbreviation" -> "abbreviations"

riscv-elf.md
```diff
@@ -785,3 +804,68 @@ typedef struct
...
+information, if not given by `-march`, must be accordance with the default
```
jim-wilson: "be accordance" -> "be in accordance", or maybe just "agree"

riscv-elf.md
```diff
@@ -785,3 +804,68 @@ typedef struct
...
+specified by the tool. For example, the architecture “RV32I” had to be recorded
```
jim-wilson: "had" -> "has"

riscv-elf.md
```diff
@@ -785,3 +804,68 @@ typedef struct
...
+as `RV32I2P0_M2P0_A2P0_F2P0_D2P0` in which the abbreviation `G` is expanded
+to the IMAFD combination with default versions of the standard extensions.
+
+#### Tag_RISCV_unaligned_access, 6, uleb128=value
+Tag_RISCV_unaligned_access denotes the code generation policy for this object file.
+Its values are defined as follows:
+ - 0: This object does not allow any unaligned memory accesses.
+ - 1: This object allows unaligned memory accesses.
+
+
```
jim-wilson: extra blank lines here

riscv-elf.md
```diff
@@ -785,3 +804,68 @@ typedef struct
...
+
+#### Tag_RISCV_priv_spec, 8, uleb128=version
+#### Tag_RISCV_priv_spec_minor, 10, uleb128=version
+#### Tag_RISCV_priv_spec_revision, 12, uleb128=version
+
+Tag_RISCV_priv_spec contains the major/minor/revision version information about
```
jim-wilson: "about" -> "of"

jim-wilson:
priv spec 1.9.1 is not compatible with priv spec 1.9, so we do need 3 numbers to precisely describe a priv spec version.  Though hopefully this mistake won't be made again, and 1.9.1 will be the only priv spec version with 3 numbers.

jim-wilson:
On the plus side here, the binutils patches have a lot of flexibility.  You can specify at configure time whether attribute support is on by default or off by default.  If you don't use the configure option, then it is on by default for embedded elf targets and off by default for linux which presumably doesn't need them.  There is also an assembler option that you can use to force attribute support on or off at assembler run time.

The even/odd numbering scheme for tags gives us some forward/backward compatibility, and we have the option of adding new attributes and deprecating old ones.

On the flip side, Kito opened this pull request almost 10 months ago, and attributes were discussed in email messages even before that.  There was no comment on the proposal until I accepted a binutils patch.  As a gnu toolchain developer, when I have a problem that needs to be solved, I can't wait years for a committee to make a decision before I fix a problem.  I'm happy to work with the committee and adjust the support as necessary, but if there isn't going to be any committee decision at all then I will make my own decisions.

Another consideration here is that the RISC-V Foundation doesn't have any programming staff.  It is nice to ask for interesting features to be added to the standard, but as a practical matter, we can only add stuff to the tools that a volunteer is willing to implement.

kito-cheng:
> I think thought needs to be given to attributes that have scope narrower than an object (as supported by > ARM BuildAttributes). This is likely to be a common requirements if you consider e.g. a library with
> multiple versions of the same function compiled for different extensions. With appropriately scoped 
> attributes, objdump can correctly disassemble these functions even if there was an overlap in encodings.

ARM was deprecated that since ABI revision r2.09 (at November 2012)[1], and binutils seems not well supported now, so I think we need more discussion and investigation about that .

[1] ARM ABI. Page 29 of 45, http://infocenter.arm.com/help/topic/com.arm.doc.ihi0045e/IHI0045E_ABI_addenda.pdf

kito-cheng:
Changes:
- Revise the words.
- Add even/odd rule.
- Add reserved tag value range for non-standard attribute.

kito-cheng:
Current status:
- Implemented on GCC 9 and binutils 2.32.
- Merge policy between different arch version didn't specified on spec, my thought is leave it implementation-defined, binutils only allow exact version match, but we should implement different merge policy.
- Lacking of section or symbol scoped attributes which is nice to have for ifunc, so that we can dis-assemble binary correctly.
  - Upstream binutils DIDN'T implement section or symbol scoped attributes even for ARM port, so here is lots work need to do.

kito-cheng:
LLVM support has been merged into trunk:
  * ELF Attributes (MC layer): https://reviews.llvm.org/D74023
  * ELF Attributes (llvm-readelf): https://reviews.llvm.org/D75833

kito-cheng:
Changes:
- Rebase.

riscv-elf.md
```diff
@@ -996,6 +1015,71 @@ There are no RISC-V specific definitions relating to dynamic tables.
...
+
+Each attribute is described in the following structure:
+```<Tag name>, <Value>, <Parameter type 1>=<Parameter name 1>[, <Parameter type 2>=<Parameter name 2>]```
+
+#### Tag_RISCV_stack_align, 4, uleb128=value
+Tag_RISCV_strict_align records the N-byte stack alignment for this object. The
+default value is 16 for RV32I or RV64I, and 4 for RV32E.
+
+The smallest value will be used if object files with different Tag_RISCV_stack_align
+values are merged.
```
jrtc27, line 1053: I'm not sure about this one. Normally the alignment guaranteed by a .o and the alignment assumed by a .o are the same, meaning that if this differs between files they're incompatible (i.e. if the smaller one calls the bigger one, the bigger one might end up making an unaligned access). Do we want separate guaranteed+required fields? Or do we just want to say this is "the alignment you believe is in the ABI", and thus fail if they differ, at which point why does it exist, because the ABI specifies that already?

riscv-elf.md
```diff
@@ -996,6 +1015,71 @@ There are no RISC-V specific definitions relating to dynamic tables.
...
+in the attribute as `RV32I2P0` in which `2P0` stands for the default version of
+its based ISA. On the other hand, the architecture `RV32G` has to be presented
+as `RV32I2P0_M2P0_A2P0_F2P0_D2P0` in which the abbreviation `G` is expanded
+to the IMAFD combination with default versions of the standard extensions.
+
+#### Tag_RISCV_unaligned_access, 6, uleb128=value
+Tag_RISCV_unaligned_access denotes the code generation policy for this object
+file. Its values are defined as follows:
+ - 0: This object does not allow any unaligned memory accesses.
+ - 1: This object allows unaligned memory accesses.
```
jrtc27, line 1073: I think `allow` is the wrong word, perhaps `perform`?

riscv-elf.md
kito-cheng: I agree the merge policy should be  fail if they differ rather than min value of all input objects, different stack alignment seems hard to work together correctly.

The reason why we record that is we have `-mpreferred-stack-boundary=` option to control to stack alignment, so I thought it should record that to prevent hard to track when something is wrong due to that,

riscv-elf.md
kito-cheng: Thanks, `may perform` sounds better word here.

riscv-elf.md
jrtc27: X86 also has a separate `-mincoming-stack-boundary=` to realign the stack to a higher value if you want to use `-mpreferred-stack-boundary=` with a _higher_ value than the ABI requires (if you want to use a _lower_ value then that's always going to be incompatible). So I think we should require these always match until GCC or Clang have both options for RISC-V.

jim-wilson: The GCC RISC-V -mpreferred-stack-boundary= option only lets you increase alignment.  You can not decrease alignment below the ABI required alignment.  But there is still risk of incompatibliity if not everything is compiled with same alignment, so an error if different should be OK.

jim-wilson (approved):
with the stack align attribute error (or warning) on mismatch

kito-cheng:
Changes:
- Address @jrtc27's comment 
- Change the merge policy for Tag_RISCV_stack_align.

jrtc27:
Something that occurred to me the other day regarding this is that we need a program header too, not just a section header, since ELF loaders only look at program headers rather than the more complicated section headers (and require them to do otherwise will result in pushback). PT_LOPROC and PT_HIPROC are the same value as SHT_LOPROC and SHT_HIPROC, so I suggest we add a PT_RISCV_ATTRIBUTES with the same value as SHT_RISCV_ATTRIBUTES. This unfortunately means binaries produced today by GNU ld and LLD will not have information that loaders can usefully use, but that's true of binaries produced by toolchains before they added initial RISC-V attributes support, and so just have to be assumed compatible (which they should generally be, since IMAFDC are all that you can use without enabling experimental extensions in either toolchain).

kito-cheng:
@jrtc27 Good point, add PT_RISCV_ATTRIBUTES would make kernel or dynamic linker easier to get the location of ELF attribute.

kito-cheng:
binutils patch for PT_RISCV_ATTRIBUTES.
https://sourceware.org/pipermail/binutils/2021-June/116990.html

MaskRay:
Is there a justification for `PT_RISCV_ATTRIBUTES`? There are reasons why other `SHT_*_ATTRIBUTES` arches don't introduce `PT_*_ATTRIBUTES`.

I think it may be because these `SHT_*_ATTRIBUTES` are only for static link time checks which don't really need runtime detection checks. At runtime GNU_PROPERTY is prevailing, though I don't necessarily with the existing GNU_PROPERTY usage.

kito-cheng:
> Is there a justification for PT_RISCV_ATTRIBUTES? There are reasons why other SHT_*_ATTRIBUTES arches don't introduce PT_*_ATTRIBUTES.
> 
> I think it may be because these SHT_*_ATTRIBUTES are only for static link time checks which don't really need runtime detection checks. At runtime GNU_PROPERTY is prevailing, though I don't necessarily with the existing GNU_PROPERTY usage.

Yes, currently is used that for static link time only, and we are try to bring it into loader and dynamic linker, for checking the minimum ISAs, it's same propose as GNU_PROPERTY, but I think we already have build attribute for a while, so leverage that is kind of straightforward.

[1] https://sourceware.org/legacy-ml/gnu-gabi/2016-q4/msg00000.html

kito-cheng:
@MaskRay ping.

MaskRay:
I hope every program header type can find a use case. If you find PT_RISCV_ATTRIBUTES, it's fine.

The binutils patch stills needs a fix as it searches for the attribute section by name instead of by type.

kito-cheng:
Updated patch for PT_RISCV_ATTRIBUTES, accepted by RISC-V binutils maintainer, and will merge after binutils 2.37 branch created.
https://sourceware.org/pipermail/binutils/2021-June/117182.html

kito-cheng:
Gonna merge this PR now, welcome send PR or create issue if any further comment!

MaskRay:
Actually, I find `PT_RISCV_ATTRIBUTES` strange. I don't know a precedent that a program header includes a non-allocable section. The `SHT_RISCV_ATTRIBUTES` section is not contained by a `PT_LOAD` program header, so the  `SHT_RISCV_ATTRIBUTES` section may not be mapped, and dynamic loaders cannot use `PT_RISCV_ATTRIBUTES`.

PR #73, opened 2018-06-29
Mention that the fcsr is thread-local

sebhub:


jim-wilson:
The handling for fcsr is no different than the handling for f8 for instance, but we don't note that f8 is thread-local.  Though fcsr is a little less obvious than f8, so maybe some mention here is worthwhile.

I'm not an OS guy, but I find the use of the term thread-local a little confusing, because it sounds like thread local storage (TLS) to me.  But in this case, you are saying the fcsr must be part of a process/thread context.  So for instance the glibc getcontext function reads every callee-saved register and fcsr.  And the glibc setcontext function writes every callee-saved register and fcsr.  Similarly, the linux kernel puts every int and fp register in the process context, and also fcsr.

sebhub:
I don't think that the fcsr is a simple callee-saved register (e.g. f8) which is preserved across calls. Changes in the fcsr are under the control of the program and not the compiler, e.g. via C11 <fenv.h> or floating-point operations (e.g. underflow, overflow, etc.).

It is also probably not a caller-saved register.

It is more like thread-local storage. See for example:

https://en.cppreference.com/w/c/numeric/fenv

> The floating-point environment is the set of floating-point status flags and control modes supported by the implementation. It is thread-local, each thread inherits the initial state of its floating-point environment from the parent thread. Floating-point operations modify the floating-point status flags to indicate abnormal results or auxiliary information. The state of floating-point control modes affects the outcomes of some floating-point operations.

It should be clear from reading the psABI document what an operating system should do with the fcsr. In RTEMS we do currently the following:

1. Save/restore it at interrupt entry/exit.

2. Save/restore it before/after a thread context switch

sebhub:
Maybe we could write the following:

> The Floating-Point Control and Status Register (fcsr) must have thread storage duration in accordance with C11 section 7.6 "Floating-point environment <fenv.h>".

palmer-dabbelt:
@sebhub I think that is a reasonable wording, and since that's the only way to portably access FCSR it seems like a sufficient definition as far as the ABI is concerned.

palmer-dabbelt:
Thanks!

PR #74, opened 2018-07-06
Modify arg passing info for struct with one int and one fp to match GCC behavior.

jim-wilson:
This is in response to an FSF GCC bug report
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86387
which states that GCC doesn't match the ABi in the handling of structs with one int and one fp real.

Unfortunately, addressing this in GCC would cause a non-backward compatible ABI break, and we promised that the ABI would not change when we upstreamed glibc in February.  So we instead have to fix this in the ABI.


asb:
For the reasons you state, I think this is the pragmatic path forwards.

Perhaps we should have a changelog or errata section so it will be obvious to people reading the document that this detail recently changed?

jim-wilson:
I updated the patch.

Issue #79, opened 2018-09-25
Using RVE ABI with rv32i.


jim-wilson:
The current docs say
The RV32E calling convention may only be used with the RV32E ISA, hence the role of registers x16-x31 and f0-f31 is not defined. A future version of this specification may relax this constraint.

But it is easy to make rve work with rv32i by making the extra registers call clobbered (i.e caller saved).  This is how FP regs already work when using for instance the ilp32 ABI with the rv32gc architecture.  This is how the proposed EABI works with registers x16-x31.  And this is also how the gcc RVE implementation works, though it needed a small bug fix to make a6 and a7 usage depend on the ABI instead of the architecture.  So I think this should be allowed.

aswaterman:
IIRC the only reason we didn't define it was to punt on defining the role of x16-x31, but your proposal is the most natural one.  Any comments from @Kevin-Andes @asb @PkmX @kito-cheng ?

asb:
I think allowing -mabi=ilp32e to be used with -march=rv32i* in this way is the logical approach and can't see a good reason not to allow it.

kito-cheng:
LGTM, I heard some people are talking about possibility of extending RV32E into RV32I (e.g. existed RV32E system want to upgrading ISA to RV32I but don't want to rebuild whole SW system.), so I think it make it possible by this spec change.

PkmX:
LGTM.

Rather than opening another issue in `riscv-c-api-doc`, should `__riscv_32e` be defined with `-march=rv32i -mabi=ilp32e`? Or should we have another `__riscv_32e_abi` to differentiate the case?

aswaterman:
@jim-wilson do you mind contributing a PR that conveys your proposal?

palmer-dabbelt:
@Kevin-Andes IIRC we tried to design something that would be compatible with the fast-interrupts ABI but found that there would be some incompatibilities anyway.  Jim would know for sure, though, as he's actually doing the design :)

@PkmX I would be in favor of splitting the flags here.  Imagine you're trying to write some assembly function that must adhere to the RV32E ABI, but can be made better somehow by taking advantage of the extra registers in RV32I.  This is essentially the same as `-march=rv32if -mabi=ilp32` vs `-march=rv32if -mabi=ilp32f`.

asb:
[...]

PR #80, opened 2018-09-27
Make x16-x31 and f0-f31 temporary registers in the RV32E ABI.

jim-wilson:
Proposed wording to fix issue #79.


jim-wilson:
I don't have write access to riscv-elf-psabi-doc, so someone else will have to merge this.

PR #81, opened 2018-10-22
Clarify that "downwards" means negative

palmer-dabbelt:
Thanks to Michael D. Adams for pointing this out.  I also always forget
which way is up :)


asb:
Looks good to me.

PR #83, opened 2018-11-19
Remove assembler relocation function table.

jim-wilson:
Duplicates info in riscv-asm-manual, which is a better place for it.  And is
out-of-date.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -436,23 +436,6 @@ Enum | ELF Reloc Type       | Description                     | Details
 Nonstandard extensions are free to use relocation numbers 192-255 for any
 purpose.  These relocations may conflict with other nonstandard extensions.
 
-### Assembler Relocation Functions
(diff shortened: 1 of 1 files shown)


rofirrim:
I'm happy to see these going, I was confused recently thinking that the ELF psABI was the authoritative document for the relocation function syntax.

Would it be useful to explicitly state that the assembler syntax of relocations is to be found in the riscv-asm-manual?

For instance, a new paragraph after the relocation table that reads something like 

> This specification contains fragments written in assembler. The precise assembler syntax, including that of the relocations, is described in the RISC-V Assembly Programmer's Manual.

What do you think?

Thanks!

palmer-dabbelt:
@rofirrim That sounds fine to me, can you submit a PR?

rofirrim:
@palmer-dabbelt I will. Thanks!

PR #85, opened 2018-12-08
Add a reference to the assembler manual

rofirrim:
PR https://github.com/riscv/riscv-elf-psabi-doc/pull/83 will remove the obsolete table containing the assembler syntax of the relocation functions.

It seems a good idea to explicitly state that their syntax is specified in the assembler manual.

**Note**: I'm aware this might conflict with https://github.com/riscv/riscv-elf-psabi-doc/pull/83 so I'll wait until that PR is merged


jim-wilson:
I resolved the conflict and committed the patch.

rofirrim:
Great! Thanks @jim-wilson

PR #89, opened 2019-03-22
Fix contradictory description of extension requirements for int in int+fp structs for the FP ABI

asb:
Pull request #74 clarified that when an int+fp struct is passed in a
GPR+FPR in the floating point ABI, the integer isn't extended to XLEN
(i.e. it is anyext, just as if it were passed on the stack).
Unfortunately, a previous part of the sentence which mandated
sign/zero-extension was left in. This patch removes the contradictory
information.


palmer-dabbelt:
Thanks, Alex!

PR #96, opened 2019-06-08
Indicate that s0 is the frame pointer.

palmer-dabbelt:
Fixes: https://github.com/riscv/riscv-elf-psabi-doc/issues/95
Signed-off-by: Palmer Dabbelt <palmer@sifive.com>

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -62,6 +62,9 @@ x28-x31 | t3-t6        | Temporary registers    | No
 In the standard ABI, procedures should not modify the integer registers tp and
 gp, because signal handlers may rely upon their values.
 
+The presence of a frame pointer is optional.  If a frame pointer exists
+it must reside in x8 (s0), the register remains callee-saved.
+
 Floating-point Register Convention <a name=floating-point-register-convention>
 -------------------------------------------------------------------------
 Name    | ABI Mnemonic | Meaning                | Preserved across calls?


palmer-dabbelt:
This is better than nothing, but it caused me to ask the question: how do we know if s0 contains a frame pointer or some other value.  I don't think mandating that this is indicated by DWARF will fly, because the sort of people who want frame pointers tend to be allergic to DWARF.  The only other thing I could think of was to have a canonical function entry indicate the presence of a frame pointer.  Both GCC and LLVM appear to have easy to describe sequences frame pointer manipulation

    https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(j:1,lang:c%2B%2B,source:'int+ext(int+a)%3B%0A%0Aint+func(int+a)%0A%7B%0A++++return+ext(a+%2B+1)+%2B+2%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:riscv820,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'1',trim:'1'),lang:c%2B%2B,libs:!(),options:'-O3+-fno-omit-frame-pointer',source:1),l:'5',n:'0',o:'RISC-V+gcc+8.2.0+(Editor+%231,+Compiler+%231)+C%2B%2B',t:'0')),k:50,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:rv64clang,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'1',trim:'1'),lang:c%2B%2B,libs:!(),options:'-O3+-fno-omit-frame-pointer',source:1),l:'5',n:'0',o:'RISC-V+rv64gc+clang+(trunk)+(Editor+%231,+Compiler+%232)+C%2B%2B',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:50,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4

so I think mandating something is at least an option.  I'm not sure if it's worth the headache, though.

jim-wilson:
There are a number of different ways to emit the frame manipulation depending on the stack size, as we try to minimize the number of total instructions and maximize the number of compressed instructions.  So there might be one sp subtract or two, the subtract amount might be an immediate, loaded with one instruction, or loaded with two instructions.  Etc.

There are also multiple optimizations that can affect where the frame exists.  Instruction scheduling can move the fp set down into the body.  Shrink wrapping can move the frame construction into an if statement, and you can even have two different frame sizes in two different if statements.  If you are using alloca or variable length arrays, then the sp value can change in the middle of the function.  Sibling call optimization (aka tail call optimization) can confuse the frame, as it effectively overwrites the frame of the callee instead of allocating a new frame eliminating a function from the backtrace.  Etc.

From my point of view, if you want to unwind, you must use the DWARF unwind info.  And if you aren't willing to use the DWARF unwind info then I'm not going to waste my time trying to help you because you will never manage to get your unwinder working correctly without the DWARF unwind info.

The ARM ABIs by the way mandate that the unwind info must always be emitted by the compiler.  Though there is still an option to turn it off as some people never learn.

palmer-dabbelt:
I'm happy with just telling people they have to emit DWARF.  The people who are afraid to use DWARF seem to have no idea what they're talking about anyway.

PR #97, opened 2019-06-08
Mandate DWARF

palmer-dabbelt:
Signed-off-by: Palmer Dabbelt <palmer@sifive.com>

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -791,6 +791,8 @@ typedef struct
 
 # <a name=dwarf></a>DWARF
 
+DWARF unwind info must be present in all RISC-V ELF images.
+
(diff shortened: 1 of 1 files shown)


aswaterman:
Note this isn't currently the case in RISC-V GCC when compiling for a language that doesn't have exceptions (namely C).  By contrast, x86 GCC has `-fasynchronous-unwind-tables` by default, which I guess is what we should do.

jim-wilson:
I'm not sure if we should go this far.  This should certainly be discussed with the clang folks.

kito-cheng:
x86 is not only target enable `-fasynchronous-unwind-tables` by default, aarch64, s390, rs6000, tilegx and tilepro are also enable that by default.

PR #100, opened 2019-07-03
Further clarify treatment of bitfields in the floating-point calling convention

### LINKED ISSUE #99: Clarity on rules for bitfields in fp+int structs eligible for the FP calling convention
There is a mention for bitfields, but it's still slightly under-documented.

The behaviour I'm observing from GCC is that:
* `struct s1 { float f; long long int j : 32; }` and similar are passed as FPR+GPR on ILP32D (should be clarified in the doc that this is still allowable, despite `long long int` being 64 bits.
* A struct with a float + any number of zero-width bitfields is passed in an FPR, but passed according to the integer calling convention if there is e.g. a zero-width bitfield + a int or non-zero-width bitfield

Can someone please confirm the above are expected behaviour, suitable for documentation in the psabi-doc?

asb:
This patch reflects the observed behaviour from GCC, as discussed in
issue #99.

Note: this is a solution to #99 but not necessarily the best one (see my comment in that thread). Alternatively, we could always ignore zero-width bitfields when considering eligibility for passing according to the floating point calling convention which seems more consistent even if it doesn't fully match what GCC currently does.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -197,8 +197,9 @@ register if it is no more than FLEN bits wide and at least one floating-point
 argument register is available.  Otherwise, it is passed according to the
 integer calling convention.
 
-A struct containing just one floating-point real is passed as though it were
-a standalone floating-point real.
+A struct containing just one floating-point real or (as a special case) one
+floating-point real and any number of zero-size bitfields is passed as though
+it were a standalone floating-point real.
 
 A struct containing two floating-point reals is passed in two floating-point
 registers, if neither is more than FLEN bits wide and at least two floating-point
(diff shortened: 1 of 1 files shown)


aswaterman (approved):
I agree this might not be the best technical solution, but simply documenting what the GCC port happens to do here still seems preferable to me.

jim-wilson (approved):
This looks like a correct description of how gcc is implementing this.

jrtc27 (dismissed):

riscv-elf.md
```diff
@@ -215,7 +216,13 @@ without extending the integer to XLEN bits, provided the floating-point real
...
 and at least one floating-point argument register and at least one integer
 argument register is available. Otherwise, it is passed according to the
-integer calling convention.
+integer calling convention. Note that the bitfield may have a type that is
+larger than XLEN but can still be considered eligible for passing in this way
+provided the bitfield size is no more than XLEN bits wide. e.g. `struct {
+float f; int64_t i : 32; }` is eligible for being passed in a floating-point
+argument register and an integer argument register even when XLEN is 32.
+Zero-size bitfields are not ignored, so `struct { float f; int64_t i : 32; int
+0;`} would be passed according to the integer calling convention.
```
jrtc27, line 225: Hm, actually, https://github.com/riscv/riscv-elf-psabi-doc/issues/99#issuecomment-514844705 suggests GCC has since changed ABI for zero-size bitfields, meaning I think this sentence (and the diff above to add a special special case on top of the special case) should go. @jim-wilson do you agree?

PR #101, opened 2019-07-03
Clarify Usage of Paired Registers

lenary:
This document does not seem to specify how to split the bits in a 2*XLEN-bit scalar into the pair of XLEN-bit registers. It is not clear which register should contain the low word and which should contain the high word.

This change documents that the low-order XLEN bits go into the lower-numbered register, and the high-order XLEN bits go into the higher-numbered register, which seems to match what GCC and LLVM do.


aswaterman:
Good point.

jim-wilson:
This is correct for a little endian system, we want both little-byte endian and little-word endian.  When we add big-endian support we are going to need a lot of changes probably.

aswaterman:
I was thinking about that, too.  Even in this particular paragraph, there was a little-endian-oriented constraint in the case that the 2*XLEN value is split between a register and the stack.

PR #105, opened 2019-07-11
Document value of _Alignof(max_align_t) on RV32G and RV64G

### LINKED ISSUE #104: Document _Alignof(max_align_t)
The "Defualt ABIs and C type sizes" section does not document the value of `_Alignof(max_align_t)`, which is implementation-defined and part of the platform ABI.

`long double` has an alignment requirement of 16, which forces `_Alignof(max_align_t) >= 16` on the 32-bit and 64-bit ISAs. 

All memory allocations returned by `malloc` for sizes `>= _Alignof(max_align_t)` need to be aligned to a `_Alignof(max_align_t)` boundary, so from this POV, it makes sense to pick it "as large as necessary, as small as possible".

The SysV 32 and 64-bit ABIs set `_Alignof(max_align_t) == 16`, and this would also make sense here AFAICT. 

The only reason to pick a higher value is if we plan to have types in the ISA with a fundamental alignment larger than 16 bytes according to the C standard.

gnzlbg:
Closes #104.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -300,7 +300,7 @@ the imaginary part.
 A future version of this specification may define an ILP32 ABI for
 RV64G, but currently this is not a supported operating mode.
 
-The value of `_Alignof(max_align_t)` is 16 for both RV32G and RV64G.
+The value of `_Alignof(max_align_t)` is 16.
 
 ## <a name=va-list-va-start-and-va-arg></a> va_list, va_start, and va_arg
 
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -300,6 +300,8 @@ the imaginary part.
 A future version of this specification may define an ILP32 ABI for
 RV64G, but currently this is not a supported operating mode.
 
+The value of `_Alignof(max_align_t)` is 16 for both RV32G and RV64G.
+
 ## <a name=va-list-va-start-and-va-arg></a> va_list, va_start, and va_arg
 
 The `va_list` type is `void*`. A callee with variadic arguments is responsible


jim-wilson:
I think this should depend on ABI and not architecture.  And if it does depend on the architecture, then you have to use the base architectures rv32i, rv64i, and rv32e instead of rv32g and rv64g.  But probably simpler to just state unconditionally that it is 16 for now, because that is correct for all current ABIs.

gnzlbg:
@jim-wilson We do have options. For example, _iff_ we wanted to make, e.g., `long double` or `__int128` "optional" in some architectures, or give them an alignment requirement lower than 16 on some architectures, then it would make sense to make `_Alignof(max_align_t)` architecture-dependent. 

AFAICT, we want to require `long double` and `__int128` for all architectures, and at least `long double` is always going to have an alignment requirement of 16. So if this is the case, making this unconditional, makes the most sense. Do you agree? 

If so, I'll amend the commit to remove the "for RV32G and RV64G" part, and leave this guarantee as unconditional.

jim-wilson:
We don't have options.  The ABI was fixed when we upstreamed glibc.  We can't change it now (except in case of dire emergency).  At least that is true for rv64.  The rv32 glibc hasn't been upstreamed yet, but changing the ABI would still be disruptive and should be avoided.  So no ABI changes.  We can only document what the ABI already is.  You can see what the ABI already is by compiling small testcases with gcc.

Assuming that we need to document what max_align_t is.  I've been looking for evidence of that, the older x86 psABIs I pointed at don't document this, but I found newer ones at
    https://github.com/hjl-tools/x86-psABI/wiki/X86-psABI
I don't see a mention of max_align_t, but maybe this value is implied somewhere.  I will have to try reading it tomorrow.

gnzlbg:
> I don't see a mention of max_align_t, but maybe this value is implied somewhere. I will have to try reading it tomorrow.

See: https://gitlab.com/x86-psABIs/x86-64-ABI/blob/master/x86-64-ABI/low-level-sys-info.tex#L149

jim-wilson:
So I see that they moved from github to gitlab, and left all of the github stuff behind to confuse people like me.  Looking at the correct docs on gitlab now, I see that 6 days ago HJ added docs for max_align_t.  You filed this request 7 days ago which is interesting timing.

Anyways, yes, the latest revision looks correct, and adds something already present in the x86-64 psABI docs, so we should add it to the RISC-V psABI.

PR #106, opened 2019-07-11
Document the in-memory and in-register representation of _Bool and bool

gnzlbg:
Closes #105.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -300,6 +300,9 @@ the imaginary part.
 A future version of this specification may define an ILP32 ABI for
 RV64G, but currently this is not a supported operating mode.
 
+Booleans (`bool`/`_Bool_`), when stored in a memory object, are stored as single
+byte objects the value of which is always `0` (`false`) or `~0` (`true`).
+
 ## <a name=va-list-va-start-and-va-arg></a> va_list, va_start, and va_arg
 
 The `va_list` type is `void*`. A callee with variadic arguments is responsible
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -301,7 +301,9 @@ A future version of this specification may define an ILP32 ABI for
 RV64G, but currently this is not a supported operating mode.
 
 Booleans (`bool`/`_Bool_`), when stored in a memory object, are stored as single
-byte objects the value of which is always `0` (`false`) or `~0` (`true`).
+byte objects the value of which is always `0` (`false`) or `~0` (`true`).  When
+stored in integer registers (except for passing as arguments), all 8 bytes of
+the register are significant; any nonzero value is considered `true`.
 
 ## <a name=va-list-va-start-and-va-arg></a> va_list, va_start, and va_arg
 
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -302,8 +302,8 @@ RV64G, but currently this is not a supported operating mode.
 
 Booleans (`bool`/`_Bool_`), when stored in a memory object, are stored as single
 byte objects the value of which is always `0` (`false`) or `~0` (`true`).  When
-stored in integer registers (except for passing as arguments), all 8 bytes of
-the register are significant; any nonzero value is considered `true`.
+stored in integer registers (except for passing as arguments), all bytes of
+the register are significant and are always `0` (`false`) or `~0` (`true`).
 
 ## <a name=va-list-va-start-and-va-arg></a> va_list, va_start, and va_arg
 
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -300,10 +300,8 @@ the imaginary part.
 A future version of this specification may define an ILP32 ABI for
 RV64G, but currently this is not a supported operating mode.
 
-Booleans (`bool`/`_Bool_`), when stored in a memory object, are stored as single
-byte objects the value of which is always `0` (`false`) or `~0` (`true`).  When
-stored in integer registers (except for passing as arguments), all bytes of
-the register are significant and are always `0` (`false`) or `~0` (`true`).
+Booleans (`bool`/`_Bool`) stored in memory or registers (including when being
+passed as arguments), are either `0` (`false`) or non-zero (`true`).
 
 ## <a name=va-list-va-start-and-va-arg></a> va_list, va_start, and va_arg
 


jim-wilson:
The ISO C standard says that the only valid values for _Bool are 0 and 1.  It can't be ~0 in memory, and it can't use all 8 bytes of a register.  This patch is not OK.

gnzlbg:
Hi @jim-wilson , thanks for the feedback. Do you have a pointer to the paragraph of the C standard that requires that ?

The SysV ABI defines 0 as false, and any other bit-pattern as true - that is, there are multiple bit-patterns denoting true. I believe the SysV ABI to be compliant with the C standard. 

IIRC, the only thing the standard says is that _Bool needs to be at least large enough to store 0 or 1, and that when scalars are converted to _Bool the result is false if the scalar compares equal to 0, and otherwise the result is true. 

The true and false macros expand to the `1` and `0` integral constant expressions, which when converted to a _Bool, results in `0` being converted to false, and `1` to true; since these are constant-expressions, the conversion to the false and true bit-patterns is performed at compile-time by the compiler.

The standard is huge, so there might be a clause somewhere requiring _Bool to be either 0 or 1, but that would surprise me a bit, since that would mean that SysV is not compliant.

jim-wilson:
There is more than one SysV ABI.  You argument is meaningless without a reference.

First you say that the ISO C standard requires that values converted to bool must be 0 or 1.  Then you are saying that the correct representation in memory is ~0.  That is a contradiction.  It can't be both at the same time, unless we emit extra instructions to convert _Bool values which is pointless.  _Bool should be as efficient as possible.

It is wrong to try to define one true representation for _Bool.  It would be reasonable to say that 0 if false and any other value is true.  But when you try to force one particular value as true, that causes trouble.

I was referring to the sections you quoted.  True they don't require a particular value, but the important thing is that they don't say 0 and ~0 which is what you are trying to document, and that is wrong.

gnzlbg:
> First you say that the ISO C standard requires that values converted to bool must be 0 or 1. 

I think this is not what I said. I said: 

> when "scalars" (ints, floats, etc.) are converted to _Bool the result is **false** if the scalar compares equal to 0, and otherwise the result is **true**.

I see your point of making things as efficient as possible, and from that point-of-view, it would be ideal if we could "just use" any scalar as a _Bool. That is, if the scalar equals 0, the result is false, and the scalar itself is the result, and if the scalar doesn't equal 0, the result is true, and the scalar itself is the result. Having to convert that scalar to ~0 would indeed entail additional unnecessary work.  

So

> it would be reasonable to say that 0 if false and any other value is true. 

this sounds much better to me. 

While the compiler can easily do that in "private" registers, the question then is whether we want to allow that to be the in-memory representation of bool, and also, whether this can be used when passing bools in the calling convention.

gnzlbg:
So I've made the change in the last commit, changing the in-memory and in-register representation to 0 for false and non-zero for true. 

Even if C would require exactly 0 or 1, that requirement would be compatible with this ABI. 

However, iff C requires _Bool to always be exactly 0 or 1, this requirement would be sub-optimal for C, e.g., when calling _unknown_ functions that return booleans, because these function does not need to necessarily return 1 for true - any non-zero bit pattern would do. This might cause C to have to convert these values to 1, which isn't desired, but AFAICT, if this is required by the C standard, these conversions are unavoidable, e.g., the integer 3 would need to be converted to 1. So going one way or the other only switches the places where this conversions happen. 

A programming language that isn't C, and where `true` could be any non-zero bit-pattern, wouldn't have such a cost with this ABI specification though.

jim-wilson:
As mentioned in the other thread, we can't change the ABI.  It is too late for that.  We can only document what the ABI already is.  You can see what the ABI already is by compiling small examples with gcc.

gnzlbg:
It's unclear to me what the ABI already is. I found this program for x86_64:  https://gcc.godbolt.org/z/Y6fI4T

```c
unsigned char reg(_Bool b) {
    union U {
        unsigned char f0;
        _Bool f1;
    };
    union U u;
    u.f1 = b;
    if (u.f0 > 1) { 
        // This cannot happen
        // if b is only allowed
        // to be 0 or 1:
        return 42;   
    }
    return 13;
}
```

Anyone can call the `reg` function according to the ABI. Clang assumes that _Bool is 0 or 1, and always returns 13. GCC does not assume that. Does this mean that GCC supports multiple true values for _Bool? That GCC is just missing an optimization ? That Clang is mis-optimizing this code? No idea. 

AFAICT, the chances are that, since this is currently unspecified, and compilers need to assume something, different compilers assume different things and specifying this either way will break somebody.

GCC and Clang break their ABIs on every single release due to ABI-bugfixes. Compiler flags like `-fabi-version=clang7` exist to allow users to re-introduce the bugs and remain abi compatible. [0]

If the priority is not breaking any assumptions that GCC made about this, I think the best bet here is pinging someone working on GCC RV and asking them to explain which assumptions does GCC make about this. We can then specify that, and ping clang and other compiler vendors to tell them to upgrade their ABIs.

I think that the best way forward would be to pick something that's compatible with the C standard, and that makes sense for the architecture, and then propose a specification and let compiler-vendors chime in and listen to their feedback. If, for whatever reason, what we pick is implementable in GCC, but impossibly hard to implement in Clang, the better trade-off might be to pick something that works for both.

---

[0] For example, GCC fixed the value of `_Alignof(max_align_t)` from 8 to 16 on Intel386 psABI in GCC 7.0. Clang still has a bug (https://bugs.llvm.org/show_bug.cgi?id=42525), and this bug was introduced in Clang because Clang wanted to be "bug-for-bug compatible" with GCC at the ABI level. When GCC fixed the bug, Clang didn't notice, so here we are.

gnzlbg:
I'm going to close this and just document equality with 0 and 1. That should be compatible with GCC, since in the case that GCC assumes that multiple value true values exist, 1 is definitely one of them.

jim-wilson:
Interesting example.  I think the problem is only that gcc is failing to optimize based on the range of a boolean/_Bool argument.  We do have a value range propagation pass, but it apparently isn't propagating the range of the _Bool b incoming arg, which is 0 to 1.

gnzlbg:
At the end of the day is up to the ABI to say what’s ok and what isn’t.

I’m not super familiar with RISC-V but on x86 you can use a “0 for false
and true otherwise” definition quiet efficiently using “jump on nonzero”
instructions. I believe this is why the x86-64 psABI has the weird sentence
saying that inside functions in register compilers are allowed to do better
than using 0 or 1. I think there is a patch to remove that sentence
somewhere that argues that compilers must only conform to the ABI at
boundaries and inside functions they can do whatever they want. At some
point somebody must have thought that making it clear that these
optimizations are ok was worth spelling out.

I don’t know whether RISC-V has these jump on non zero instructions or
similar, but ABIs can be quiet impactful here.

PR #111, opened 2019-07-23
Ignore zero-length bit-fields when flattening.

jim-wilson:
Fixes the zero-length bit-field question in issue #99.


aswaterman:
lgtm

asb:
Looks good to me.

jim-wilson:
Two lgtms from the main interested parties, so I'm considering it approved and merging it.

PR #113, opened 2019-08-22
[checker] Initial Sanity Checker (for C compilers)

lenary:
The RISCV psABI document currently specifies a lot of platform-specific details, but we have no automated way of checking whether a given compiler matches these details.

This Pull Request is the initial version of a simple test suite for checking a C compiler matches the specification. At the moment, it supports:
- Checking type sizes
- Checking type alignments
- Checking atomic type sizes and alignments (for power-of-two-sized types only)
- Checking abi-specific preprocessor defines (such as `__riscv` and `__riscv_float_abi_soft`).

Current Results:
- GCC Currently passes all checks. 
- Clang does not pass any `ilp32e` tests. My test machine also seems to have path-handling issues with the `ilp32f` and `lp64f` tests.

The C sources are written using standard C11 features.

Licensing:
Creative Commons licences are not intended for use on code. I have chosen to licence this work under Apache 2.0.

Future steps for the C part of this sanity checker:
- Checking architecture-specific preprocessor defines (such as `__riscv_mul`).
- Checking compilers accept both abi and numeric register names.

Out of scope of the C part of this sanity checker:
- Checking Calling Convention
- Checking ELF Object Files
- Checking DWARF Information.


asb:
I'm (obviously) a fan of this work, though appreciate people might question whether this repo is the right place for it. It seems a reasonable home from my perspective though.

Given the scope of this tool doesn't include the whole ABI and it necessarily relies on some things which aren't directly required by the ABI (e.g. checking preprocessor defines), I wonder whether renaming it the "sanity checker" might be clearer as to its scope and purpose?

lenary:
Re-raising this issue: Is there any interest in this sanity checker becoming part of this repository? A clear yes/no would be helpful so I can plan the next steps for this code.

aswaterman:
My opinion is that this looks like a very useful tool that belongs in its own repository.

lenary:
@aswaterman thanks for the feedback. I have split this into its own repository here: https://github.com/lowRISC/riscv-elf-psabi-sanity-checker

Issue #36, opened 2017-08-15
R_RISCV_32_PCREL relocation is undocumented


asb:
I see binutils 2.29 added a new R_RISCV_32_PCREL relocation. Could somebody please document its purpose and semantics?

dvc94ch:
I started modeling some of the relocations to better understand what they are for and what they should do. [gist](https://gist.github.com/dvc94ch/9dc249d0dcaab5e4160ce03718d08c71)

It seems that one of the models is wrong, I didn't look at the binutils code, just compared the output with objdump.
> The successive instruction has a signed 12-bit immediate so the value of the preceding high 20-bit relocation may have 1 added to it.

PkmX:
Correct me if I'm wrong. `R_RISCV_32_PCREL` is mainly used in `.eh_frame` in the PC begin field of a FDE; the semantics is simply the offset from the relocation to the symbol (`S - P`).

PkmX:
@dvc94ch Your `R_RISCV_HI20` and `R_RISCV_LO12_{I,S}` doesn't handle the case where the lo part is sign-extended, no? See the "Absoluate Addresses" section for details.

dvc94ch:
Oh I see how you're supposed to load 0x7FFFFFFF now, thanks! I just tested the example from the asm manual and the objdump seems wrong or am I still missing something (like li works different on rv64)? [https://github.com/riscv/riscv-asm-manual/blob/master/riscv-asm.md#load-immediate](https://github.com/riscv/riscv-asm-manual/blob/master/riscv-asm.md#load-immediate)
Maybe a section on writing assembly that is portable accross rv32 and rv64 could be added or something like that?
Sorry for hijacking the thread...

PkmX:
The code you linked for loading a0 with `0xcafebabe` is incorrect too. It actually loads `0xc6febabe`. 

`li a0, 0xcafebabe` on RV32 gives:

```asm
lui a0, 0xcafec
addi a0, a0, 0xabe
```

On RV64 it is:

```asm
lui     a0,0x33
addiw   a0,a0,-1029
slli    a0,a0,0xe
addi    a0,a0,-1346
```

jim-wilson:
Fixed via #115

PR #115, opened 2019-09-20
Add pc-relative relocation for FDE initial location.

Nelson1225:
According to this commit:
https://github.com/riscv/riscv-binutils-gdb/commit/a6cbf936e3dce68114d28cdf60d510a3f78a6d40

Also, I notice that asb had already support the relocation in LLVM
https://reviews.llvm.org/D64715

Therefore, we should add it to the psabi spec :)


jim-wilson:
This is issue #36.  This is also covered by pull request #110, but that was making multiple changes and there is an unresolved issue with one of the changes, so I asked that it be split into multiple changes which didn't happen, so this is still pending.

Anyways, this is correct.

PR #116, opened 2019-09-23
Add relocation `R_RISCV_RELAX_NORVC` for linker relaxation.

Nelson1225:
Dear all,

According to this discussion:
https://github.com/riscv/riscv-gnu-toolchain/issues/445

I think this can be divided into two different problems:
1. Current linker may relax instruction sequence to the compressed one for the norvc code.
2. Alignment issue when both rvc and norvc code present.

For now, GNU ld checks the ELF e_flag of each input object, and then decide whether or not to do the rvc relaxations.  Consider the following case:

.option rvc # Enable rvc
call Label # The first call
...
.option norvc # Disable rvc
call Label # The second call
...

The ELF e_flag's EF_RISCV_RVC of the object file is set to true if the rvc is enabled somewhere, even if we disable the rvc in another place.  Therefore, linker will convert the two `call` to 2-byte compressed `jal`, but the second `c.jal` isn't allowed since the `.option norvc` is set.  Checking the ELF e_flag is a file level (or object level) method, so we need another checking method to solve the first problem when both rvc and norvc code present in the same object.

Add a new relocation `R_RISCV_RELAX_NORVC` seemed the only solution.  The instruction sequence paired with `R_RISCV_RELAX` indicating that linker can relax this sequence to the shorter one, including the compressed instruction.  The `R_RISCV_RELAX_NORVC` is same as `R_RISCV_RELAX` except that the sequence can't be relaxed to the compressed instruction.

Even if the first problem is resolved, there still have some cases cause the second alignment problem, so I think that is another issue.  This change of elf-psabi try to solve the first problem, not the second one :)

Thanks
Best Regards
Nelson

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -445,7 +445,8 @@ Enum | ELF Reloc Type       | Description                     | Details
 55   | R_RISCV_SET16        | Local label subtraction         |
 56   | R_RISCV_SET32        | Local label subtraction         |
 57   | R_RISCV_32_PCREL     | PC-relative reference           | word32 = S + A - PC
-58-191  | *Reserved*        | Reserved for future standard use |
+58   | R_RISCV_RELAX_NORVC  | Instruction pair can be relaxed, but without the use of compressed instructions |
+59-191  | *Reserved*        | Reserved for future standard use |
 192-255 | *Reserved*        | Reserved for nonstandard ABI extensions |
 
 Nonstandard extensions are free to use relocation numbers 192-255 for any
@@ -575,14 +576,14 @@ The pseudoinstruction:
 expands to the following assembly and relocation:
 
 ```
-    auipc ra, 0           # R_RISCV_CALL (symbol), R_RISCV_RELAX (symbol)
+    auipc ra, 0           # R_RISCV_CALL (symbol), R_RISCV_RELAX
     jalr  ra, ra, 0
 ```
 
 and when `-fpic` is enabled it expands to:
 
 ```
-    auipc ra, 0           # R_RISCV_CALL_PLT (symbol), R_RISCV_RELAX (symbol)
+    auipc ra, 0           # R_RISCV_CALL_PLT (symbol), R_RISCV_RELAX
     jalr  ra, ra, 0
 ```
 
@@ -645,6 +646,75 @@ label:
    add t2, t2, t1
    sw t2, t0, %pcrel_lo(label)   # R_RISCV_PCREL_LO12_S (label)
 ```
+
+
+### R_RISCV_RELAX and R_RISCV_RELAX_NORVC
+
+A sequence with instructions marked by `R_RISCV_RELAX` indicates that the
+linker can relax them to a shorter one, including by using compressed
+instructions.  `R_RISCV_RELAX_NORVC` is same as `R_RISCV_RELAX`, except that
+compressed instructions cannot be used, only uncompressed ones.
+
+Example assembler shows how these two relocations work:
+
+```
+   .option rvc              # Enable C extension
+
+   call  symbol1
+
+   lui   a0,%hi(symbol2)
+   addi  a0,a0,%lo(symbol2)
+
+   .option norvc            # Disable C extension
+
+   call  symbol1
+
+   lui   a0,%hi(symbol2)
+   addi  a0,a0,%lo(symbol2)
+```
+
+expands to the following assembly and relocation:
+
+```
+   auipc ra, 0              # R_RISCV_CALL (symbol1), R_RISCV_RELAX
+   jalr  ra, ra, 0
+
+   lui   a0,%hi(symbol2)    # R_RISCV_HI20 (symbol2), R_RISCV_RELAX
+   addi  a0,a0,%lo(symbol2) # R_RISCV_LO12 (symbol2), R_RISCV_RELAX
+
+   auipc ra, 0              # R_RISCV_CALL (symbol1), R_RISCV_RELAX_NORVC
+   jalr  ra, ra, 0
+
+   lui   a0,%hi(symbol2)    # R_RISCV_HI20 (symbol2), R_RISCV_RELAX_NORVC
+   addi  a0,a0,%lo(symbol2) # R_RISCV_LO12 (symbol2), R_RISCV_RELAX_NORVC
+```
+
+and then the linker may relax them to the following code:
+
+```
+   c.j/c.jal symbol1           # `c.j` exists on RV32 and RV64, but `c.jal`
+                                 is RV32-only.
+or
+   jal/jalr symbol1            # If the immediate of compressed instructions
+                                 can not cover the symbol1, then try this
+
+   addi a0, gp, (symbol2 - gp) # Do this first if symbol2 is in range of gp
+or
+   c.lui a0 %hi(symbol2)       # If symbol2 isn't in range of gp, then try
+   addi a0, a0, %lo(symbol2)     to relax the `lui` to `c.lui`
+
+   jal/jalr symbol1            # R_RISCV_RELAX_NORVC is paired with `call`,
+                                 the `c.j` and `c.jal` are not allowed
+
+   addi a0, gp, (symbol2 - gp) # R_RISCV_RELAX_NORVC are paired with `lui`
+                                 and `addi`, the `c.lui` is not allowed
+```
+
+The first `call` may be relaxed to a compressed `jal` or `j`, and the second
+`lui and addi` pattern may be relaxed to the `c.lui and addi` pattern by the
+linker if possible, but the third and fourth one can only be relaxed to a 4-byte
+`jal`/`jalr` and `addi with gp relative address` since the `.option norvc`
+directive is set.
 
 
 ## <a name=thread-local-storage></a>Thread Local Storage


kito-cheng:
@asb @jim-wilson @palmer-dabbelt @aswaterman

jim-wilson:
The general concept looks correct to me.

It mentions instruction pairs, but that is really more a property of R_RISCV_CALL than R_RISCV_RELAX.  R_RISCV_RELAX only applies to the one previous relocation, and that relocation only applies to an instruction pair if it is R_RISCV_CALL.  The other relocs only apply to one instruction.  For instance, if we have a lui/addi sequence, then each instruction has its own reloc and each reloc has its own R_RISCV_RELAX (or R_RISCV_RELAX_NORVC).

marceg:
Consider perhaps implementing something like Xtensa's property tables (see Xtensa port in BFD).  They're an extra ELF section with {vma, size, flags} triplets that identify various properties of sequences of code or data.  (The Xtensa equivalent of norvc is XTENSA_PROP_INSN_NO_DENSITY.) 
Extremely useful for this sort of thing, including required alignment of various subblocks of a section, which allows the linker to relax correctly in the presence of .align directives.  And a lot of the linker work is already done.  Relocations, on the other hand, target a specific instruction or word/address.

jim-wilson:
ARM has mapping symbols.  Every time you switch from one type of output to another, gas emits a mapping symbol, and a mapping symbol can indicate that the following block is arm code, thumb code, or data.  This is used by the disassembler.  Arm and thumb are different overlapping ISAs so you always have to know which mode you are in.  Also, ARM has constant pools that can appear in the middle of functions, which we don't want to disassemble as code.  These aren't used by the linker though.  The only bfd support is to make sure that strip doesn't remove the mapping symbols.

RISC-V doesn't have this problem as we don't have two different overlapping ISAs that can be linked together, and we don't have constant pools in the middle of functions.  But maybe we will have some of these problems later.

The xtensa property tables do look interesting.  This is a much more general solution than the ARM solution.  But it looks like a lot of code both in the assembler and in bfd to make this work.  This does not look like an easy approach.  I don't see anything being done by the linker, other than the fact it will copy the sections around for us.  There is still a lot of hard work being done in the xtensa bfd port to read the property tables and parse them and use them.  With a property table entry created for each frag, this also looks like it will increase object code size and maybe executable size if we copy the info over to executables.

We have two assemblers/linkers (clang/lld and GNU as/ld).  Property tables look like too much work to solve the simple problem we have right now.

kito-cheng:
Forgot to ping @MaskRay @PkmX , our important friends on LLD side :P

marceg:
Property tables add a bit to the ELF file, but they're never loaded in target memory so the size hasn't been an issue.  Mapping symbols might be able to do the same thing (and probably take a little more space in the ELF file).  Perhaps the most useful and most complicated benefit is the ability of the linker to shrink and expand sections when it applies relocations, instead of having to pad NOPs -- which requires handling alignment directives properly.  It definitely helps code size and performance.  However doing this correctly is a bit tricky, and sometimes unintuitive:  for example, shrinking section A followed by an align directive then section B, might increase the distance between a branch and its target across A and B, causing it to go out of range.  I haven't looked at the code lately, perhaps that's some of the complexity you were seeing (or perhaps it's more in ld than bfd); and indeed no need for constant pool handling code (not expecting a single function to span more than 4 GB :).

Is it currently possible to reliably identify what parts of a RISC-V ELF section are code vs data?  In particular when .align directives are present in the code (whether for performance, cache alignment, or otherwise).  The ability to know exactly what parts of the ELF file contains code vs data has been particularly helpful for profiling tools and code transformation tools.

PkmX:
Well, the concept of `R_RISCV_RELAX_NORVC` seems sound to me.

@marceg Would you explain how property tables would help in those situations? RISC-V kind of workaround the cross-section relaxation with alignment issue by considering the maximum alignment when trying to determine if a relocation can be relaxed.

marceg:
They identify .align directives within a section.  My example of sections A and B is more accurately blocks A and B within the same section, as relative branches (conditional branches in particular, which have limited reach) don't generally cross section boundaries.  It's not possible generally to grow or shrink an instruction or instruction sequence (such as for a call, given a relocation) without information about any .align directive following these instructions in the same section.  Such directives are not very common, so if the linker currently shrinks/expands code based on RELAX_RVC, it may be lucky in that it usually isn't an issue, but it might not be correct.  Mid-section .align directives happen a lot more in embedded code than Linux-ish code, I suspect.  Or to align branch targets for performance (for those, there's a performance loss but no correctness issue).

Granted such embedded .align were common in Xtensa due to its requirement to 4-byte align PC-relative call targets.  However, there were several other cases (cache-line alignment within a function; vector-table alignment, for example regularly-spaced sets of vectors with variable code in each, with relocations that affect their size; jump tables; others too far to remember).

jim-wilson:
For RISC-V, we currently don't mix code and data in the same section, so there is no problem telling them apart.  This would change if we ever implemented constant pools in the text section.

PkmX:
The way RISC-V handles `.align` is that it emits `alignment - 2` bytes of `nop` and the linker runs an extra pass after relaxation to remove those `nop`s to the correct alignment. Since during the whole relaxation + alignment process offsets within a single section only decrease monotonically, those relocations cannot go out of range. Of course, this comes with a cost of some missed relaxation opportunities, since emitting `alignment - 2` bytes of `nop` pretty much forces the linker to compute offsets using the worst case scenario.

marceg:
Interesting.  Effectively the object file is not aligned until the linker fixes it.  (No easy direct use of .o's as observed in the wild.)  I'm not sure how that plays with vector section sizing checks.  I suppose you haven't seen a lot of .org in assembly code.  (Yet :)

RISC-V's PC-relative accesses have a +/- 2 GB range (you pretty much have to use AUIPC), so literal pools can probably safely be after the whole text section, and thus always be in a separate section, assuming code never reaches 2 GB in size (so far, a fair assumption).  Perhaps more likely are the weird things sometimes done in hand-written assembly, especially for the embedded/OS space.

Nelson1225:
Dear Jim, Marc, Kito and PkmX,

Thanks for your suggestion and reply :)

Currently, it seems that add a new relocation `R_RISCV_RELAX_NORVC` is the more direct and simple way to resolve the problem that linker may relax instruction sequence to the compressed one for the norvc code.  Just like Jim said, we would use the more appropriate method if we ever implemented the constant pools in the text section.

I have rewritten the description of the new relocation `R_RISCV_RELAX_NORVC`.  Please see the details in the newest commit.

Thanks again
Best Regards
Nelson

riscv-elf.md
```diff
@@ -647,6 +648,79 @@ label:
...
+   call  symbol1
+
+   lui   a0,%hi(symbol2)
+   addi  a0,a0,%lo(symbol2)
+```
+
+expands to the following assembly and relocation:
+
+```
+   auipc ra, 0              # R_RISCV_CALL (symbol1), R_RISCV_RELAX (symbol1)
```
jrtc27: `R_RISCV_RELAX` has zero symbol and addend (printed as `*ABS*` by objdump), as all the information it needs is in the other relocation for the same offset. I notice however the existing text about `R_RISCV_RELAX` gets this wrong too.

riscv-elf.md
```diff
@@ -647,6 +648,79 @@ label:
...
+
+   jal/jalr symbol1            # R_RISCV_RELAX_NORVC is paired with `call`,
+                                 the `c.j` and `c.jal` are not allowed
+
+   addi a0, gp, (symbol2 - gp) # R_RISCV_RELAX_NORVC are paired with `lui`
+                                 and `addi`, the `c.lui` is not allowed
+```
+
+The first `call` may be relaxed to a compressed `jal` or `j`, and the second
+`lui and addi` pattern may be relaxed to the `c.lui and addi` pattern by linker
```
jrtc27: ... pattern by _the_ linker

riscv-elf.md
```diff
@@ -445,7 +445,8 @@ Enum | ELF Reloc Type       | Description                     | Details
 55   | R_RISCV_SET16        | Local label subtraction         |
 56   | R_RISCV_SET32        | Local label subtraction         |
 57   | R_RISCV_32_PCREL     | PC-relative reference           | word32 = S + A - PC
-58-191  | *Reserved*        | Reserved for future standard use |
+58   | R_RISCV_RELAX_NORVC  | Instruction pair cannot be relaxed to rvc |
```
jrtc27: Description is unclear (without reading the later content); it can still be relaxed, just not with compressed instructions. What about something like "Instruction pair can be relaxed, but without the use of compressed instructions" (based on the description of `R_RISCV_RELAX`)?

riscv-elf.md
```diff
@@ -647,6 +648,79 @@ label:
...
+   jal/jalr symbol1            # R_RISCV_RELAX_NORVC is paired with `call`,
+                                 the `c.j` and `c.jal` are not allowed
+
+   addi a0, gp, (symbol2 - gp) # R_RISCV_RELAX_NORVC are paired with `lui`
+                                 and `addi`, the `c.lui` is not allowed
+```
+
+The first `call` may be relaxed to a compressed `jal` or `j`, and the second
+`lui and addi` pattern may be relaxed to the `c.lui and addi` pattern by linker
+if possible, but the third and forth one can only be relaxed to a 4-byte
```
jrtc27: fourth

riscv-elf.md
```diff
@@ -647,6 +648,79 @@ label:
 ```
 
 
+### R_RISCV_RELAX and R_RISCV_RELAX_NORVC
+
+The instruction and the instruction sequence paired with `R_RISCV_RELAX`
+indicating that linker can relax them to the shorter one, including the
+compressed instruction.  The `R_RISCV_RELAX_NORVC` is same as `R_RISCV_RELAX`
+except that the instruction and the sequence can't be relaxed to the
+compressed instruction.
```
jrtc27: Paragraph has some minor grammatical issues, perhaps:

> A sequence with instructions marked by `R_RISCV_RELAX` indicates that the linker can relax them to a shorter one, including by using compressed instructions. `R_RISCV_RELAX_NORVC` is the same as `R_RISCV_RELAX`, except that compressed instructions cannot be used, only uncompressed ones.

riscv-elf.md
Nelson1225: Yes, you are right.  How about use `R_RISCV_RELAX` only rather than attach a symbol after it?

For example:
auipc ra, 0              # R_RISCV_CALL (symbol1), R_RISCV_RELAX
jalr ra, ra, 0

riscv-elf.md
Nelson1225: OK, already fixed.

riscv-elf.md
```diff
@@ -647,6 +648,79 @@ label:
...
+                              R_RISCV_RELAX_NORVC (symbol1)
+   jalr  ra, ra, 0
+
+   lui   a0,%hi(symbol2)    # R_RISCV_HI20 (symbol2),
+                              R_RISCV_RELAX_NORVC (symbol2)
+   addi  a0,a0,%lo(symbol2) # R_RISCV_LO12 (symbol2),
+                              R_RISCV_RELAX_NORVC (symbol2)
+```
+
+and then linker may relax them to the following code:
```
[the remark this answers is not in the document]
Nelson1225: OK, already fixed.

riscv-elf.md
Nelson1225: OK, already fixed.

riscv-elf.md
Nelson1225: Thanks for the suggestion, looks good to me.

riscv-elf.md
Nelson1225: Looks good to me, thanks :)

Nelson1225:
Dear jrtc27,

Sorry for the delayed response, and thank you so much for your review and suggestions :)
I have fixed the grammatical issues and modified some descriptions according to your suggestions.

Thanks again
Best Regards
Nelson

kito-cheng:
@MaskRay let me know if you think it's LGTM to you :)

asb:
I feel it would be worth finding some kind of resolution to this for 1.0.

Given the time constraints, the path to doing so would probably take the form of a note for implementers in the linker relaxation section noting the potential unresolvable interaction of alignment directives and options to disable RVC support for regions of code. @jrtc27 @kito-cheng would do you think?

Nelson1225:
BTW, if we could generate the mapping symbols with ISA, then we probably don't need to add RELAX_NORVC relocation anymore, https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/196.  But the disadvantage is that we need to check the symbol table before relaxing, that will take extra link times.

MaskRay:
Thanks for the explanation. Adding `R_RISCV_RELAX_NORVC` makes sense to me.

kito-cheng:
@Nelson1225 could you rebase that and sending PoC patch to mailing list?

MaskRay:
Ping for rebase :)

jrtc27:
My concern is this is specific to RVC and the same issue will arise when dealing with other extensions in future whose instructions the linker might want to use in general. Given R_RISCV_RELAX currently has no symbol associated with it, just the type, I would propose using dummy symbols for the ISA string, and would suggest using the same encoding for them as the mapping symbols (then you can reuse the same symbols, saving space and making the implementation easy, and thinking forward to things like CHERI-RISC-V the mapping symbols will need to include the encoding mode (we'd do $c<ISA> rather than $x<ISA> to mirror what Arm did for Morello) as the encoding mode is an orthogonal concept to the list of extensions so this ensures that information is still available. A symbol index of 0 would continue to mean what it does today. Using the symbol for this also has the advantage of backwards-compatibility, since I believe both binutils and LLD totally ignore the symbol for R_RISCV_RELAX, rather than a new relocation which would be a hard error unless you have a new enough linker.

kito-cheng:
TL;DR: I think @jrtc27's proposal is good idea and I support that.

@Nelson1225 and me has some offline discussion about the possibility of using mapping symbol instead of adding `R_RISCV_RELAX_NORVC` to resolve this issue, but we thought that might complicated the linker implementation, because linker need to scan the `R_RISCV_RELAX` located in which region, but I think @jrtc27 just give a good idea to resolve this issue! let `R_RISCV_RELAX` to bind with a dummy symbol sounds simple and more scalable solution for future, we don't need any new relocation `R_RISCV_RELAX_NORVYYY` for future extensions, and also resolve the immediate issue for `Zc*` extensions.

Nelson1225:
I agree with @jrtc27, refer the mapping symbol to R_RISCV_RELAX should work.

Nelson1225:
[PATCH 1/2] RISC-V: Output mapping symbols with ISA string.
https://sourceware.org/pipermail/binutils/2022-September/123192.html

[PATCH 2/2] RISC-V: Refer mapping symbol to R_RISCV_RELAX for rvc relaxations.
https://sourceware.org/pipermail/binutils/2022-September/123191.html

Here are Jessica's proposal, work as expected, pass regressions of riscv-gnu-toolchain.

PR #117, opened 2019-10-09
Clarify Named ABIs and C type sizes and alignments

lenary:
I have found the specification confusing, because C type sizes and
alignments have so far been defined only for the *architectures* RV32G
and RV64G. This is confusing, as there are 32- and 64-bit architectures
without all parts of the G extension which also share these type sizes
and alignments.

This pull request separates the definitions of C type sizes and alignments
into new section, and defines them for particular *ABIs*, not
architectures. The definitions themselves remain largely unchanged,
though wording has been clarified and extra subsection headers have been
introduced.

For the avoidance of doubt around C struct alignment rules, I have
defined the RISC-V psABI rules around size and alignment of structures
and unions. This is based on both the x86-64 and Intel386 System V psABI 
(which is not mentioned in the specification so far).

This now explicitly enumerates the sizes and alignments of C11's complex
types into the tables of sizes and alignments, based on the defined rules.

This also adds definitions of the following "named" ABIs, in terms of the
calling convention and C type details. These definitions match the
interpretation given by the `-mabi` argument to GCC and Clang.
- ilp32
- ilp32f
- ilp32d
- ilp32e
- lp64
- lp64f
- lp64d
- lp64q (I don't know if this is actually supported by GCC and Clang, but this combination is mentioned in "Default ABIs", so it seems reasonable to define it based on the existing naming scheme)

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -35,15 +35,16 @@
 
 This RISC-V ELF psABI specification document is
 
- &copy; 2016 Palmer Dabbelt <palmer@dabbelt.com>
- &copy; 2016 Stefan O'Rear <sorear2@gmail.com>
- &copy; 2016 Kito Cheng <kito.cheng@gmail.com>
- &copy; 2016-2017 Andrew Waterman <aswaterman@gmail.com>
- &copy; 2016-2017 Michael Clark <michaeljclark@mac.com>
- &copy; 2017-2019 Alex Bradbury <asb@asbradbury.org>
- &copy; 2017 David Horner <ds2horner@gmail.com>
- &copy; 2017 Max Nordlund <max.nordlund@gmail.com>
- &copy; 2017 Karsten Merker <merker@debian.org>
+ &copy; 2016 Palmer Dabbelt <palmer@dabbelt.com>  
+ &copy; 2016 Stefan O'Rear <sorear2@gmail.com>  
+ &copy; 2016 Kito Cheng <kito.cheng@gmail.com>  
+ &copy; 2016-2017 Andrew Waterman <aswaterman@gmail.com>  
+ &copy; 2016-2017 Michael Clark <michaeljclark@mac.com>  
+ &copy; 2017-2019 Alex Bradbury <asb@asbradbury.org>  
+ &copy; 2017 David Horner <ds2horner@gmail.com>  
+ &copy; 2017 Max Nordlund <max.nordlund@gmail.com>  
+ &copy; 2017 Karsten Merker <merker@debian.org>  
+ &copy; 2019 Sam Elliott <selliott@lowrisc.org>  
 
 It is licensed under the Creative Commons Attribution 4.0 International
 License (CC-BY 4.0).  The full license text is available at
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -310,6 +310,8 @@ There are two conventions for C type sizes and alignments.
 
 The alignment of `max_align_t` is 16.
 
+`CHAR_BITS` is 8.
+
 Structs and unions are aligned to the alignment of their most strictly aligned
 member. The size of any object is a multiple of its alignment.
 
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -26,6 +26,9 @@
 	* [Hash Table](#hash-table)
 5. [DWARF](#dwarf)
 	* [Dwarf Register Numbers](#dwarf-register-numbers)
+6. [Linux-specific ABI](#linux-abi)
+	* [Linux-specific C type sizes and alignments](#linux-c-type-sizes)
+	* [Linux-specific C type representations](#linux-c-type-representations)
 
 ## Copyright and license information
 
@@ -833,3 +836,29 @@ Dwarf Number  | Register Name | Description
 The alternate frame return column is meant to be used when unwinding from
 signal handlers, and stores the address where the signal handler will return
 to.
+
+# <a name=linux-abi></a> Linux-specific ABI
+
+**This section of the RISC-V ELF psABI specification only applies to Linux-based
+systems.**
+
+In order to ensure compatibility between different implementations of the C
+library for Linux, we provide some extra definitions which only apply on those
+systems. These are noted in this section.
+
+## <a name=linux-c-type-sizes></a> Linux-specific C type sizes and alignments
+
+The following definitions apply for all ABIs defined in this document. Here
+there is no differentiation between ILP32 and LP64 abis.
+
+Type        | Size (Bytes)  | Alignment (Bytes)
+------------|---------------|------------------
+wchar_t     |  4            |  4
+wint_t      |  4            |  4
+
+## <a name=linux-c-type-representations></a> Linux-specific C type representations
+
+The following definitions apply for all ABIs defined in this document. Here
+there is no differentiation between ILP32 and LP64 abis.
+
+`wchar_t` is signed.  `wint_t` is unsigned.
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -294,16 +294,13 @@ While various different ABIs are technically possible, for software
 compatibility reasons it is strongly recommended to use the following
 default ABIs for specific architectures:
 
-  * **on RV64G**: LP64D, with floats and doubles passed in floating point
-    registers (i.e. ELFCLASS64 and EF_RISCV_FLOAT_ABI_DOUBLE).
+  * **on RV64G**: [LP64D](#abi-lp64d)
 
-    Although RV64GQ systems can technically use EF_RISCV_FLOAT_ABI_QUAD,
-    it is strongly recommended to use EF_RISCV_FLOAT_ABI_DOUBLE on
-    general-purpose RV64GQ systems for compatibility with standard RV64G
-    software.
+    Although RV64GQ systems can technically use [LP64Q](#abi-lp64q), it is
+    strongly recommended to use LP64D on general-purpose RV64GQ systems for
+    compatibility with standard RV64G software.
 
-  * **on RV32G**: ILP32D, with floats and doubles passed in floating point
-    registers (i.e. ELFCLASS32 and EF_RISCV_FLOAT_ABI_DOUBLE).
+  * **on RV32G**: [ILP32D](#abi-ilp32d)
 
 # <a name=c-types></a> C type details
 ## <a name=c-type-sizes></a> C type sizes and alignments
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -8,9 +8,12 @@
 	* [Integer Calling Convention](#integer-calling-convention)
 	* [Hardware Floating-point Calling Convention](#hardware-floating-point-calling-convention)
 	* [RV32E Calling Convention](#rv32e-calling-convention)
-	* [Default ABIs and C type sizes](#default-abis-and-c-type-sizes)
+	* [Default ABIs](#default-abis)
+3. [C type details](#c-types)
+	* [C type sizes and alignments](#c-type-sizes)
+	* [C type representations](#c-type-representation)
 	* [va_list, va_start, and va_arg](#va-list-va-start-and-va-arg)
-3. [ELF Object Files](#elf-object-file)
+4. [ELF Object Files](#elf-object-file)
 	* [File Header](#file-header)
 	* [Sections](#sections)
 	* [String Tables](#string-tables)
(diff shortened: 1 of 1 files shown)


lenary:
For clarity, these changes do not include any changes to sizes/alignments of atomic types, as raised in #112 - those changes are on hold while the x86-64 ABI change is discussed.

aswaterman:
Agreed, these are clearly properties of an ABI, not an architecture.

riscv-elf.md
```diff
@@ -241,72 +241,93 @@ s0-s1, and only three temporaries, t0-t2.
...
+    double               |  8            |  8
+    long double          | 16            | 16
+    float _Complex       |  8            |  4
+    double _Complex      | 16            |  8
+    long double _Complex | 32            | 16
+
+The alignment of `max_align_t` is 16.
+
+Structs and unions are aligned to the alignment of their most strictly aligned
+member. The size of any object is a multiple of its alignment.
```
asb: I think this isn't true of packed structs. Although the psABI might not want to over-specify behaviour in this case, it might still be worth adding an explicit acknowledgement that this statement won't hold in some cases.

lenary: I just went spelunking in the clang codebase, as well as every spec I could find.

- The C spec leaves this implementation-/un-defined. This applies to any value returned from `sizeof(…)`, but also the implementation of `#pragma packed` and the like.
- the C11 spec does require objects to have a size in bits that is a multiple of CHAR_BITS (§6.2.6.1p4). (This is unchanged in the C17 spec)
- `CHAR_BITS` is required by C11 (and C17) to be at least 8.
- This psABI defines `sizeof(char)` and `_Alignof(char)`: they are both 1 (table above). 
- This has to be the weakest alignment requirement (C11 §6.2.8p6).
- The psABIs I've found so far have not mentioned packed structs at all.
- `#pragma packed(n)` is documented pretty well by microsoft (the `n` specifies the maximal allowed alignment, in bytes). The GCC docs match this understanding.
- `#pragma packed` is the same as `#pragma packed(1)`, ie choosing to align to the nearest byte.
- C11 defines arrays as "a contiguously allocated nonempty set of objects with a particular member object type" (§6.2.5p10). This means that there is no padding between array elements. To ensure that each element of an array of structs is aligned correctly, trailing padding may have to be added to the struct itself (trailing padding on structs is allowed), to bring it to its specified alignment.
- Looking at the implementation in clang (which I expect, though have not checked, matches GCC fairly closely), packed structs have the same alignment as `char`, unless you're explicit about which alignment is requested (using `n`, described above).

Therefore:
- If you request `#pragma packed`, your struct is aligned to the same alignment as `char` (1), and so this sentence is true.
- If you request `#pragma packed(n)`, each field of your struct is aligned to at most `n` times the alignment of `char`, and so the whole struct is aligned to at most `n` times the alignment of `char` (using the sentence above). This is a case where trailing padding may have to be used so that the whole struct size (in multiples of `char`) is a multiple of `n`.

So I think the definition in this paragraph remains correct in the presence of `#pragma packed`.

lenary:
Noted about the redundancy in XLEN. Happy to update those paragraphs.

I think we're also missing a definition of `CHAR_BITS`, which I will add. It needs to be 8, I think.

riscv-elf.md
```diff
@@ -241,72 +241,93 @@ s0-s1, and only three temporaries, t0-t2.
...
-    long        |  8            |  8
-    long long   |  8            |  8
-    __int128    | 16            | 16
-    void *      |  8            |  8
-    float       |  4            |  4
-    double      |  8            |  8
-    long double | 16            | 16
+default ABIs for specific architectures:
+
+  * **on RV64G**: LP64D for XLEN=64, with floats and doubles passed in floating
```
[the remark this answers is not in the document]
lenary: Removed

riscv-elf.md
```diff
@@ -241,72 +241,93 @@ s0-s1, and only three temporaries, t0-t2.
...
-    double      |  8            |  8
-    long double | 16            | 16
-
-`char` is unsigned.  `wchar_t` is signed.  `wint_t` is unsigned.
-
-`_Complex` types have the alignment and layout of a struct containing two
-fields of the corresponding real type (`float`, `double`, or `long double`),
-with the first field holding the real part and the second field holding
-the imaginary part.
+  * **on RV32G**: ILP32D for XLEN=32, with floats and doubles passed in floating
```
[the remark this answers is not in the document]
lenary: Removed

riscv-elf.md
asb: I think I could have been convinced without such a detailed response, but I appreciate the super in-depth correction all the same! :)

asb:
Also, please squash into a single commit (or at least squash the "Removes redundant XLEN clauses" if you'd rather keep "Adds definition of `CHAR_BITS` separate).

riscv-elf.md
```diff
@@ -241,72 +241,95 @@ s0-s1, and only three temporaries, t0-t2.
...
+The alignment of `max_align_t` is 16.
+
+`CHAR_BITS` is 8.
+
+Structs and unions are aligned to the alignment of their most strictly aligned
+member. The size of any object is a multiple of its alignment.
+
+## <a name=c-type-signedness></a> C type signedness
+
+`char` is unsigned.  `wchar_t` is signed.  `wint_t` is unsigned.
```
jim-wilson: wchar_t must be compatible with the C library, and different C libraries have different definitions for wchar_t.  Vxworks for instance defines wchar_t as short unsigned int, so if we specify that wchar_t is signed, then we can never have a RISC-V vxworks port.  Cygwin defines wchar_t as unsigned short too, which suggests that MS Windows uses an unsigned short.  If we ever want a Windows port to RISC-V, then we need to allow unsigned wchar_t.  wchar_t should just not be mentioned at all.

Likewise, wint_t is C library dependent and some systems require a signed wint_t and some systems require an unsigned wint_t.  Freebsd uses a signed wint_t, and we already have a freebsd port.  So we must allow wint_t to be either signed or unsigned.  It should just not be mentioned at all.

riscv-elf.md
```diff
@@ -241,72 +241,95 @@ s0-s1, and only three temporaries, t0-t2.
...
+  * **LP64, LP64F, and LP64D**: use the following type sizes and alignments
+    (based on the LP64 convention):
+
+    Type                 | Size (Bytes)  | Alignment (Bytes)
+    ---------------------|---------------|------------------
+    bool/_Bool           |  1            |  1
+    char                 |  1            |  1
+    short                |  2            |  2
+    int                  |  4            |  4
+    wchar_t              |  4            |  4
```
jim-wilson: wchar_t is a property of the C library, and different C libraries have different definitions for wchar_t.  Some use int and some use short.  I think it should not be mentioned here.

riscv-elf.md
```diff
@@ -241,72 +241,95 @@ s0-s1, and only three temporaries, t0-t2.
...
+    wint_t               |  4            |  4
```
jim-wilson: Likewise, wint_t is a property of the C library, and different C libraries have different definitions for wint_t, with different type sizes.  I think it should not be mentioned here.

riscv-elf.md
```diff
@@ -241,72 +241,95 @@ s0-s1, and only three temporaries, t0-t2.
...
+    (based on the ILP32 convention):
+
+    Type                 | Size (Bytes)  | Alignment (Bytes)
+    ---------------------|---------------|------------------
+    bool/_Bool           |  1            |  1
+    char                 |  1            |  1
+    short                |  2            |  2
+    int                  |  4            |  4
+    wchar_t              |  4            |  4
+    wint_t               |  4            |  4
```
jim-wilson: As above, we should not specify the size of wchar_t and wint_t as they are C library dependent.

lenary:
@asb I shall squash out the XLEN change, but not the `CHAR_BITS` change.

@jim-wilson I think definitions of the signedness, size, and alignment of `wchar_t` and `wint_t` were introduced in #57 and #58, so it would be good to get approval from @aswaterman and @asb before I remove them. I'm happy to do so, I just want to ensure this is the consensus. I'll note there is no mention of them in the system-v psABI documents for intel that I have.

aswaterman:
@jim-wilson what I’m confused about is, how are they _not_ part of the ABI, given that you can’t correctly link objects that use different wchar_t disciplines? Wouldn’t it be more correct to say that it is part of the ABI, but these different C libraries implement different ABIs than each other?

jim-wilson:
> Wouldn’t it be more correct to say that it is part of the ABI, but these different C libraries implement different ABIs than each other?

Yes, that is one way to look at it.  There is no expectation that code compiled for linux will work on freebsd and vice versa for instance, so they can have different ABIs.  But we need a common base for portability between different systems, and that is the part we are defining here.  So part of the ABI is specified by the processor, and part of the ABI is specified by the C library and/or OS.  We should only be defining the processor specific parts of the ABI here, and that does not include wchar_t and wint_t.  Unless you want to start specifying C library and/or OS specific ABIs, then you could include wchar_t and wint_t in those documents.

I suppose another way to look at this is as historical error, that we didn't make things like wchar_t and wint_t processor specific at the start.  But these C libraries and operating systems are mostly written independently, and it is hard to get everyone to agree on everything before hand.  We didn't discover that we had a problem until too late.  Some of these problems happened before linux and gcc/clang started dominating development, and coordination was a lot harder back then.  When everyone has their own compiler and OS sources that are different than everyone else, it is easy for different systems to end up with different types for things like wchar_t and wint_t.  But maybe someday in the future these wchar_t and wint_t problems will be fixed.

aswaterman:
Makes sense. Maybe we should move wchar_t and friends to another table at the bottom of this document (“Additional Linux ABI Types” or something) so this table can be “Base ABI Types” or something.

asb:
I think it's definitely worth continuing to indicate the properties of `wchar_t` and co under "standard" or "Linux" ABIs - it's useful to compiler authors and there's not much point in moving out to a separate document. Putting it under "Additional Linux ABI Types" sounds sensible.

riscv-elf.md
lenary: The definitions for `wchar_t` and `wint_t` have been moved into a Linux-specific section.

lenary:
I have:
- Squashed the XLEN commit
- Fixed up the table of contents and reworded the first commit message to finish the unfinished sentence.
- Introduced the linux-specific ABI section (see the final commit in the branch)
- Moved the definitions for `wchar_t` and `wint_t` into the linux-specific ABI section.

lenary:
I have added a commit to name some specific ABIs. These should match commonly accepted definitions currently used in GCC/clang (though I know clang does not yet support all of them).

riscv-elf.md
```diff
@@ -235,18 +236,55 @@ type would be passed.
...
 provided they hold values no more than FLEN bits wide.
 
-## <a name=rv32e-calling-convention></a> RV32E Calling Convention
+## <a name=ilp32e-calling-convention></a> ILP32E Calling Convention
 
-The RV32E calling convention is designed to be usable with the RV32E ISA.
-This calling convention is the same as the integer calling convention,
-except for the following differences.  The stack pointer need only be aligned
-to a 32-bit boundary.  Registers x16-x31 do not participate in the ABI, so
+The ILP32E calling convention is designed to be usable with the RV32E ISA. This
```
lenary: I renamed this ABI to "ILP32E", because that is the ABI name, not the ISA name.

riscv-elf.md
```diff
@@ -235,18 +236,55 @@ type would be passed.
...
+* **LP64**: Integer calling-convention only, hardware floating-point calling
+  convention is not used (i.e. ELFCLASS64 and EF_RISCV_FLOAT_ABI_SOFT).
+
+* **LP64F**: LP64 with hardware floating-point calling convention for FLEN=32
+  (i.e. ELFCLASS64 and EF_RISCV_FLOAT_ABI_SINGLE).
+
+* **LP64D**: LP64 with hardware floating-point calling convention for FLEN=64
+  (i.e. ELFCLASS64 and EF_RISCV_FLOAT_ABI_DOUBLE).
+
+* **LP64Q**: LP64 with hardware floating-point calling convention for FLEN=128
```
lenary: This combination is noted under the "Default ABIs" section, so I decided to give it an explicit name.

lenary:
These two commits do two things:
- Add `<a name=>`s so we can link to the named ABIs.
- Reword the Default ABIs in terms of the named ABIs.

aswaterman:
Any objections to my merging this PR?

aswaterman:
@lenary can you also update the PR to include your name in the list of contributors at the top of the file, and perhaps reformat it to add some line breaks :)

lenary:
@aswaterman Done. [Rendered](https://github.com/lenary/riscv-elf-psabi-doc/blob/lenary/c-types-alignments/riscv-elf.md)

This uses the "two spaces before a newline" linebreak in markdown, so stripping all whitespace on save will remove the linebreaks. Happy to change it to a bulleted list if this trailing whitespace will be problematic.

aswaterman:
Thank you. I’m going to pull the trigger; if there are any lingering concerns, please address with follow-up PRs/issues.

PR #121, opened 2019-10-29
Add DWARF register number information for CSRs

T-J-Teru:
There are situations where it might be necessary to reference some
CSRs from the DWARF information.  For example, when unwinding from an
exception handler it could be necessary to reference 'mepc' as the
return address from the exception handler.

Rather than just add DWARF registers numbers for specific CSRs one at
a time as we need them, I'd like to propose that we allocate numbers
for all CSRs in one go.

When thinking how to number the CSRs I see a couple of possible
solutions, first we could allocate 4096 (the maximum number of CSRs)
DWARF register numbers and assign these to CSRs based on the CSR's
number within the specification (Volume II: Privileged Architecture).
The benefit is that the number for a CSR shouldn't need to change in
the future (I think), and adding new CSRs is easy.  The draw back is
that this means allocating a large number of DWARF register numbers,
many of which are currently not used.

A second possibility was to allocate DWARF register numbers
sequentially for the CSRs in the order they currently appear in the
spec.  This would require far fewer DWARF register numbers to be
allocated (maybe just ~250) but if we add more CSRs to the spec later
then these would need to be added to the end of the DWARF register
number list, and things would get very confused.  To try and avoid
this confusion we would probably have to list all of the CSRs in this
document.

One last detail that I wasn't sure about is if we should reserve some
register numbers, say 65 -> 100 for future as special purpose
registers, much as 64 is currently used as the alternative frame
return column.  Maybe we wouldn't need to set aside that many, but it
might be useful if any future special purpose registers didn't have to
be placed at 4160+.

In this patch I am suggesting that we take the first approach.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -868,10 +868,17 @@ Dwarf Number  | Register Name | Description
 0-31          | x0-x31        | Integer Registers
 32-63         | f0-f31        | Floating-point Registers
 64            |               | Alternate Frame Return Column
+65 - 3071     |               | Reserved for future standard extensions
+3072 - 4095   |               | Reserved for custom extensions
+4096 - 8191   |               | CSRs
 
 The alternate frame return column is meant to be used when unwinding from
 signal handlers, and stores the address where the signal handler will return
 to.
+
+There space for 4096 CSRs.  Each CSR is assigned a DWARF register
+number corresponding to its CSR number given in **Volume II: Privileged
+Architecture** of **The RISC-V Instruction Set Manual** plus 4096.
 
 # <a name=linux-abi></a> Linux-specific ABI
 


jim-wilson:
The vector extension adds 32 registers.  I'm not tracking all of the extensions, so I don't know if other extensions are adding registers.  Most of them wouldn't.  The P (packed) extension is one that might possibly add registers, but the spec I just found says it is using the GPRs so apparently no new registers.  In my gcc vector port in progress I numbered the vector regs from 96 to 127 as that made the gcc internal tables easier to maintain, but it is trivial to map those numbers to any others for the DWARF support.

I think we should allocate DWARF numbers for all of the CSRs at once, so that we only need to do this once.

Another option might be to place the CSRs at 4096 to 8191, that would make the arithmetic easier, and would leave lots of room for expansion of user registers, but if that is inconvenient for the debuggers it might not be worth the hassle.  I think it is just a simple mapping from the DWARF number scheme to the gdb (for instance) internal number scheme, so I don't think this would cause trouble for the debugger.  8191 still fits in two bytes in leb128 encoding so that shouldn't affect the DWARF debug info size.

T-J-Teru:
Jim,

Thanks for the feedback.  I think your suggestion sounds very sensible, so I updated the patch accordingly.

One thought I did have was maybe we should officially allocate some of the gap between 96 and 4095 for custom / private extensions? So maybe we set aside 65 to 3071 for future architecture extensions, and then 3072 to 4095 for custom / private extensions.  That way we help users of the ISA avoid creating tools that use a DWARF register that conflicts with say, the vector extensions when they come out.

jim-wilson:
The custom extensions idea sounds good to me.

aswaterman:
Agreed, it seems prudent to reserve some space for custom extensions.

riscv-elf.md
```diff
@@ -868,11 +868,25 @@ Dwarf Number  | Register Name | Description
 0-31          | x0-x31        | Integer Registers
 32-63         | f0-f31        | Floating-point Registers
 64            |               | Alternate Frame Return Column
+65 - 3071     |               | Reserved set 1
+3072 - 4095   |               | Reserved set 2
```
aswaterman: Following the nomenclature in the ISA spec, we should call these "Reserved for future standard extensions" and "Reserved for custom extensions" instead of set 1/2.  Then I think we can delete the two paragraphs below that explain what set 1/2 mean.

T-J-Teru: Thank you.  I suspected there would be better words I could use!  I've updated the patch inline with your suggestion.

PR #129, opened 2020-01-31
Comments on combining int+float conventions

knightsifive:
There are two lines in the integer calling convention (concerning floating-point values) which are understood to change when additionally using the hardware floating-point calling convention. I propose adding parentheticals to highlight these instances, to avoid apparent contradiction.


aswaterman:
Thanks!

PR #131, opened 2020-02-05
Add dynamic relocation R_RISCV_IRELATIVE

MaskRay:
It is used by non-preemptible IFUNC.

---

I will add non-preemptible IFUNC support to lld  https://reviews.llvm.org/D74022


jim-wilson:
We have no ifunc support in the GNU toolchain as yet.  It should be added.

Every target with ifunc support has a IRELATIVE reloc, so we will need one too.  This looks reasonable to me.

There are two conflicting proposals adding new relocs that are also using 58, but neither one is ready to go, so they can be renumbered.  This may be an inconvenience for Andes though if they are already using reloc numbers 58 and 59 without permission from the psABI.  I know that they have a patch, but I don't know if they are actively using it or not.

MaskRay:
Is anyone actively communicating with Andes people?

kuanlinchentw:
Hi,
Andes aren't actively using relocation numbers 58 and 59.  We are free to renumber them.

PR #134, opened 2020-03-23
Document __global_pointer$

lenary:
Given that the binutils ld seems to use `__global_pointer$` and this is something that affects the correctness of the GP-relative relocations, it makes sense to mention it in the documents.

I'm not sure this is the best place for this definition, and would welcome any feedback - I felt it was most important to at least mention `__global_pointer$`

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -532,8 +532,10 @@ A              | Addend field in the relocation entry associated with the symbol
 B              | Base address of a shared object loaded into memory
 G              | Offset of the symbol into the GOT (Global Offset Table)
 S              | Value of the symbol in the symbol table
-GP             | Global Pointer register (x3)
+GP             | Value of `__global_pointer$` symbol.
 
+**Global Pointer**: It is assumed that program startup code will load the value
+of the `__global_pointer$` symbol into register `gp` (aka `x3`).
 
 ### Absolute Addresses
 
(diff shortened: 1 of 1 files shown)


lenary:
@aswaterman hopefully I have addressed your comments. I realised that actually `x3` had little to do with the address calculation, so updated the table to only reference the symbol.

We could define `GP` (the relocation calculation symbol, and also the value of the register during runtime) to be zero if `__global_pointer$` is undefined (matching the GNU ld implementation), or we could leave it undefined (given lld does something slightly different). I'm happy for that clarification to happen in a follow-up PR.

aswaterman:
I don't believe defining it to be 0 is necessary--as I replied on that mailing-list thread, `ld` does _not_ assume it's 0 if undefined.  The use of 0 inside of the BFD code is as a sentinel value to disable relaxation to `gp`.  (For addresses near zero, we relax to `x0` instead.)

It's also probably undesirable to add such a constraint, because some special programs (e.g. JITs) might be able to use `x3` for another purpose if its value is unspecified when `__global_pointer$` is not defined.

riscv-elf.md
```diff
@@ -558,6 +560,21 @@ The following assembly and relocations show loading an absolute address:
...
+immediate offset from `__global_pointer$`, then the address can be loaded with a
+single instruction which has one relocation, a `R_RISCV_GPREL_I` or
+`R_RISCV_GPREL_S`.
+
+The instruction is an I-Type instruction (add immediate or load) with an
+`R_RISCV_GPREL_I` or an S-type instruction (store) with an `R_RISCV_GPREL_S`
+relocation. The following assembly show loading a gp-relative address:
+
+```
+     addi a0, gp, 0          # R_RISCV_GPREL (symbol)
```
aswaterman: This should be R_RISCV_GPREL_I, right?

lenary: the `R_RISCV_LO12` above does not use the suffix, but now I see that looks more like an omission. Update incoming.

aswaterman:
thanks!

PR #138, opened 2020-03-25
Stack slots are at least XLEN-bit aligned

### LINKED ISSUE #137: Is the alignment of arguments passed on the stack specified correctly?
Currently, we write "Aggregates or scalars passed on the stack are aligned to the minimum of the object alignment and the stack alignment."

But I think that's incomplete, since `int`s/`float`s passed on the stack are spaced 8 bytes apart for RV64.  Did I miss somewhere else in the text that clarifies that primitives have XLEN-bit alignment when passed on the stack?

For example, `i` and `j` are spaced 8 bytes apart in this example:

```
int foo(int a, int b, int c, int d, int e, int f, int g, int h, int i, int j)
{
  return i + j;
}

foo:
	lw	a0,0(sp)
	lw	a5,8(sp)
	addw	a0,a0,a5
	ret
```

aswaterman:
And, for integral arguments, the unused space is not undefined.

Resolves #137

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -99,6 +99,8 @@ register, or on the stack by value if none is available.
 When passed in registers or on the stack, integer scalars narrower than XLEN
 bits are widened according to the sign of their type up to 32 bits, then
 sign-extended to XLEN bits.
+When passed in registers or on the stack, floating-point types narrower than
+XLEN bits are widened to XLEN bits, with the upper bits undefined.
 
 Scalars that are 2✕XLEN bits wide are passed in a pair of argument registers,
 with the low-order XLEN bits in the lower-numbered register and the high-order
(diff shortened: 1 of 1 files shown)


aswaterman:
Added another commit to clarify FP arg behavior, too.

riscv-elf.md
```diff
@@ -119,8 +122,8 @@ available, the aggregate is passed on the stack. Bits unused due to
 padding, and bits past the end of an aggregate whose size in bits is not
 divisible by XLEN, are undefined.
 
-Aggregates or scalars passed on the stack are aligned to the minimum of the
-object alignment and the stack alignment.
+Aggregates or scalars passed on the stack are aligned to the greater of the
+type alignment and XLEN bits, but never more than the stack alignment.
```
lenary: Am I right in thinking this means in ILP32E, everything is aligned to 32-bits (the stack alignment == XLEN)? This seems like it could cause issues.

jim-wilson: Yes, it causes issues.  This is one of the reasons why we don't have rv32efd toolchain support.  The plan is to fix this with a new embedded ABI, which uses the same calling convention for rv32e and rv32i, and has changes to fix the stack alignment issues, and has changes to reduce the number of registers that must be saved by an interrupt handler.  But this discussion has been stalled for about 2 years or so.

PR #142, opened 2020-04-24
Fix lo12 calculation of PC-Relative Symbol

laokz:
Just like `Absolute Addresses` example, it should shift hi20 back to calculate lo12.


aswaterman:
Thanks.

PR #153, opened 2020-06-19
riscv-elf.md: add description of the code models

ebahapo:
Add a brief descritption of the existing code models.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -331,20 +331,34 @@ via its literal.  And, when not, other data structures are used to help the
 code to address the memory space.  The coding conventions governing their use
 are known as code models.
 
-### Medium low
+### Small
 
-The medium low code model, or `medlow`, allows the code to address the lower
-2 GiB of the address space.  By using the instruction `lui` and `ld` or `addi`
-instructions, for example, a 32-bit address literal can be produced.
+The small code model, or `medlow`, allows the code to address the whole RV32
+address space or the lower 2 GiB of the RV64 address space.
+By using the instructions `lui` and `ld` or `st`, when referring to an object, or
+`addi`, when calculating an address literal, for example,
+a 32-bit address literal can be produced.
 This code model is not position independent.
 
-### Medium any
+### Medium
 
-The medium any code model, or `medany`, allows the code to address the range
-between -2 GiB and +2 GiB from its position.  By using the instruction `auipc`
-and `st` or `addi` instructions, for example, a signed 32-bit offset, relative
-to the value of the `pc` register, can be produced.  This code model is
-position independent.
+The medium code model, or `medany`, allows the code to address the range
+between -2 GiB and +2 GiB from its position.  By using the instructions `auipc`
+and `ld` or `st`, when referring to an object, or
+`addi`, when calculating an address literal, for example,
+a signed 32-bit offset, relative to the value of the `pc` register,
+can be produced.
+This code model is position independent.
+
+### Compact
+
+The compact code model allows the code to address the whole 64-bit address space,
+especially when code and data are located far apart.  By using the Global
+Offset Table, or GOT, to hold the 64-bit address literals, any memory position
+can be referred.  By using the instructions `lui` and `addi`, a signed 32-bit
+offset, relative to the value of the `gp` register, can be produced, referring
+to address literals in the GOT.  This code model is position independent.
+Does not apply to the ILP32 ABIs.
 
 # <a name=c-types></a> C type details
 ## <a name=c-type-sizes></a> C type sizes and alignments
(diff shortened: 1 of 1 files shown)


riscv-elf.md
```diff
@@ -320,6 +321,31 @@ default ABIs for specific architectures:
...
+memory position using a literal as its argument.  Rather, instructions exist
+that, when combined together, can then be used to refer to a memory position
+via its literal.  And, when not, other data structures are used to help the
+code to address the memory space.  The coding conventions governing their use
+are known as code models.
+
+### Medium low
+
+The medium low code model, or `medlow`, allows the code to address the lower
+2 GiB of the address space.  By using the instruction `lui` and `ld` or `addi`
```
jim-wilson, line 337: This says lui and ld, but the medany description says auipc and st.  Not clear if it is useful or confusing for the ld/st difference here.

riscv-elf.md
```diff
@@ -320,6 +321,31 @@ default ABIs for specific architectures:
...
+address space.  There is no single instruction that can refer to an arbitrary 
+memory position using a literal as its argument.  Rather, instructions exist
+that, when combined together, can then be used to refer to a memory position
+via its literal.  And, when not, other data structures are used to help the
+code to address the memory space.  The coding conventions governing their use
+are known as code models.
+
+### Medium low
+
+The medium low code model, or `medlow`, allows the code to address the lower
```
jim-wilson, line 336: This says lower 2GiB.  The medany description says +/- 2GiB which is inconsistent.  The med low case is actually low 2GiB and upper 2GiB except that I don't know of any hardware that supports negative addresses in rv64, which would require 64 address lines, so for rv64 it is only the low 2GiB we support, where as for rv32 we support the entire 32-bit address range.  Not clear if it helps or confuses to go into the details though.

riscv-elf.md
```diff
@@ -320,6 +321,31 @@ default ABIs for specific architectures:
...
+2 GiB of the address space.  By using the instruction `lui` and `ld` or `addi`
+instructions, for example, a 32-bit address literal can be produced.
+This code model is not position independent.
+
+### Medium any
+
+The medium any code model, or `medany`, allows the code to address the range
+between -2 GiB and +2 GiB from its position.  By using the instruction `auipc`
```
jim-wilson, line 344: While the instructions do support +/-2GiB.  The code model itself only supports 2GiB objects.  We need to support the case where the first byte references the last byte, and the last byte references the first byte, and hence we can't support in general executables larger than 2GiB total.  Anything larger than 2GiB might fail with out of range reloc errors and hence isn't officially supported.  Again, I'm not sure whether the details help or confuse.

lenary:
It would be great to:
- note the LLVM names for these (small, medium)
- show the function calling sequences explicitly

riscv-elf.md
ebahapo: This is about `medlow` here...

ebahapo:
> It would be great to:
> 
> * note the LLVM names for these (small, medium)
> * show the function calling sequences explicitly

Can you please point to where these remarks apply?

riscv-elf.md
```diff
@@ -320,6 +321,31 @@ default ABIs for specific architectures:
...
+
+The RISC-V architecture constrains the addressing of positions in the
+address space.  There is no single instruction that can refer to an arbitrary 
+memory position using a literal as its argument.  Rather, instructions exist
+that, when combined together, can then be used to refer to a memory position
+via its literal.  And, when not, other data structures are used to help the
+code to address the memory space.  The coding conventions governing their use
+are known as code models.
+
+### Medium low
```
lenary, line 334: The name for this in clang is 'small' (though clang accepts 'medlow').

riscv-elf.md
```diff
@@ -320,6 +321,31 @@ default ABIs for specific architectures:
...
+
+The medium low code model, or `medlow`, allows the code to address the lower
+2 GiB of the address space.  By using the instruction `lui` and `ld` or `addi`
+instructions, for example, a 32-bit address literal can be produced.
+This code model is not position independent.
+
+### Medium any
```
lenary, line 341: The name for this in clang is 'medium' (though clang accepts 'medany').

riscv-elf.md
ebahapo: Methinks that this is implied in the description above, albeit perhaps too implicitly.

riscv-elf.md
jim-wilson: I was trying to make a point that one example uses ld and the other example uses st.  It isn't clear why they are different, and whether that is irrelevant or potentially confusing. But I think this is a very minor point.

jim-wilson: Oh, I see you fixed it.  Nevermind.

riscv-elf.md
jim-wilson: Yes, I think it is a bit too implicit, but it is a minor point.

ebahapo:
Ping! 🛎

PR #154, opened 2020-06-19
riscv-elf.md: add new definitions for the compact code model

ebahapo:
Add the basic structures to support the compact code model.


bluewww:
Type there
`### Meidum`

riscv-elf.md
```diff
@@ -320,6 +321,41 @@ default ABIs for specific architectures:
 
   * **on RV32G**: [ILP32D](#abi-ilp32d)
 
+## <a name=code-models /> Code models
```
jim-wilson, line 324: This is duplicating the text of pr #153

riscv-elf.md
```diff
@@ -441,6 +477,8 @@ rules about 2✕XLEN aligned arguments being passed in "aligned" register pairs.
   * EF_RISCV_RVE (0x0008): This bit is set when the binary targets the E ABI.
   * EF_RISCV_TSO (0x0010): This bit is set when the binary requires the RVTSO
     memory consistency model.
+  * EF_RISCV_COMPACT (0x0020): This bit is set when the binary targets the
+    compact code model.
```
jim-wilson, line 481: Why do we need an elf header flag for this?  There are a limited number of flag bits, and we will need to be careful not to use them unnecessarily.  If we do need a flag, maybe we should set aside a group of bits for a code model field instead of one bit per code model?   If we consder Maciej's FDPIC to be a code model, then we have 2 supported and 2 proposed.

riscv-elf.md
```diff
@@ -469,65 +507,76 @@ Global Offset Table or DWARF meta data.
...
+51   | R_RISCV_RELAX             | Instruction pair can be relaxed |
+52   | R_RISCV_SUB6              | Local label subtraction         |
+53   | R_RISCV_SET6              | Local label subtraction         |
+54   | R_RISCV_SET8              | Local label subtraction         |
+55   | R_RISCV_SET16             | Local label subtraction         |
+56   | R_RISCV_SET32             | Local label subtraction         |
+57   | R_RISCV_32_PCREL          | PC-relative reference           | _word32_    | S + A - P
+58   | R_RISCV_IRELATIVE         | Runtime relocation              | _wordclass_ | `ifunc_resolver(B + A)`
+59   | R_RISCV_64_PCREL          | PC-relative reference           | _word64_    | S + A - P
+60   | R_RISCV_GPREL_HI20        | GP-relative reference           | _U-type_    | S + A - GP  | `%gprel_hi(symbol)`
```
jim-wilson, line 568: Do we have an implementation of the compact model yet?  I'm not aware of one.  If not, then I think it is premature to allocate relocations for it when we don't even know if it is workable yet.

riscv-elf.md
lenary: I’m likely to propose something around ROPI/RWPI soon, which will need a bit too (It’s subtly different to both Compact and FDPIC. 

On further thought, maybe there should be a Bit dedicated to whether GP has to be preserved or not (Something Compact and FDPIC both require, and I don’t expect ROPI/RWPI to need), and maybe the rest of compatibility is down to the relocations present in the ELF?

ebahapo: When using LTO, the compiler will have to know what is the desired code model.  A case could be made about `medlow` and `medany` as well.

riscv-elf.md
ebahapo: No, there is not working prototype yet.  Should the numbers be omitted?

riscv-elf.md
lenary: Ah, that's fair (though the LLVM LTO implementation is going to use module metadata when I get around to it).

riscv-elf.md
jim-wilson: I haven't maintained an ABI before, so I don't know if there is existing practice.  I would like to avoid holes in the numbering scheme though, and if we don't have a working accepted implementation, then we can't be sure that this list of relocs is sufficient and necessary.  Though that suggests that maybe we shouldn't add the reloc list yet, or maybe mark them as proposed and put them in a separate list.

riscv-elf.md
jim-wilson: FYI medlow and medany can be linked together.  The result will be medlow.  But we don't track that anywhere.  You just get a link error if you try to put medlow code above 0x80000000 in a 64-bit address space.

jim-wilson: I'm not an expert on the GNU LTO implementation, but just trying a simple check, I see that there is a .gnu.lto_.opts section that contains the command line options.  Though the fact that different toolchains may be configured with different default code models or ABIs could perhaps be an issue.  I don't know enough to be sure.  But if we do need to know the defaults, we could forcibly add them to the option list.  We already have arch/abi in the list, so we would just need to expose the code model in the option list.

ebahapo: Which tools fills in the `.gnu.lto_.opts` section, the compiler (or assembler) or the linker?  I assume that it's the compiler (or assembler), so that at LTO time the objects a compiled again with the same options.  In this case, it's important to at least flag that the code model specified at link time is not the same as at compile time.

jim-wilson: You can't specify code model to GNU ld, nor arch or abi.  They only come from object files.  But yes, if object files have arch abi or code models that aren't compatible then the linker should complain.  Currently we have only medany and medlow, and they are compatible in the sense that they can be linked together and produce medlow output, so code models aren't tracked by the toolchain.  That will change when we have more than two code models, so we will need to do something to fix that.  So we probably need 3 or 4 bits in elf header flags for the code model.  We will need unspecified (for backward compatibility), medlow, medany, compact, fdpic, ropi/rwpi, so we have 6 already.  Unless perhaps we can determine code model from the relocations used, or base link compatibility decisions on conflicts between individual relocations.  We should probably look at how other targets handle this problem.

GCC (e.g. cc1) creates the .gnu.lto_.opts section.  It adds in the default options and the user specified options.  arch and abi are default options, but code model is not, currently.

ebahapo: The compact code model is compatible with the others, which would also make the result medium or small.  But I doubt that it's the case with the upcoming FDPIC and RWOPI.  Since RISC-V is also an embedded ISA, it'd need not 3 but 4 bits to encode this with some room for the future.  Yet, just 4 out of 32.

LTO should preserve the choices made at compile time, wherever this information is stored.

lenary: I might be out of date, but I thought the compact code model had to save/restore `gp` at function entry/exit (because you might be returning to a function in a different elf object). This would make it incompatible with medlow/medany, surely?

ebahapo: Only in a DSO.  But it would still be compatible, since both the small and the medium code would preserve the `gp` as well.

ebahapo: Actually, the small and medium code models should save, change and restore the `gp` in a DSO as well in order to access globals of local scope with relaxed code.

sorear: Anything which temporarily modifies gp is incompatible because signal handlers can contain gp-relative references, and neither Linux, glibc, nor musl will restore the main-program gp before invoking a signal handler.  You cannot link anything which writes to gp with medany/medlow code.

ebahapo: That's a good point.  If the signal handlers are built using the compact code model, there should be no problem.  But, if built using the small or medium code models, then the only way to avoid any issue would be to disable relaxation for them.

lenary: @ebahapo In medlow/medany, GP is not currently saved around calls, even for cross-DSO calls. Is this merely because `__global_pointer$` is only defined in the main executable (and therefore the gp-relative relaxations won't happen for shared objects), or am I missing something?

My understanding from when we discussed this before, is that in the compact code model, `gp` *has* to be saved around calls to extern functions (I think it might also have to be saved around local calls, because you don't know if a local call tail-calls into an external function). I could be confusing this with FDPIC though, so am I missing a detail as to how the compact model works?

ebahapo: @lenary, that's right, no relaxation for small and medium DSOs because, if they defined their own  `__global_pointer$`, the `gp` would have to be preserved and set by all dynamic functions.

Actually, only compact DSOs would have to preserve and set the `gp`.  Executables could share the same `__global_pointer$`.

sorear:
Having now read the whole proposal, I rather dislike it.  I don't think this is the appropriate solution for its stated objectives.  I apologize for the belatedness of this reply.

---

For its minimal stated objectives, the proposal is unnecessarily complicated and unnecessarily incompatible with existing code.  We can define a viable `-mcmodel=large` that requires _zero_ new relocations, changes to existing relocations, or ELF flags, and is completely linking-compatible with the default code model.  (I would argue that this is the essential difference between a "code model" and an "ABI": linking code from two code models succeeds, but produces the more restrictive model.)

You can access data anywhere in the address space using a pc-relative GOT entry.  The auipc + load + load sequence requires, when compressed, 10 bytes, the same as the post-relaxation version of this proposal.  A `-mcmodel=large` would allow data to be accessed anywhere by forcing all accesses to programmer-defined data objects to go through GOT references. The performance impact for most programs will be small because most programs access static data very infrequently.  Even though the large model is oriented toward access to large amounts of static data, such a large amount of data is likely to be in arrays, and arrays imply loops where the address setup can be hoisted.

Distributions compile most programs with `-fpie` for security reasons, and the current gcc and clang compilers always generate GOT references for data if `-fpie` is passed.  My proposal is no slower than `-fpie` code generation currently is.  `-fpie` could be improved (x86_64 generates non-GOT pcrel references), but the fact that it has not been is further evidence that the performance impact is not compelling for most users.

I would rather not spend any time on a relaxation scheme specific to the large model until there is clear and compelling evidence that one is needed.

"large" also implies support for code larger than 2^31 bytes.  This can be done with linker changes **only** and does not require psABI or compiler adjustments.  The linker merely needs to be taught to generate trampolines for R_RISCV_JAL and R_RISCV_CALL relocations which are out of range, and to generate multiple GOTs if GOTPCREL relocations occur over a sufficiently large range.  Trampolines are implemented as `auipc; l[wd]; jr` — the same as a PLT entry.  ld.bfd has support for trampoline generation on several architectures, including arm, and I believe that it supports multiple GOTs on some architectures (possibly powerpc?) but I have not confirmed this.  Trampoline support has already been added to the (non-ELF) Go linker.

Binaries with more than 2GB of text are even more of a niche case than binaries with more than 2GB of data, so we do not need to implement this in the linkers immediately.  However, this needs to be regarded as a **linker limitation** and not a **code model limitation**; you are using the correct large code model, you just need to add missing code to the linker.  As such, the code model should be called `large`.

---

The proposal as written appears to mix up three things:

* A true large code model, albeit misnamed

* Accessing all writable data in the main executable, including the GOT, via `gp` and `__global_pointer$` — in other words, RWPI

* Accessing data in shared libraries by first using a PC-relative large model access to materialize `__global_pointer$` in a register, and then using RWPI-like accesses relative to the materialized value

I believe that both a large code model and RWPI should be implemented, but as they are orthogonal they should be proposed, reviewed, implemented, and enabled _separately_, not combined in a single "compact model" PR.

I am not convinced at this time that the third part of your proposal makes sense as a thing to do.  There is no savings of bytes, instructions, or loads in shared-library functions which access any number of interposable globals or one non-interposable global.  Saving loads requires a function to access two non-interposable globals, saving instructions would require a function to access two non-interposable globals which successfully reach the small data region.  Both cases are much less likely than a function accessing one global, which is already an uncommon case.

The third piece as proposed also has a major ABI compatibility issue since it uses `gp` without restoring the main-program value prior to invoking signal handlers or callback functions.  However this is easily fixed by using any call-clobbered register instead of `gp`.  The psABI text implies that GPREL_I and GPREL_S relocations can be used with any base register as long as that base register contains `__global_pointer$`, so it doesn't need to be an ABI definition, it can be allocated by the compiler.  This should be made explicit in the psABI.

(I checked what ld.bfd does in this case.  ld.bfd has an undocumented feature where GPREL relocations can be used to access addresses near zero, and overwrites the rs1 field with either gp or zero.  Since the ld.bfd behavior can be argued to violate the psABI, is undocumented, and is not used by compilers, I would argue that it is a bug in ld.bfd and should be changed.  It would of course be possible to define new versions of the GPREL relocations with unambiguous semantics, but the relocation type field is only 8 bits on ELFCLASS32 and they should not be used frivolously.)

ebahapo:
> For its minimal stated objectives, the proposal is unnecessarily complicated and unnecessarily incompatible with existing code. We can define a viable `-mcmodel=large` that requires _zero_ new relocations, changes to existing relocations, or ELF flags, and is completely linking-compatible with the default code model. (I would argue that this is the essential difference between a "code model" and an "ABI": linking code from two code models succeeds, but produces the more restrictive model.)

This proposal does not attempt to address the situation when both text and data are more than 2GiB in size.  Rather, when text and data are more than 2GiB apart.  It just so happens that this compact code mode also supports data more than 2GiB in size.

> You can access data anywhere in the address space using a pc-relative GOT entry. The auipc + load + load sequence requires, when compressed, 10 bytes, the same as the post-relaxation version of this proposal. A `-mcmodel=large` would allow data to be accessed anywhere by forcing all accesses to programmer-defined data objects to go through GOT references. The performance impact for most programs will be small because most programs access static data very infrequently. Even though the large model is oriented toward access to large amounts of static data, such a large amount of data is likely to be in arrays, and arrays imply loops where the address setup can be hoisted.

If the GOT is more than 2GIB away from the PC, then the existing relocations overflow and cannot be used to reach it.

> The proposal as written appears to mix up three things:
> 
> * A true large code model, albeit misnamed

Truly, a compact code model.

> * Accessing all writable data in the main executable, including the GOT, via `gp` and `__global_pointer$` — in other words, RWPI

This is a side effect from leveraging the existing data structures meant for PIC.  A welcome one, methinks.

> * Accessing data in shared libraries by first using a PC-relative large model access to materialize `__global_pointer$` in a register, and then using RWPI-like accesses relative to the materialized value
> I am not convinced at this time that the third part of your proposal makes sense as a thing to do. There is no savings of bytes, instructions, or loads in shared-library functions which access any number of interposable globals or one non-interposable global. Saving loads requires a function to access two non-interposable globals, saving instructions would require a function to access two non-interposable globals which successfully reach the small data region. Both cases are much less likely than a function accessing one global, which is already an uncommon case.

Not sure what your point is.  DSOs can use global data with non public scope (non interposable).  However, most systems that support DSOs have VM, so this compact code model is not necessary then.  Those systems without VM that still support DSOs in unconventional ways would still benefit from this code model.
 
> The third piece as proposed also has a major ABI compatibility issue since it uses `gp` without restoring the main-program value prior to invoking signal handlers or callback functions. However this is easily fixed by using any call-clobbered register instead of `gp`. The psABI text implies that GPREL_I and GPREL_S relocations can be used with any base register as long as that base register contains `__global_pointer$`, so it doesn't need to be an ABI definition, it can be allocated by the compiler. This should be made explicit in the psABI.
> 
> (I checked what ld.bfd does in this case. ld.bfd has an undocumented feature where GPREL relocations can be used to access addresses near zero, and overwrites the rs1 field with either gp or zero. Since the ld.bfd behavior can be argued to violate the psABI, is undocumented, and is not used by compilers, I would argue that it is a bug in ld.bfd and should be changed. It would of course be possible to define new versions of the GPREL relocations with unambiguous semantics, but the relocation type field is only 8 bits on ELFCLASS32 and they should not be used frivolously.)

This is true.  For instance, DSOs could use any register to play the role of the "GP".  That would address the case of signal handlers.

sorear:
> This proposal does not attempt to address the situation when both text and data are more than 2GiB in size. Rather, when text and data are more than 2GiB apart.

> However, most systems that support DSOs have VM, so this compact code model is not necessary then.

If your goal is to support statically linked programs with code and _all_ writable data more than 2GiB apart, what you want is RWPI, which is a standard concept with well-understood implications and interactions with other standard concepts.

> Those systems without VM that still support DSOs in unconventional ways would still benefit from this code model.

Supporting _shared_ objects in a system without VM means that a single copy of the text can access writable data at several different addresses, which means that the offset from text to writable data cannot be a compile time constant and cannot be loaded from a rodata address.  As such your proposal is not useful in such systems.  What is useful in such systems is a FDPIC ABI, another standard concept.

jrtc27:
Re multiple GOTs, this is supported for at least MIPS in both BFD and LLD, since MIPS originally used 16-bit offsets in its GOT which quickly overflow, so now each .o gets its own GOT and the linker then optimistically merges them as much as it can.

(There is the optional `-mxgot` flag to use 32-bit offsets, but due to the TLS entries being required to be at the end of the GOT (MIPS is weird, seriously, its GOT is terribly designed, there are strict ordering requirements) and _not_ having corresponding relocations for getting 32-bit offsets, `-mxgot` is a bit of a waste of time that solves the problem for code that doesn't use certain types of TLS but does not fully address the problem. Moreover it requires an additional compiler flag, whereas multi-GOT support is an always-on feature of the linker that "just works".)

ebahapo:
> > Those systems without VM that still support DSOs in unconventional ways would still benefit from this code model.
> 
> Supporting _shared_ objects in a system without VM means that a single copy of the text can access writable data at several different addresses, which means that the offset from text to writable data cannot be a compile time constant and cannot be loaded from a rodata address. As such your proposal is not useful in such systems. What is useful in such systems is a FDPIC ABI, another standard concept.

As far as I understand, this proposal shares structures used by the FDPIC proposal and does not interfere with it.

sorear:
> As far as I understand, this proposal shares structures used by the FDPIC proposal and does not interfere with it.

Yes, the RWPI part shares a lot of relocations with FDPIC.

However, you're claiming that _this proposal_ is useful for nommu shared libraries and I am having trouble understanding that.  If `__global_pointer__` is located at a fixed location relative to pc, and pc is in ROM or a shared text area, then `__global_pointer__` must also be in a shared area, which prevents a single library loaded into several processes from having different data in each process.

Conversely, if you're considering systems that _do not_ have multiple processes and text sharing, this is effectively the loadable kernel module situation and it can be handled using plain RWPI and a loader that works at the `.o` level.

What am I missing?

ebahapo:
> However, you're claiming that _this proposal_ is useful for nommu shared libraries and I am having trouble understanding that. If `__global_pointer__` is located at a fixed location relative to pc, and pc is in ROM or a shared text area, then `__global_pointer__` must also be in a shared area, which prevents a single library loaded into several processes from having different data in each process.

On the contrary, without MMU, this proposal is not enough for DSOs.  I did add the case in an example of how they could be supported, by preserving and setting the `gp`, until it was pointed out the issue with signal handlers, but it is not its goal.  I do believe that if signal handlers are forbidden from using the `gp`, it could work, but the goal of this proposal is to allow code to reside far away from RW data and it can live without supporting DSOs or another register is used instead of the `gp` register.

ebahapo:
Ping, please.

riscv-elf.md
```diff
@@ -455,7 +455,7 @@ rules about 2✕XLEN aligned arguments being passed in "aligned" register pairs.
 
    Bit 0 | Bit  1 - 2 | Bit 3 | Bit 4 | Bit 5 - 31
   -------|------------|-------|-------|------------
-   RVC   | Float ABI  |  RVE  |  TSO  | *Reserved*
+   RVC   | Float ABI  |  RVE  |  TSO  | _Reserved_
```
jrtc27, line 458: Unrelated

riscv-elf.md
```diff
@@ -509,65 +509,83 @@ Global Offset Table or DWARF meta data.
...
+50       | R_RISCV_TPREL_S              | TP-relative TLS LE store        | _S-Type_
+51       | R_RISCV_RELAX                | Instruction pair can be relaxed |
+52       | R_RISCV_SUB6                 | Local label subtraction         | _word6_     | V - S - A
+53       | R_RISCV_SET6                 | Local label assignment          | _word6_     | S + A
+54       | R_RISCV_SET8                 | Local label assignment          | _word8_     | S + A
+55       | R_RISCV_SET16                | Local label assignment          | _word16_    | S + A
+56       | R_RISCV_SET32                | Local label assignment          | _word32_    | S + A
+57       | R_RISCV_32_PCREL             | PC-relative reference           | _word32_    | S + A - P
+58       | R_RISCV_IRELATIVE            | Runtime relocation              | _wordclass_ | `ifunc_resolver(B + A)`
+59       | R_RISCV_64_PCREL             | PC-relative reference           | _word64_    | S + A - P
```
jrtc27, line 569: Unrelated?

riscv-elf.md
```diff
@@ -509,65 +509,83 @@ Global Offset Table or DWARF meta data.
...
+61       | R_RISCV_GPREL_LO12_I         | GP-relative reference           | _I-type_    | S + A - GP  | `%gprel_lo(symbol)`
+62       | R_RISCV_GPREL_LO12_S         | GP-relative reference           | _S-type_    | S + A - GP  | `%gprel_lo(symbol)`
+63       | R_RISCV_GPREL_ADD            | GP-relative usage               |             |             | `%gprel_add(symbol)`
+64       | R_RISCV_GPREL_LOAD           | GP-relative usage               |             |             | `%gprel(symbol)`
+65       | R_RISCV_GPREL_STORE          | GP-relative usage               |             |             | `%gprel(symbol)`
+66       | R_RISCV_GOT_GPREL_HI20       | GP-relative GOT reference       | _U-type_    | G + A - GP  | `%got_gprel_hi(symbol)`
+67       | R_RISCV_GOT_GPREL_LO12_I     | GP-relative GOT reference       | _I-type_    | G + A - GP  | `%got_gprel_lo(symbol)`
+68       | R_RISCV_GOT_GPREL_ADD        | GP-relative GOT usage           |             |             | `%got_gprel_add(symbol)`
+69       | R_RISCV_GOT_GPREL_LOAD       | GP-relative GOT usage           |             |             | `%got_gprel(symbol)`
```
jrtc27, line 579: The LOAD/STOREs aren't used at all for anything.

riscv-elf.md
```diff
@@ -696,18 +716,51 @@ in the PLT occupies two 16 byte entries:
...
+1:  auipc t0, %hi_pcrel(2f) # address of 2f
+    addi  t0, %lo_pcrel(1b)
+    ld    t2, (t0)          # difference between .got.plt - 2f
+    add   t0, t0, t2        # address of .got.plt
+    add   t0, t0, t3        # address of the function pointer
+    ld    t3, (t0)          # dereference the function pointer
+    jr    t3
+    nop
+    nop
+    nop
```
jrtc27, line 733: The number of `nop`s changes based on the pointer size.

riscv-elf.md
```diff
@@ -696,18 +716,51 @@ in the PLT occupies two 16 byte entries:
      jr     t3
 ```
 
-Subsequent function entry stubs in the PLT take up 16 bytes and load a
-function pointer from the GOT. On the first call to a function, the
-entry redirects to the first PLT entry which calls `_dl_runtime_resolve`
-and fills in the GOT entry for subsequent calls to the function:
+For the compact code model, the third entry in the PLT has a stub that
+calculates the absolute address of a function pointer in the GOT.
+It occupies three 16 byte entries:
```
jrtc27, line 721: Not enough space for RV128 with the current scheme, though arguable whether it's a meaningful combination.

riscv-elf.md
ebahapo: It's used in the first PLT entry for the compact code model.

riscv-elf.md
ebahapo: Will add the cases when they are used.

riscv-elf.md
```diff
@@ -696,18 +716,51 @@ in the PLT occupies two 16 byte entries:
...
+
+```
+1:  auipc t0, %hi_pcrel(2f) # address of 2f
+    addi  t0, %lo_pcrel(1b)
+    ld    t2, (t0)          # difference between .got.plt - 2f
```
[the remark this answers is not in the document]
ebahapo: Please, elaborate.

riscv-elf.md
ebahapo: The compact code model does not apply to RV32, unless you mean RV128.

riscv-elf.md
```diff
@@ -696,18 +716,51 @@ in the PLT occupies two 16 byte entries:
...
+For the compact code model, subsequent function entries in the PLT
+take up 16 bytes and load a function pointer from the GOT.
+On the first call to a function, the entry redirects to the first PLT entr
+which calls `_dl_runtime_resolve` and fills in the GOT entry
+for subsequent calls to the function:
+
+```
+1:  lui	t3, %hi(function@.got.plt - .got.plt) # offset to the function pointer
+    addi	t3, %lo(function@.got.plt - .got.plt)
+    jal	t1, stub@.plt
```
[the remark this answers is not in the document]
ebahapo: Shared objects do not have a valid `gp` set.

ebahapo:
Ping, pretty please.

jrtc27:
Pinging isn't going to help the fact that this is adding a whole new ABI that needs to go through thorough analysis before being declared official, especially when there are outstanding concerns described many months ago.

ebahapo:
What concerns do you believe are outstanding and how would you suggest this to be analyzed?

We have a prototype downstream and I'd be glad to inquire about sharing our results.

kito-cheng:
Our colleges has report their memory layout is:
- `rodata` and `data` is put close
- `bss` put very far from `rodata` and `data`.

And seems like compact code model can't resolve such issue, maybe we need a real large model...

ebahapo:
The cost of a large code model is... large.  When all data is up to 2GiB and share the same 2GiB range, the cost can be smaller, as in this compact code model.

jrtc27:
However, such large applications are exceedingly rare and a compact code "model" is a whole new ABI not a code model due to changing how PLTs and GP work, and thus needs much more of a reason to exist than a large code model. Distributions aren't going to be shipping two sets of libraries, for example.

kito-cheng:
I know the cost of large code model is large, the most demand on the large code model is come from the bare-metal without MMU environment, in such situation, they don't have too much choice, of cause the best solution is changing the memory layout of hardware platform, but it's hard to ask hardware guy to change things in generally...:P

kito-cheng:
Few more word about large code model, in my experience, most user who large code model is not because the program is too large, but the platform/hardware memory layout.

But that's might be my bias since most of my working experience are in the embedded world.

ebahapo:
> However, such large applications are exceedingly rare and a compact code "model" is a whole new ABI not a code model due to changing how PLTs and GP work, and thus needs much more of a reason to exist than a large code model. Distributions aren't going to be shipping two sets of libraries, for example.

I think that it this proposal does not change how either the PLT or the GP work.  It definitely expands how they work, while supporting the current way in which they work.  In other targets, code models have been supported very much in the same way as this.

As @kito-cheng pointed out, the current code models do not support bare metal embedded applications without MMU, such as code ROM far away from data RAM, for whatever hardware reason.  Distributions usually don't address embedded systems and embedded systems do rebuild the libraries to fit their needs, so what distributions do is tangential to the needs of embedded applications.

jrtc27:
> > However, such large applications are exceedingly rare and a compact code "model" is a whole new ABI not a code model due to changing how PLTs and GP work, and thus needs much more of a reason to exist than a large code model. Distributions aren't going to be shipping two sets of libraries, for example.
> 
> I think that it this proposal does not change how either the PLT or the GP work. It definitely expands how they work, while supporting the current way in which they work. In other targets, code models have been supported very much in the same way as this.
> 
> As @kito-cheng pointed out, the current code models do not support bare metal embedded applications without MMU, such as code ROM far away from data RAM, for whatever hardware reason. Distributions usually don't address embedded systems and embedded systems do rebuild the libraries to fit their needs, so what distributions do is tangential to the needs of embedded applications.

If the desire is to support embedded systems then that would be better off as part of the EABI, not the Unix psABI.

PR #155, opened 2020-06-26
riscv-elf.md: describe relaxations

ebahapo:
Add a section where the current and future suggested relaxations may be described.


ebahapo:
Ping! 🛎

riscv-elf.md
```diff
@@ -969,6 +970,20 @@ there is no differentiation between ILP32 and LP64 abis.
...
+some assumptions about the memory layout made at compile time.
+
+Some relocation types, in certain situations, indicate to the linker where this
+can happen.  Additionally, some relocation types were added to indicate to the
+linker the assciated parts of a code sequence that can be thusly simplified,
+rather than to instruct the linker how to apply a relocation.
+
+The linker should only perform such relaxations when a candidate relocation type
+is followed by the relocation R_RISCV_RELAX.
+
```
lenary, line 986: We should probably cover the `_ALIGN` relocation (which affects correctness, iirc).

ebahapo: As far as I could grok `R_RISCV_ALIGN` makes sure that the symbol is at an address aligned to the byte boundary specified in the addend, yes?

lenary: https://reviews.llvm.org/D77694 seems to point to the relocation's addend being the number of `nop`s present, and the alignment required is the next power of two higher than that? (There's a comment in the code in `relaxAlign`). The meaning here should definitely be documented if that's what it is, as that's not obvious.

I'm not sure what the object file requirements are - for instance, I think if `R_RISCV_ALIGN` is present, you have to relax those to ensure alignment ends up being correct. Conversely, I think, if you use `-mno-relax` when compiling, then the compiler has to insert the right number of nops to get the right alignment, and leaves out the relocation entirely, I think.

ebahapo: I checked what `gas` and `ld` do and the former pads the code with `nop` and add the relocation, if relaxing while the latter verifies whether the symbol is aligned to the next power of 2 greater than or equal to the byte aligned in the addend and does so if not.  

I gather that this is done in order to keep the alignments emitted by `gas` even when `ld` relaxes.

@jim-wilson, what say you>

jim-wilson: With relaxation, the assembler adds the maximum number of possible nops that may be required.  The linker does relaxation in multiple passes,  The align relocs are handled in the very last pass, and the excess nops are removed to get the desired alignment.  Since this is only a decrease in code size, nothing else relaxable can break.  This must be done last, because other relaxations may decrease code size and hence it won't be aligned anymore if other relaxations are handled after align relaxations.  There may be other ways to do this, but this is how binutils does it.

There is an edge case when moving out of rvc mode.  E.g. if you exit rvc mode and then align, we might need a 2-byte nop in norvc mode which is illegal, and we get a confusing linker error.  The error is confusing because we have only 1 align reloc, but it can be used in both 2-byte aligned rvc code and 4-byte aligned norvc code, and there is no way for the linker to know what happened here based on the reloc, so no way to know if this is a linker bug or a user bug.  There is an unresolved proposal to add another reloc to resolve the ambiguity.  There is also an assembler syntax issue here that you need to align before exiting rvc mode for correct assembly code.  But this is an ugly detail that probably doesn't need to be covered here.

jrtc27: > Since this is only a decrease in code size, nothing else relaxable can break.

That's not strictly true if you, say, relax a backwards jump to a compressed jump across output sections in pass 1 and then delete a bunch of NOPs before the jump target in pass 2.

ebahapo:
Ping! 🛎

riscv-elf.md
```diff
@@ -969,6 +970,20 @@ there is no differentiation between ILP32 and LP64 abis.
 
 `wchar_t` is signed.  `wint_t` is unsigned.
 
+# <a name=code-relax /> Code relaxation
```
jrtc27, line 973:
```suggestion
# <a name=code-relax><a/> Code relaxation
```

riscv-elf.md
```diff
@@ -969,6 +970,20 @@ there is no differentiation between ILP32 and LP64 abis.
...
+
+At link time, when all the memory objects have been resolved, the code sequence
+used to refer to them may be simplified and optimized by the linker by relaxing
+some assumptions about the memory layout made at compile time.
```
jrtc27, line 977: More a _lack_ of assumptions

riscv-elf.md
```diff
@@ -969,6 +970,20 @@ there is no differentiation between ILP32 and LP64 abis.
...
+
+Some relocation types, in certain situations, indicate to the linker where this
+can happen.  Additionally, some relocation types were added to indicate to the
```
jrtc27, line 980:
```suggestion
can happen.  Additionally, some relocation types indicate to the
```

riscv-elf.md
```diff
@@ -969,6 +970,20 @@ there is no differentiation between ILP32 and LP64 abis.
...
+linker the assciated parts of a code sequence that can be thusly simplified,
```
jrtc27, line 981:
```suggestion
linker the associated parts of a code sequence that can be thusly simplified,
```

riscv-elf.md
```diff
@@ -969,6 +970,20 @@ there is no differentiation between ILP32 and LP64 abis.
...
+rather than to instruct the linker how to apply a relocation.
+
+The linker should only perform such relaxations when a candidate relocation type
+is followed by the relocation R_RISCV_RELAX.
```
jrtc27, line 985: Followed, or that there is one for the same address as another relocation that can be relaxed?

riscv-elf.md
jrtc27: ALIGN is the most important one to document, more than the optimisations, as unlike those it's _required_ to process it (the others are just hints that can be ignored).

riscv-elf.md
ebahapo: Pessimistic assumptions?

riscv-elf.md
ebahapo: What words would you suggest, please?

jim-wilson:
Just as a general comment, I don't see anything wrong with what Evandro is adding.  It just could be improved and extended with more info.  But if we refuse to accept changes because they aren't perfect, then we might get into a situation where no patch ever gets accepted.  So maybe the comments about missing R_RISCV_ALIGN docs should be submitted as an issue, and this one gets accept as initial R_RISCV_RELAX docs that can be expanded later.

jrtc27:
Yeah that's fair. #163 encompasses both relocations; I don't know whether it's better to leave that one open or open a new one just for R_RISCV_ALIGN?

jim-wilson:
163 got a little off track and covers multiple issues of which I fixed the most important ones.  I'd suggest a new issue just for R_RISCV_ALIGN.

riscv-elf.md
```diff
@@ -558,7 +559,7 @@ Enum | ELF Reloc Type        | Description                     | Field       | C
 48   | R_RISCV_GPREL_S       | GP-relative reference           | _S-Type_    | S + A - GP
 49   | R_RISCV_TPREL_I       | TP-relative TLS LE load         | _I-Type_
 50   | R_RISCV_TPREL_S       | TP-relative TLS LE store        | _S-Type_
-51   | R_RISCV_RELAX         | Previous reloc can be relaxed   |
+51   | R_RISCV_RELAX         | Other relocation can be relaxed |.            |              | Other relocation at the same address
```
jrtc27, line 562:
```suggestion
51   | R_RISCV_RELAX         | Instruction can be relaxed      |             |              | Paired with a normal relocation at the same address
```
perhaps? The current wording reads a little clunkily.

ebahapo:
Can someone please merge this PR?

ebahapo:
Ping, please.

jim-wilson:
I believe the current status is that we need a bug report for the missing R_RISCV_ALIGN documentation and then it is OK.  I opened issue #173 for that.

PR #162, opened 2020-09-25
Add R_RISCV_SET_ULEB128 and R_RISCV_SUB_ULEB128.

kuanlinchentw:
Sorry for re-createing the pull request, because I deleted my fork of the previous pull request (https://github.com/riscv/riscv-elf-psabi-doc/pull/124).

Since the number 58 is occupied by R_RISCV_IRELATIVE. So I update relocation numbers of  R_RISCV_SET_ULEB128  and R_RISCV_SUB_ULEB128.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -566,7 +566,9 @@ Enum | ELF Reloc Type        | Description                     | Field       | C
 56   | R_RISCV_SET32         | Local label assignment          | _word32_    | S + A
 57   | R_RISCV_32_PCREL      | PC-relative reference           | _word32_    | S + A - P
 58   | R_RISCV_IRELATIVE     | Runtime relocation              | _wordclass_ | `ifunc_resolver(B + A)`
-59-191  | *Reserved*         | Reserved for future standard use
+59   | R_RISCV_SET_ULEB128   | Local label subtraction         | uleb128 = S + A
+60   | R_RISCV_SUB_ULEB128   | Local label subtraction         | uleb128 = old - S - A
+61-191  | *Reserved*         | Reserved for future standard use
 192-255 | *Reserved*         | Reserved for nonstandard ABI extensions
 
 Nonstandard extensions are free to use relocation numbers 192-255 for any


kito-cheng:
Latest version of this patch:

https://sourceware.org/pipermail/binutils/2020-January/109672.html

There is two issue left in last PR: 

1. From @jim-wilson 

> The assumption that the uleb128 address size can only decrease with relaxation is true only if both addresses are in the same section. If we are subtracting a text section address from a data section address, then that value could increase due to alignment padding between sections/segments. So I think these relocs should only be allowed when the symbols are in the same section. I think the binutils patch should be fixed to enforce this.

@kuanlinchentw I am not family with binutils code especially for linker part, did you have handle this on your latest patch?

2. From @MaskRay 

> Apologies for making a comment without reading all previous discussions. I am a bit concerned about a relocation type without a clear width. See http://sourceware.org/PR4029 for an example that oscillating .uleb128 caused a bug. I don't really read through Alan Modra's workaround but I think it is about overaligning the label immediately preceding Type Tables. It works before type tables are accessed by adding a negative offset to TTBase.

@MaskRay I think that issue is happens on assemble time, it seems like not related to linker time, and I think it's hard to describe that clear width, or maybe we could add note that and mention this should not shirk the size after relaxation?

kuanlinchentw:
> The assumption that the uleb128 address size can only decrease with relaxation is true only if both addresses are in the same section. If we are subtracting a text section address from a data section address, then that value could increase due to alignment padding between sections/segments. So I think these relocs should only be allowed when the symbols are in the same section. I think the binutils patch should be fixed to enforce this.

A uleb128 subtraction expression is only allowed both addresses in the same section currently.

> Apologies for making a comment without reading all previous discussions. I am a bit concerned about a relocation type without a clear width. See http://sourceware.org/PR4029 for an example that oscillating .uleb128 caused a bug. I don't really read through Alan Modra's workaround but I think it is about overaligning the label immediately preceding Type Tables. It works before type tables are accessed by adding a negative offset to TTBase.

The linker relaxation for uleb128 should not shrink even the value could be represented with shorter length, it should pad with 0x80 to keep the same length just as before the relaxation.

kito-cheng:
> The linker relaxation for uleb128 should not shrink even the value could be represented with shorter length, it should pad with 0x80 to keep the same length just as before the relaxation.

@kuanlinchentw could you write this down into the document?

https://github.com/riscv/riscv-elf-psabi-doc/blob/master/riscv-elf.md#code-relax

Issue #163, opened 2020-11-05
Inadequate and incorrect relocation documentation


PaulLyndonCurtis:
Hi,

In https://github.com/riscv/riscv-elf-psabi-doc/blob/master/riscv-elf.md#procedure-calls the documentation indicates that the relocation R_RISCV_RELAX is against a symbol, which is incorrect.  Relocation is against the null symbol, or the symbol is ignored according to the table above.

In https://github.com/riscv/riscv-elf-psabi-doc/blob/master/riscv-elf.md#relocations the documentation has a relocation R_RISCV_ALIGN yet there is no semantic description of this relocation when relaxation is supposed to happen.

If these could be correctly documented, it would be most excellent.

Forgive me if this is not the right place, but I need to correctly work with RISC-V ELF modules in order to correctly relocate during linking.

jim-wilson:
For R_RISCV_ALIGN, in the GNU toolchain, if linker relaxation is enabled, then the assembler emits the maximum possible number of alignment bytes that might be required and adds a R_RISCV_ALIGN reloc to the first nop that lists the total number of bytes of nops emitted.  For example, for an alignment of 16 bytes, we emit 14 bytes of nops if rvc, and 12 bytes of nops if norvc.  During linker relaxation, GNU ld handles all relaxations except R_RISCV_ALIGN first.  And then handles the R_RISCV_ALIGN separately deleting as many excess nops as necessary to get the desired alignment, where the desired alignment is the arg to R_RISCV_ALIGN rounded up to the next multiple of 2.  Since code size can only decrease at this point, no other relaxations can be broken by this.  R_RISCV_ALIGN relaxation must be done last, as doing other relaxations after R_RISCV_ALIGN would change alignment giving the wrong result.

R_RISCV_RELAX indicates that the relaxation immediately before it is relaxable.  The other fields in a R_RISCV_RELAX reloc are irrelevant. GNU as doesn't set them, and GNU ld doesn't look at them.

Would you like to volunteer to write a patch?  The poor state of the docs is mostly due to lack of volunteers to work on them.

PaulLyndonCurtis:
Hi Jim,

> For example, for an alignment of 16 bytes, we emit 14 bytes of nops if rvc, and 12 bytes of nops if not rvc [corrected]

[...]

PR #164, opened 2020-11-07
Clean up R_RISCV_CALL* and R_RISCV_RELAX docs.

jim-wilson:
R_RISCV_RELAX applies to previous reloc not to a pair of instructions.
R_RISCV_CALL and R_RISCV_CALL_PLT are now interchangeable.
R_RISCV_RELAX appears after R_RISCV_CALL* only when relaxation enabled.


jim-wilson:
This is a partial fix for problems raised in issue #163.  This fixes some minor mistakes made in one of Evandro's rewrites over the summer, and adds new language to make the two call relocs interchangeable, as LLVM and Binutils already handles them that way.

aswaterman (approved):

PR #168, opened 2021-01-07
Fix relocations fields

ebahapo:
Fix the fields for the `R_RISCV_CALL` and `R_RISCV_CALL_PLT` relocations, since they apply to two instructions, `auipc` and `jalr`.
    
Also, space out sections in order to improve legibility when editing.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -542,8 +542,8 @@ Enum | ELF Reloc Type        | Description                     | Field       | C
 11   | R_RISCV_TLS_TPREL64   | Runtime TLS relocation          | _word64_    | S + A + TLSOFFSET
 16   | R_RISCV_BRANCH        | PC-relative branch              | _B-Type_    | S + A - P
 17   | R_RISCV_JAL           | PC-relative jump                | _J-Type_    | S + A - P
-18   | R_RISCV_CALL          | PC-relative call                | _J-Type_    | S + A - P   | Macros `call`, `tail`
-19   | R_RISCV_CALL_PLT      | PC-relative call (PLT)          | _J-Type_    | S + A - P   | Macros `call`, `tail` (PIC)
+18   | R_RISCV_CALL          | PC-relative call                | _U+J-Type_  | S + A - P   | Macros `call`, `tail`
+19   | R_RISCV_CALL_PLT      | PC-relative call (PLT)          | _U+J-Type_  | S + A - P   | Macros `call`, `tail` (PIC)
 20   | R_RISCV_GOT_HI20      | PC-relative GOT reference       | _U-Type_    | G + A - P   | `%got_pcrel_hi(symbol)`
 21   | R_RISCV_TLS_GOT_HI20  | PC-relative TLS IE GOT offset   | _U-Type_    |             | Macro `la.tls.ie`
 22   | R_RISCV_TLS_GD_HI20   | PC-relative TLS GD reference    | _U-Type_    |             | Macro `la.tls.gd`
@@ -633,6 +633,7 @@ _I-Type_    | Specifies a field as the immediate field in an I-type instruction
 _S-Type_    | Specifies a field as the immediate field in an S-type instruction
 _U-Type_    | Specifies a field as the immediate field in an U-type instruction
 _J-Type_    | Specifies a field as the immediate field in a J-type instruction
+_U+J-Type_  | Specifies a field as the immediate fields in a U-type and J-type instruction pair
 
 ### Constants
 
@@ -779,7 +780,7 @@ and when `-fpic` is enabled it expands to:
 
 ### PC-Relative Jumps and Branches
 
-Unconditional jump (UJ-Type) instructions have a `R_RISCV_JAL` relocation
+Unconditional jump (U+J-Type) instructions have a `R_RISCV_JAL` relocation
 that can represent an even signed 21-bit offset (-1MiB to +1MiB-2).
 
 Branch (SB-Type) instructions have a `R_RISCV_BRANCH` relocation that


riscv-elf.md
```diff
@@ -558,7 +584,7 @@ Enum | ELF Reloc Type        | Description                     | Field       | C
 48   | R_RISCV_GPREL_S       | GP-relative reference           | _S-Type_    | S + A - GP
 49   | R_RISCV_TPREL_I       | TP-relative TLS LE load         | _I-Type_
 50   | R_RISCV_TPREL_S       | TP-relative TLS LE store        | _S-Type_
-51   | R_RISCV_RELAX         | Instruction pair can be relaxed |
+51   | R_RISCV_RELAX         | Previous reloc can be relaxed   |
```
jim-wilson, line 587: I already fixed this.  Maybe you need to update your tree?

riscv-elf.md
```diff
@@ -711,15 +739,19 @@ and fills in the GOT entry for subsequent calls to the function:
...
 
-In position dependent code (`-fno-pic`) the `AUIPC` instruction in the
+In position dependent code (`-fno-pic`), with relaxation enabled,
+the `AUIPC` instruction in the
 `AUIPC+JALR` pair has both a `R_RISCV_CALL` relocation and a `R_RISCV_RELAX`
 relocation indicating the instruction sequence can be relaxed during linking.
 
-In position independent code (`-fPIC`, `-fpic` or `-fpie`) the `AUIPC`
+In position independent code (`-fPIC`, `-fpic` or `-fpie`),
+with relaxation enabled, the `AUIPC`
```
jim-wilson, line 754: These 3 hunks look like text that is already in the current version of the document, added by one of my earlier changes.

riscv-elf.md
```diff
@@ -525,8 +551,8 @@ Enum | ELF Reloc Type        | Description                     | Field       | C
 11   | R_RISCV_TLS_TPREL64   | TLS relocation                  | _word64_    | S + A + TLS + S_TLS_OFFSET - TLS_DTV_OFFSET
 16   | R_RISCV_BRANCH        | PC-relative branch              | _B-Type_    | S + A - P
 17   | R_RISCV_JAL           | PC-relative jump                | _J-Type_    | S + A - P
-18   | R_RISCV_CALL          | PC-relative call                | _J-Type_    | S + A - P   | Macros `call`, `tail`
-19   | R_RISCV_CALL_PLT      | PC-relative call (PLT)          | _J-Type_    | S + A - P   | Macros `call`, `tail` (PIC)
+18   | R_RISCV_CALL          | PC-relative call                | _UJ-Type_   | S + A - P   | Macros `call`, `tail`
+19   | R_RISCV_CALL_PLT      | PC-relative call (PLT)          | _UJ-Type_   | S + A - P   | Macros `call`, `tail` (PIC)
```
jim-wilson, line 555: Older ISA doc revisions have an actual UJ type instruction format, so this is potentially confusing.  I'd suggest that U-and-J might be better if that doesn't mess up the table formatting too much.  Or maybe an ampersand can work.

The old UJ instruction format got renamed to J.  Though I see that binutils still has multiple UJ format references such as ENCODE_UJTYPE_IMM.  I think binutils should be fixed to stop using the UJ name in macros.  I'll submit a binutils bug report for that.  Though we still accept ".insn uj, ..." for backward compatibility and can't fix this without risking breaking old user code.

jim-wilson (commented):
Other than the fixes for stuff already fixed, this looks OK to me.

riscv-elf.md
ebahapo: How about _JU-Type_?

riscv-elf.md
ebahapo: My bad.

riscv-elf.md
jrtc27: @jim-wilson Probably worth emitting a warning for `.insn uj` if you don't already so new users of the legacy form don't appear? Though LLVM will never support that syntax (or at least I will be against it), and I imagine if you emit a warning now you can kill the support off in a few years.

As for the name, what about _U+J-type_? That makes it clearer it's a pair.

ebahapo: > As for the name, what about _U+J-type_? That makes it clearer it's a pair.

I can live with that.

ebahapo:
Ping, please.

jim-wilson (approved):

jrtc27:
Lots of unnecessary whitespace changes in this

jim-wilson:
Git is reporting patch conflicts now that I committed pr #155.  This is preventing a merge.  Can you please fix that?

riscv-elf.md
```diff
@@ -614,6 +640,7 @@ _I-Type_    | Specifies a field as the immediate field in an I-type instruction
 _S-Type_    | Specifies a field as the immediate field in an S-type instruction
 _U-Type_    | Specifies a field as the immediate field in an U-type instruction
 _J-Type_    | Specifies a field as the immediate field in a J-type instruction
+_U+J-Type_  | Specifies a field as the immediate field in an U-type and a J-type instruction
```
jrtc27, line 643: Should be "fields" plural, it's "a U-type" (two lines above is wrong) and probably best to say "instruction pair", so I'd suggest overall:
```suggestion
_U+J-Type_  | Specifies a field as the immediate fields in a U-type and J-type instruction pair
```

jrtc27:
You've got two fixup commits and a merge commit in there, those should all be squashed into the original single commit via a rebase, otherwise the history will be a tangled mess.

riscv-elf.md
```diff
@@ -1,5 +1,6 @@
 # RISC-V ELF psABI specification
 
+
```
jrtc27, line 3: Why all these whitespace changes?

ebahapo: It makes editing the file easier when the sections are thus spaced, IMO.

jrtc27: That should be an entirely separate pull request. It has no place in this one.

kito-cheng:
@jrtc27 are you OK with this PR except unnecessary whitespace change issue? it's acceptable to mixed few style fix to me :P

jrtc27:
> @jrtc27 are you OK with this PR except unnecessary whitespace change issue? it's acceptable to mixed few style fix to me :P

I disagree with all of the whitespace changes which make things double-spaced, which is the vast majority of them. There are a couple of places where things were lacking a blank line that probably should have one between them that this also fixes, but those are the minority. I think the best approach is if I just push a fix for the few whitespace issues myself and then @ebahapo rebases atop that, dropping all his whitespace changes?

kito-cheng:
> I disagree with all of the whitespace changes which make things double-spaced, which is the vast majority of them. There are a couple of places where things were lacking a blank line that probably should have one between them that this also fixes, but those are the minority. I think the best approach is if I just push a fix for the few whitespace issues myself and then @ebahapo rebases atop that, dropping all his whitespace changes?

SGTM, thanks.

jrtc27 (approved):
Thanks, looks good now

PR #174, opened 2021-03-08
Normalise number of blank lines around sections

jrtc27:
The predominant style is to have a single blank line before and a single blank line after, so fix the cases where this isn't true to match.


kito-cheng (approved):
It's trivial changes for only change the layout of blank line, so I gonna merge this.

PR #178, opened 2021-03-14
Add support for building a PDF with Pandoc

jrtc27:


riscv-elf.md
```diff
@@ -1,3 +1,47 @@
...
+
+classoption:
+- twoside
+- 11pt
+
+title: RISC-V ELF psABI specification
+subtitle: Document Version 0.1-draft
+author:
+- Kito Cheng
+- Jessica Clarke
```
jrtc27, line 12: Should we be crediting other people (Andrew, Jim, Palmer) here too, or just in the more complete list in the copyright section?

riscv-elf.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+classoption:
+- twoside
+- 11pt
+
+title: RISC-V ELF psABI specification
+subtitle: Document Version 0.1-draft
+author:
+- Kito Cheng
+- Jessica Clarke
```
jrtc27, line 11: Should we be crediting other people (Andrew, Jim, Palmer) here too, or just in the more complete list in the copyright section?

jrtc27:
[PDF for the current version](https://github.com/riscv/riscv-elf-psabi-doc/files/6137898/riscv-elf.pdf)

jrtc27:
[Now with a table of contents](https://github.com/riscv/riscv-elf-psabi-doc/files/6137928/riscv-elf.pdf)

riscv-elf.yaml
kito-cheng: I would like to list Andrew, Jim, Palmer here too, and other listed in the copyright section.

kito-cheng:
I got `pandoc: unrecognized option `--metadata-file'`, but I believe it's because my pandoc is too old, after remove I can get pdf by myself.

kito-cheng (approved):
@jrtc27 thanks you, that's LGTM, feel free to merge that yourself after update `Editors`

jrtc27:
> I got `pandoc: unrecognized option `--metadata-file'`, but I believe it's because my pandoc is too old, after remove I can get pdf by myself.

Hm, didn't realise it was a recent feature. The alternative is putting the YAML in riscv-elf.md itself, but GitHub's renderer isn't aware of the Pandoc extension so displays a rather ugly table at the start of the document containing all the options, so I don't think we can easily support older Pandoc versions and will just have to require people update if they want to build the PDF.

kito-cheng:
> so I don't think we can easily support older Pandoc versions and will just have to require people update if they want to build the PDF.

My thought is we could add that into github action to let it automatically generated latest PDF, so in theory most people don't need to care that, but put into my long TODO list at this moment, I guess I don't have too much free time until end of this month.

Issue #90, opened 2019-04-14
Should we constrain the target label of a %pcrel_lo to be in the same section?


rofirrim:
From the wording, it is unclear to me if the following is acceptable or not

```asm
.text
.globl _start

foo:
    addi a1, a0, %pcrel_lo(label)
    ret

.section .text.new_section
_start:
label:
    auipc a0, %pcrel_hi(bar)
    j foo

bar:
    ret
```

GNU ld complains with `dangerous relocation: %pcrel_lo missing matching %pcrel_hi`, so perhaps there is an underlying assumption that the reference and the label must be found in the same section?

I think the assumption is reasonable but I fail to see the text explicitly constraining this. For example, something like this

```diff
The `R_RISCV_PCREL_LO12_I` or `R_RISCV_PCREL_LO12_S` relocations contain
a label pointing to an instruction with a `R_RISCV_PCREL_HI20` relocation
entry that points to the target symbol:

 - At label: `R_RISCV_PCREL_HI20` relocation entry ⟶ symbol
- - `R_RISCV_PCREL_LO12_I` relocation entry ⟶ label
+ - `R_RISCV_PCREL_LO12_I` relocation entry ⟶ label. The reference to the label
+   and the label definition must be in the same section.
```

What do you think?

Thanks!

aswaterman:
My inclination is that this behavior should be OK, and the linker should not emit an error.  But I haven't thought through the implications on linker relaxation, which might for some reason favor the hi/lo relocs being in the same section.

jim-wilson:
​GNU ld relocates one section at a time.  It is while we are relocating a section that we match pcrel_lo relocs to pcrel_hi relocs.  In order to resolve a pcrel_lo reloc, we need both the address of the pcrel_hi reloc which is part of the pcrel_lo reloc, and the address of the symbol that the pcrel_hi reloc points to, which is part of the pcrel_hi reloc.  So we have to find the matching pcrel_hi reloc to resolve a pcrel_lo reloc.  As we scan through the section resolving relocs, we record all of the pcrel_hi and pcrel_lo relocs we see, then match them together at the end to resolve them.  And we do this one section at time.  Hence, the current GNU ld implementation requires that matching pcrel_hi and pcrel_lo relocs are in the same section.

There might be a way to fix GNU ld to make this work, but linkers are complicated, and someone would have to spend some time looking at GNU ld to figure out how to fix it.

[...]

PR #180, opened 2021-03-16
Require R_RISCV_PCREL_LO12_[IS] be in the same section as their other half

jrtc27:
This is currently required by GNU ld, which will warn if it's ever not the case, and LLD, which will give an error.

Whilst here, fix an "a" to be an "an".

Closes: #90

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -779,8 +779,8 @@ with an `R_RISCV_PCREL_LO12_I` relocation or an S-Type instruction (store)
 and an `R_RISCV_PCREL_LO12_S` relocation.
 
 The `R_RISCV_PCREL_LO12_I` or `R_RISCV_PCREL_LO12_S` relocations contain
-a label pointing to an instruction with a `R_RISCV_PCREL_HI20` relocation
-entry that points to the target symbol:
+a label pointing to an instruction in the same section with an
+`R_RISCV_PCREL_HI20` relocation entry that points to the target symbol:
 
  - At label: `R_RISCV_PCREL_HI20` relocation entry ⟶ symbol
  - `R_RISCV_PCREL_LO12_I` relocation entry ⟶ label


aswaterman (approved):

jim-wilson (approved):

riscv-elf.md
```diff
@@ -779,8 +779,8 @@ with an `R_RISCV_PCREL_LO12_I` relocation or an S-Type instruction (store)
 and an `R_RISCV_PCREL_LO12_S` relocation.
 
 The `R_RISCV_PCREL_LO12_I` or `R_RISCV_PCREL_LO12_S` relocations contain
-a label pointing to an instruction with a `R_RISCV_PCREL_HI20` relocation
-entry that points to the target symbol:
+a label pointing to an instruction in the same section with an
+`R_RISCV_PCREL_HI20` relocation entry that points to the target symbol:
```
MaskRay, line 783: While here, does it need to clarify whether a label can be reused for more than one `R_RISCV_PCREL_LO12_I`?

jim-wilson:
The GNU toolchain allows a label to be used for more than one pcrel-lo.  But this is not used by default because it is known to  fail on rare occasion due to alignment issues that can result in overflow.  I don't know of any good solution other than redesigning how the auipc instruction works which seems unlikely to happen.  Anyways, that should not prevent the linker from supporting the feature and the psabi should allow it.

```
rohan:2204$ cat tmp.c
long long i;
void sub (void) { i = 0x1234567887654321LL; }
rohan:2205$ riscv32-unknown-elf-gcc -O2 -S -mcmodel=medany -mexplicit-relocs tmp.c
rohan:2206$ cat tmp.s
	.file	"tmp.c"
	.option nopic
	.attribute arch, "rv32i2p0_m2p0_a2p0_f2p0_d2p0_c2p0"
	.attribute unaligned_access, 0
	.attribute stack_align, 16
	.text
	.align	1
	.globl	sub
	.type	sub, @function
sub:
	.LA1: auipc	a5,%pcrel_hi(.LC0)
	lw	a2,%pcrel_lo(.LA1)(a5)
	lw	a3,%pcrel_lo(.LA1+4)(a5)
	.LA0: auipc	a5,%pcrel_hi(i)
	sw	a2,%pcrel_lo(.LA0)(a5)
	sw	a3,%pcrel_lo(.LA0+4)(a5)
	ret
	.size	sub, .-sub
	.globl	i
	.section	.srodata.cst8,"aM",@progbits,8
	.align	3
.LC0:
	.word	-2023406815
	.word	305419896
	.section	.sbss,"aw",@nobits
	.align	3
	.type	i, @object
	.size	i, 8
i:
	.zero	8
	.ident	"GCC: (GNU) 10.2.0"
rohan:2207$ 
```

GNU ld adds both the pcrel_lo and pcrel_hi relocs into a hash table so that they can be matched.  It allows the pcrel_lo to come before the pcrel_hi which can happen on rare occasion with basic block reorg or some other optimization that moves code around.  it allows a pcrel_hi to match more than one pcrel_lo.

jrtc27:
Yeah, LLD will also support that case just fine, though LLVM (currently) will never generate such code. It's expressible in assembly and the ELF output, and nothing in the proposed wording says you can't do it, therefore you can. We could explicitly call it out, but I don't particularly want to encourage people writing assembly to do that kind of thing to be honest, people manage to get confused and screw up writing assembly enough as it is already.

kito-cheng (approved):
LGTM, @jrtc27 could you add a description to address @MaskRay's comment, or maybe you prefer create another PR for that, it's also OK to me too.

riscv-elf.md
jrtc27: That's already covered by "Note the compiler emitted instructions for PC-relative symbol addresses are
not necessarily sequential or in pairs." and the code example below, which uses one auipc for both an lw and an sw.

Issue #184, opened 2021-04-16
Clarify usage of addend in R_RISCV_PCREL relocations


PkmX:
The ABI doesn't define how addends in `R_RISCV_PCREL_HI20` and `R_RISCV_PCREL_LO12_[IS]` should be handled. Currently there are discrepancies between ld.bfd and lld:

* ld.bfd: The addend of `PCREL_LO12_[IS]` is ignored when looking up the `PCREL_HI20`, and addends from both parts are added together to form the final address of the symbol:

```asm
foo:
    auipc   a0, %pcrel_hi(bar + 4)
    addi    a0, a0, %pcrel_lo(foo + 4)
```

This eventually loads the address `bar + 4 + 4` into `a0`.

* lld: `PCREL_LO12_[IS]` are not allowed to have addends (throws error), but addend on `PCREL_HI20` is allowed and is taken into account for address calculation.

It would be good if this is clarified in the spec. Note that if we were to follow the implementation of binutils, we can lift the restriction on lld, but not vice-versa without breaking compatibility. Although I have to say it is probably very counter-intuitive for programmers.

aswaterman:
I think lld is getting this right and the spec should be updated accordingly.  Probably the GNU ld behavior should be changed accordingly... it would technically be backwards incompatible, but it is hard to imagine anyone is relying on this behavior for good reasons.

PkmX:
I can see the addend in `PCREL_LO12` being useful to address struct members or array as an offset in compiler generated code. However it does add a bit of complexity during relaxation as multiple `PCREL_LO12` may share a single `PCREL_HI20` but not all of them can be relaxed, so the linker needs to record whether an `auipc` has no references to it and can be safely deleted.

jrtc27:
If this were AArch64 where ADRP masks out the low bits then yes, you could do that. But because the low bits of AUIPC are left untouched, unless you have over-aligned code, you have no way of knowing at compile time whether you can reuse the PCREL_HI20 or not, regardless of the alignment of the addressed data, as the offset might overflow. So I don't think supporting such uses is ever going to have any point.

kito-cheng:
Thanks for report that, as you said allow `PCREL_LO12_[IS]` with addend would be useful for some case, but it seems dangerous to doing that, and I believe GCC and LLVM are both not rely on this behavior, so I think best way is to forbid `PCREL_LO12_[IS]` with addend and update bintuils to emit error too.

PkmX:
[...]

PR #185, opened 2021-04-19
Clarify that addend of R_RISCV_PCREL_LO12_[IS] must be 0

kito-cheng:
Address #184 

We've document the `addend` will be consider during resolve, so it's kind of
backward incompatible change, but actually we didn't generating
`R_RISCV_PCREL_LO12_[IS]` with `addend` on both LLVM and GCC.

So it's should be a safe change.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -548,8 +548,8 @@ Enum | ELF Reloc Type        | Description                     | Field       | C
 21   | R_RISCV_TLS_GOT_HI20  | PC-relative TLS IE GOT offset   | _U-Type_    |             | Macro `la.tls.ie`
 22   | R_RISCV_TLS_GD_HI20   | PC-relative TLS GD reference    | _U-Type_    |             | Macro `la.tls.gd`
 23   | R_RISCV_PCREL_HI20    | PC-relative reference           | _U-Type_    | S + A - P   | `%pcrel_hi(symbol)`
-24   | R_RISCV_PCREL_LO12_I  | PC-relative reference           | _I-type_    | S + A - P   | `%pcrel_lo(address of %pcrel_hi)`
-25   | R_RISCV_PCREL_LO12_S  | PC-relative reference           | _S-Type_    | S + A - P   | `%pcrel_lo(address of %pcrel_hi)`
+24   | R_RISCV_PCREL_LO12_I  | PC-relative reference           | _I-type_    | S - P       | `%pcrel_lo(address of %pcrel_hi)`, the addend must be 0
+25   | R_RISCV_PCREL_LO12_S  | PC-relative reference           | _S-Type_    | S - P       | `%pcrel_lo(address of %pcrel_hi)`, the addend must be 0
 26   | R_RISCV_HI20          | Absolute address                | _U-Type_    | S + A       | `%hi(symbol)`
 27   | R_RISCV_LO12_I        | Absolute address                | _I-Type_    | S + A       | `%lo(symbol)`
 28   | R_RISCV_LO12_S        | Absolute address                | _S-Type_    | S + A       | `%lo(symbol)`
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -548,8 +548,8 @@ Enum | ELF Reloc Type        | Description                     | Field       | C
 21   | R_RISCV_TLS_GOT_HI20  | PC-relative TLS IE GOT offset   | _U-Type_    |             | Macro `la.tls.ie`
 22   | R_RISCV_TLS_GD_HI20   | PC-relative TLS GD reference    | _U-Type_    |             | Macro `la.tls.gd`
 23   | R_RISCV_PCREL_HI20    | PC-relative reference           | _U-Type_    | S + A - P   | `%pcrel_hi(symbol)`
-24   | R_RISCV_PCREL_LO12_I  | PC-relative reference           | _I-type_    | S - P       | `%pcrel_lo(address of %pcrel_hi)`, the addend must be 0 (*note 1)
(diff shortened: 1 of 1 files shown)


aswaterman (approved):

kito-cheng:
Tweak little bit for commit message.

jim-wilson:
The GNU toolchain does use non-zero offsets in pcrel_lo relocs, and removing this feature is a de-optimization that will hurt performance on benchmarks like dhrystone and coremark when compiled for the medany code model.  If this is removed from the ABI, then we will likely need to retain this as a GNU toolchain extension to the ABI.

jrtc27:
> The GNU toolchain does use non-zero offsets in pcrel_lo relocs, and removing this feature is a de-optimization that will hurt performance on benchmarks like dhrystone and coremark when compiled for the medany code model. If this is removed from the ABI, then we will likely need to retain this as a GNU toolchain extension to the ABI.

It cannot be done robustly without breaking ABI or the unprivileged ISA, so it will have to remain a non-standard GNU extension. I will strongly advocate for it _not_ being supported anywhere in LLVM.

jrtc27:
> > The GNU toolchain does use non-zero offsets in pcrel_lo relocs, and removing this feature is a de-optimization that will hurt performance on benchmarks like dhrystone and coremark when compiled for the medany code model. If this is removed from the ABI, then we will likely need to retain this as a GNU toolchain extension to the ABI.
> 
> It cannot be done robustly without breaking ABI or the unprivileged ISA, so it will have to remain a non-standard GNU extension. I will strongly advocate for it _not_ being supported anywhere in LLVM.

That is, the psABI should say it MUST be 0 to conform to the ABI, and as an implementation you are free to weaken that as a GNU extension.

jim-wilson:
It is OK with me if it is a GNU extension.  It is unsafe, but there are people using it anyways, so dropping it is an issue for us.

PkmX:
If this should remain as a GNU extension there should be ways to disable that or at least it should be documented somewhere. My concern is that users may unknowingly generate non-ABI-conforming code with gcc and then fail to link with other linkers.

jrtc27:
> It is OK with me if it is a GNU extension. It is unsafe, but there are people using it anyways, so dropping it is an issue for us.

I mean, they should really stop using it, regardless of benchmark improvements; unsound optimisations are not legitimate things to benchmark. It *will* break for them unexpectedly one day, and there is nothing they can do about it other than perturb the code so things don't line up badly.

jrtc27:
> > It is OK with me if it is a GNU extension. It is unsafe, but there are people using it anyways, so dropping it is an issue for us.
> 
> I mean, they should really stop using it, regardless of benchmark improvements; unsound optimisations are not legitimate things to benchmark. It _will_ break for them unexpectedly one day, and there is nothing they can do about it other than perturb the code so things don't line up badly.

You can do what you want with GCC ultimately, but I would strongly encourage you to deprecate and remove a broken-by-design code generation option. Anything else is just a disservice to your users and gives both GCC and RISC-V a bad name.

jim-wilson:
This requires manually enabling the -mexplicit-relocs option with medany code model, and use of this option combination is not recommended.  We could update gcc docs to make this clearer.

While I agree it is a problem, there is a practical issue with telling marketing folks that they need to reduce the dhystone/coremark numbers that they are reporting.  This is unlikely to end well.  So I have to choose between two very bad options here, and I prefer the one that ends with me still receiving a paycheck.  Even if it means the compiler support is a bit ugly.  I consider this an engineering compromise.

kito-cheng:
I add a note to describe GNU toolchain support non-zero addend, but it's deprecated and will be obsoleted.

jim-wilson (approved):

kito-cheng:
I would be merge this later after double check with LLVM folks, I believe LLVM is not generate R_RISCV_PCREL_LO12_* with non-zero addend but it would be better to notify them on the RISC-V LLVM sync up meeting.

I also tried this could be accept by LLVM MC assembler, and then generate `R_RISCV_PCREL_LO12_I` with addend, so I guess we might need to 
```
i:
        .LA0: auipc     a5,%pcrel_hi(i)
        lw      a0,%pcrel_lo(.LA0)(a5)
        lw      a1,%pcrel_lo(.LA0+4)(a5)
```

kito-cheng:
Asked @MaskRay, he is happy to see this change.

kito-cheng:
Discussed with LLVM folks on RISC-V LLVM meeting @ 2021/04/29, no objection on this change.

Wait another week for make sure no further comment, this PR will be merged at 2021/05/06.

jrtc27:
> This requires manually enabling the -mexplicit-relocs option with medany code model, and use of this option combination is not recommended. We could update gcc docs to make this clearer.

Then you lose out on things like:

```
1: auipc t0, %pcrel_hi(foo)
   lw t1, %pcrel_lo(1b)(t0)
   addi t1, t1, 1
   sw t1, %pcrel_lo(1b)(t0)
```

which is a sound transformation. In my opinion, the unsound optimisation requires -mexplicit-relocs, but should be gated behind something else, as currently you conflate the sound and unsound parts.

riscv-elf.md
```diff
@@ -593,6 +593,13 @@ This section and later ones contain fragments written in assembler. The precise
 assembler syntax, including that of the relocations, is described in the
 [RISC-V Assembly Programmer's Manual](https://github.com/riscv/riscv-asm-manual).
 
+```
+NOTE:
+
+1: GNU Toolchain support `R_RISCV_PCREL_LO12_I` and `R_RISCV_PCREL_LO12_S`
+   with non-zero addend, but it's deprecated, and will be obsoleted in future.
+```
+
```
jrtc27, line 602: Please remove this part; their broken extensions are for them to document

kito-cheng: Removed.

kito-cheng:
@jrtc27 Let me know if it's OK to you now.

jrtc27 (approved):

Issue #166, opened 2020-12-15
Zfinx Calling Convention


lenary:
LLVM Recently received a patch to add Zfinx support to the compiler: https://reviews.llvm.org/D93298

@jrtc27 correctly brings up the issue that this is going to cause some ABI compatibility issues, which we need to address going forwards.

At first thought, It seems to me that Zfinx code is compatible with the current calling convention if the ABI's FLEN is zero.

Beyond that, we probably need a single bit in the `e_flags` which represents if the F registers share storage with the X registers. 

These are just very initial thoughts about the easy path forwards, in order to start a discussion. I know there are harder issues to touch on when enabling extensions that are this intrusive to how ISA state is preserved.

jim-wilson:
Perhaps instead of dedicating a bit for zfinx, we expand the current 2-bit FLOAT_ABI field to 3 bits, and use one of the 4 new encoding for zfinx.  That leaves us with 3 free encodings that can be used for other float ABIs.  There is other FP related stuff coming, like zfh and alternate FP formats like posits and bfloat.  So we might have need for another float abi later on   Kito has made the same suggestion though I don't remember exactly where.  This does have the minor downside that the 3 bits won't be contiguous which might be a little confusing.

It is valid to use -mabi=ilp32 with -march=rv32gc, but zinfx is not compatible with rv32gc/ilp32 code, so we should not consider FLOAT_ABI_SOFT to be compatible with zfinx code.

kito-cheng:
I think we have 3 decision point here, 1) calling convention, 2) ELF flag allocation and 3) Naming, any missing?

- Calling Convention 
    1. Same as soft-float.
        - Pros:
 	      - Simple and consistent
        -  Cons:
 	      - Less efficient if parameter not aligned.
 	          - double precision instruction require register align to even register.
    2. Align to even register for parameter with 2 * XLEN data.
        - e.g. double and long long on RV32 and long double and _int128 for RV64
        - Pros:
 	      - No re-align issue for parameter.
        - Cons:
 	      - More parameters passing rule
    - Example code:
        - ```void foo(int a, double b);```
            - Option i: a passed in a0, b passed in a1 and a2 for RV32
            - Option ii: a passed in a0, b passed in a2 and a3 for RV32, a1 are skipped  for padding.
- ELF Flag
    1. Allocate dedicated bit for zfinx.
        - Pros:
            - Simple?
        - Cons:
            - Waste ELF flag bits compare to other options.
[...]

PR #189, opened 2021-05-05
Draft of ZFINX ABI: ilp32x and lp64x

kito-cheng:
- In this draft, we allocated a `e_flags` bit to represent `FINX` ABI,
  toolchian could use that to identify how to disassemble the object
  file correctly.

- We add `ilp32x` and `lp64x` to used for RV32 and RV64 with `zfinx`,
  this design is distinguish from existing `ilp32` and `lp64` calling
  convention.

- In #166, we discuss the potential optimization of this ABI, like register
  argument alignment and changing size of `long double`, but I think
  it's scope of EABI, although it's still unclear.

- Interaction with EABI:
  - Add EABI variant of `ilp32x` and `lp64x`: `eilp32x` and `elp64x`.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -305,7 +305,7 @@ This specification defines the following named ABIs:
   EF_RISCV_RVE).
 
 * <a name=abi-ilp32x></a> **ILP32X**: ILP32 with FinX calling convention (i.e.
-  ELFCLASS32, EF_RISCV_FLOAT_ABI_SOFT and EF_RISCV_FINX).
+  ELFCLASS32 and EF_RISCV_FLOAT_ABI_FINX).
 
 * <a name=abi-lp64></a> **LP64**: Integer calling-convention only, hardware
   floating-point calling convention is not used (i.e. ELFCLASS64 and
@@ -321,7 +321,7 @@ This specification defines the following named ABIs:
   convention for FLEN=128 (i.e. ELFCLASS64 and EF_RISCV_FLOAT_ABI_QUAD).
 
 * <a name=abi-lp64x></a> **LP64X**: LP64 with FinX calling convention (i.e.
-  ELFCLASS64, EF_RISCV_FLOAT_ABI_SOFT and EF_RISCV_FINX).
+  ELFCLASS64 and EF_RISCV_FLOAT_ABI_FINX).
 
 The ILP32\* ABIs are only compatible with RV32\* ISAs, and the LP64\* ABIs are
 only compatible with RV64\* ISAs. A future version of this specification may
@@ -478,9 +478,9 @@ rules about 2&times;XLEN aligned arguments being passed in "aligned" register pa
 * e_flags: Describes the format of this ELF file.  These flags are used by the
   linker to disallow linking ELF files with incompatible ABIs together.
 
-   Bit 0 | Bit  1 - 2 | Bit 3 | Bit 4 | Bit 5 | Bit 6 - 31
-  -------|------------|-------|-------|-------|-----------
-   RVC   | Float ABI  |  RVE  |  TSO  | FINX  | *Reserved*
+   Bit 0 | Bit  1 - 2 | Bit 3 | Bit 4 | Bit 5      | Bit 6 - 31
+  -------|------------|-------|-------|------------|-----------
+   RVC   | Float ABI  |  RVE  |  TSO  | Float ABI  | *Reserved*
 
   * EF_RISCV_RVC (0x0001): This bit is set when the binary targets the C ABI,
     which allows instructions to be aligned to 16-bit boundaries (the base RV32
@@ -499,17 +499,16 @@ rules about 2&times;XLEN aligned arguments being passed in "aligned" register pa
     store "float" and "double" values in F registers, but would not store "long
     double" values in F registers.  If none of the float ABI flags are set, the
     object is taken to use the soft-float ABI.
-  * EF_RISCV_FLOAT_ABI (0x0006): This macro is used as a mask to test for one
+  * EF_RISCV_FLOAT_ABI_FINX (0x0020): This bit is set when the binary targets
+    the FinX ABI which will use general purpose register for floating-point
+    instructions.
+  * EF_RISCV_FLOAT_ABI (0x0026): This macro is used as a mask to test for one
     of the above floating-point ABIs, e.g.,
     `(e_flags & EF_RISCV_FLOAT_ABI) == EF_RISCV_FLOAT_ABI_DOUBLE`.
   * EF_RISCV_RVE (0x0008): This bit is set when the binary targets the E ABI.
   * EF_RISCV_TSO (0x0010): This bit is set when the binary requires the RVTSO
     memory consistency model.
-  * EF_RISCV_FINX (0x0020): This bit is set when the binary targets the
-    FinX ABI which will use general purpose register for floating-point
-    instructions.
-
-  Until such a time that the *Reserved* bits (0xffffffe0) are allocated by
+  Until such a time that the *Reserved* bits (0xffffff80) are allocated by
   future versions of this specification, they shall not be set by standard
   software.
 
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -278,6 +278,13 @@ The ILP32E calling convention is not compatible with ISAs that have registers
 that require load and store alignments of more than 32 bits. In particular, this
 calling convention must not be used with the D ISA extension.
 
+## <a name=ilp32e-calling-convention></a> ILP32X/LP64X Calling Convention
+
+The ILP32X/LP64 calling convention is designed to be usable with the zfinx
+extension. This calling convention is the same as the integer calling
+convention, but this calling convention must be used with Zfinx extension, that
+also implied this calling convention can't used with F extension.
+
 ## <a name=named-abis></a> Named ABIs
 
 This specification defines the following named ABIs:
@@ -297,6 +304,9 @@ This specification defines the following named ABIs:
   calling convention is not used (i.e. ELFCLASS32, EF_RISCV_FLOAT_ABI_SOFT, and
   EF_RISCV_RVE).
 
+* <a name=abi-ilp32x></a> **ILP32X**: ILP32 with FinX calling convention (i.e.
+  ELFCLASS32, EF_RISCV_FLOAT_ABI_SOFT and EF_RISCV_FINX).
+
 * <a name=abi-lp64></a> **LP64**: Integer calling-convention only, hardware
   floating-point calling convention is not used (i.e. ELFCLASS64 and
   EF_RISCV_FLOAT_ABI_SOFT).
@@ -309,6 +319,9 @@ This specification defines the following named ABIs:
 
 * <a name=abi-lp64q></a> **LP64Q**: LP64 with hardware floating-point calling
   convention for FLEN=128 (i.e. ELFCLASS64 and EF_RISCV_FLOAT_ABI_QUAD).
+
+* <a name=abi-lp64x></a> **LP64X**: LP64 with FinX calling convention (i.e.
+  ELFCLASS64, EF_RISCV_FLOAT_ABI_SOFT and EF_RISCV_FINX).
 
 The ILP32\* ABIs are only compatible with RV32\* ISAs, and the LP64\* ABIs are
 only compatible with RV64\* ISAs. A future version of this specification may
@@ -465,9 +478,9 @@ rules about 2&times;XLEN aligned arguments being passed in "aligned" register pa
 * e_flags: Describes the format of this ELF file.  These flags are used by the
   linker to disallow linking ELF files with incompatible ABIs together.
 
-   Bit 0 | Bit  1 - 2 | Bit 3 | Bit 4 | Bit 5 - 31
-  -------|------------|-------|-------|------------
-   RVC   | Float ABI  |  RVE  |  TSO  | *Reserved*
+   Bit 0 | Bit  1 - 2 | Bit 3 | Bit 4 | Bit 5 | Bit 6 - 31
+  -------|------------|-------|-------|-------|-----------
+   RVC   | Float ABI  |  RVE  |  TSO  | FINX  | *Reserved*
 
   * EF_RISCV_RVC (0x0001): This bit is set when the binary targets the C ABI,
     which allows instructions to be aligned to 16-bit boundaries (the base RV32
@@ -492,6 +505,9 @@ rules about 2&times;XLEN aligned arguments being passed in "aligned" register pa
   * EF_RISCV_RVE (0x0008): This bit is set when the binary targets the E ABI.
   * EF_RISCV_TSO (0x0010): This bit is set when the binary requires the RVTSO
     memory consistency model.
+  * EF_RISCV_FINX (0x0020): This bit is set when the binary targets the
+    FinX ABI which will use general purpose register for floating-point
+    instructions.
 
   Until such a time that the *Reserved* bits (0xffffffe0) are allocated by
   future versions of this specification, they shall not be set by standard
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -280,7 +280,7 @@ calling convention must not be used with the D ISA extension.
 
 ## <a name=ilp32e-calling-convention></a> ILP32X/LP64X Calling Convention
 
-The ILP32X/LP64 calling convention is designed to be usable with the zfinx
+The ILP32X/LP64X calling convention is designed to be usable with the zfinx
 extension. This calling convention is the same as the integer calling
 convention, but this calling convention must be used with Zfinx extension, that
 also implied this calling convention can't used with F extension.


kito-cheng:
@tariqkurd-repo @lazyparser

riscv-elf.md
```diff
@@ -278,6 +278,13 @@ The ILP32E calling convention is not compatible with ISAs that have registers
 that require load and store alignments of more than 32 bits. In particular, this
 calling convention must not be used with the D ISA extension.
 
+## <a name=ilp32e-calling-convention></a> ILP32X/LP64X Calling Convention
+
+The ILP32X/LP64 calling convention is designed to be usable with the zfinx
+extension. This calling convention is the same as the integer calling
+convention, but this calling convention must be used with Zfinx extension, that
```
palmer-dabbelt, line 285: IMO this should be something like "This calling convention is the same as the integer calling convention, but the rounding  mode and accrued exceptions (I forgot if we clobber those?) are passed in FCSR."

That allows us to eventually generate code that implements the Zfinx ABI on systems that implement F.  That might not be such a useful target, but it'll avoid painting ourselves into a corner WRT Zfinx binaries.

riscv-elf.md
```diff
@@ -465,9 +478,9 @@ rules about 2&times;XLEN aligned arguments being passed in "aligned" register pa
 * e_flags: Describes the format of this ELF file.  These flags are used by the
   linker to disallow linking ELF files with incompatible ABIs together.
 
-   Bit 0 | Bit  1 - 2 | Bit 3 | Bit 4 | Bit 5 - 31
-  -------|------------|-------|-------|------------
-   RVC   | Float ABI  |  RVE  |  TSO  | *Reserved*
+   Bit 0 | Bit  1 - 2 | Bit 3 | Bit 4 | Bit 5 | Bit 6 - 31
+  -------|------------|-------|-------|-------|-----------
+   RVC   | Float ABI  |  RVE  |  TSO  | FINX  | *Reserved*
```
palmer-dabbelt, line 483: Maybe we should just call the FINX bit the third bit of the Float ABI?  It really is just an entirely different float ABI, you can't independently control FINX in conjunction with F/D.  This will encode the valid ABIs the same way (ie, 00 for float ABI and 1 for FINX/ float ABI 3) but will save us some bit patterns in case we ever need more floating-point ABIs in the future.

riscv-elf.md
kito-cheng: FCSR Clobber by function call is the current implementation, but the rounding mode and accrued exceptions isn't depend on ABI (at least in newlib), it depend on __riscv_flen, but we don't have clear description on that before, so add more word to state that sounds good idea.

[1] https://github.com/riscv/riscv-newlib/blob/riscv-newlib-3.2.0/newlib/libm/machine/riscv/fesetenv.c#L57

riscv-elf.md
kito-cheng: I consider that before[1], but I am not sure it's good way to go, since finx is slight different than the floating-point ABI, rv32if, rv32id and rv32iq can use soft float abi like ilp32, because it's really just describe how we pass the argument, but `FINX` is describe ISA feature stuff rather than floating-point calling convention.

[1] https://github.com/riscv/riscv-elf-psabi-doc/issues/166#issuecomment-798936200

riscv-elf.md
```diff
@@ -278,6 +278,13 @@ The ILP32E calling convention is not compatible with ISAs that have registers
 that require load and store alignments of more than 32 bits. In particular, this
 calling convention must not be used with the D ISA extension.
 
+## <a name=ilp32e-calling-convention></a> ILP32X/LP64X Calling Convention
+
+The ILP32X/LP64 calling convention is designed to be usable with the zfinx
```
cmuellner, line 283: Typo: s/LP64/LP64X

riscv-elf.md
palmer-dabbelt: I don't think that's the right way to go with ABIs.  IMO the whole point of splitting the ABI away from the ISA is to make sure that ABIs can be implemented on any ISA that has the required state elements to implement that ABI, which allows for compatibility between systems.  The other ABI flags don't prevent running binaries on systems, just linking them.  We've always said we'd add some other mechanism to prevent running the wrong binaries, and IDK if we ever got around to the whole 

That's what we've done (or at least was my understanding of the ABIs) for `RVC` and `RVE`: binaries without the `RVC` bit set can still be run on systems with the C extension (and can even contain C instructions), they just can't take advantage of the 16-bit alignment that the C extensions allows for (the same for binaries with the `RVE` bit set, which just can't use the extra 16 registers the I base ISA provides).  I don't see any reason that `FINX` can't be described the same way: systems with FD can still implement the ABI, they just need to juggle any FP arguments they want to compute on between the X and F registers.  For `Zfinx` that might seem a bit silly because you can't have both flavors at the same time, but decoupling the ABI allows users to gradually move between these two extensions.

I don't really care that much what we call the bits, my real goal here is to avoid allocating ELF flag bit combinations that aren't necessary.  Specifically, the only sane ABIs I can think of on Zfinx systems are "full soft float" and "FP args in X registers, global rounding mode in FCSR".  With two combinations we just need one bit, so as long as we're explicit in the ABI document about the "Float ABI" field being reserved (with soft float being the only allowed value) I'm fine what whatever names get picked.  I just don't want to lose those other "Float ABI" bit combinations, as they may be useful in the future.

That said, if we did deeply screw up the FCSR part of the ABI and it's too late to fix that then we may not even need a bit here.  Let's sort that out first?

jrtc27: AAPCS64 makes FPSR call-clobbered and FPCR requires calls to library functions (and to either preserve other bits at function boundaries or to not touch them at all), so as long as we similarly make no guarantees about FCSR then everything should just be fine to treat soft-float and Zfinx "hard"-float identically?

riscv-elf.md
palmer-dabbelt: I alluded to this being an ABI bug, but I think it's actually correct: the choice to use FCSR as opposed to a memory location is entirely contained within the C libraries, so the correct implementation will be called for the system (FSCR for systems with F registers, which makes the FP instructions function correctly).  I think that means we don't actually have any additional ABI exposed, at least at the C function call level, for global rounding modes.  Thus the presence of FCSR isn't really relevant to the ABI, that's all up to the C libraries.  It's a bit early and I'm juggling two meetings though, so I might be confused here.

I guess my question now is: do we even need a ZFINX bit?

jrtc27: That's what I keep trying to argue too :) I'm of the view we need no new ABI

aswaterman: Is the implication that compilers aren’t allowed to inline routines like fetestexcept? If all accesses to fcsr truly require a call into the C library, I think I agree.

jrtc27: That's my understanding. Neither GCC nor Clang (both trunk, both -O3) inline it for x86_64 or AArch64. I am far from an expert on floating-point in C, though, so it would be worthwhile seeking the opinion of someone better versed in this area, although I think the worst that happens is we eventually realise we do need a new ABI to make Zfinx faster for code that does use such library functions regularly, which is a much better place to be than adding an ABI we turn out to not need?

jrtc27: In particular, I'm vaguely aware of `#pragma STDC FENV_ACCESS ON`, and don't know if it has any nasty implications for this kind of thing.

aswaterman: Yeah, I'm outside of my competence zone, but I'm still tentatively thinking this is OK.

palmer-dabbelt: I hadn't heard of `FENV_ACCESS` before.  It sounds like `off` is the direction we might care about there, according to http://www.cplusplus.com/reference/cfenv/FENV_ACCESS/

    If set to on, the program informs the compiler that it might access the floating-point environment to test its status flags (exceptions) or run under control modes other than the one by default.

    If set to off, the compiler may perform certain optimizations that can subvert these tests and mode changes, and thus accessing the floating-point environment in the cases described above, causes undefined behavior.

I can't get GCC to inline `fesetenv()` via either `FENV_ACCESS` (which I think isn't supported in GCC in general) or via `-frounding-math` (which appears to require some backend hooks we don't have in GCC).

Assuming that's the case for LLVM as well I think the right way to go here is to add some sort of "fenv can only be get/set via library routines" line in the ABI somewhere.  If we need to make it fast we can add an ABI flag for that, but IMO that's not really a `Zfinx` ABI but instead a "fcsr can be accessed inline" ABI.  It would be possible to implement that ABI on either `Zfinx` or `F`.

aswaterman: @palmer-dabbelt Just to be clear, the hard-float calling conventions should continue to permit direct access to `fcsr`.  It's merely the soft-float conventions that should impose this constraint.

palmer-dabbelt: Ya, I don't see any reason to ban hard-float from direct access to `fcsr` -- it has to be there for the ABI to work otherwise, so that's fine.  Sorry that wasn't clear, I guess I was just down this rabbit hole.

jrtc27: It depends what you're doing; the AArch64 PCS says the following, which is quite a bit limited than what you say, and most of it would apply to a soft-float F or Zfinx compilation too:

> The FPSR is a status register that holds the cumulative exception bits of the floating-point unit. It contains the fields IDC, IXC, UFC, OFC, DZC, IOC and QC. These fields are not preserved across a public interface and may have any value on entry to a subroutine.
>
> The FPCR is used to control the behavior of the floating-point unit. It is a global register with the following properties.
>
> * The exception-control bits (8-12), rounding mode bits (22-23), flush-to-zero bits (24), and the AH and FIZ bits (0-1) may be modified by calls to specific support functions that affect the global state of the application.
> * The NEP bit (bit 2) must be zero on entry to and return from a public interface.
> * All other bits are reserved and must not be modified. It is not defined whether the bits read as zero or one, or whether they are preserved across a public interface.

palmer-dabbelt: OK, so just thinking about this I do think we have a bug.  The intent would be to allow something like

    double func(double a, double b) {
        fesetenv(...);
        return a * b;
    }

to function correctly in four combinations (all soft-float ABI):

* F code calling F libc: this works, they both use FCSR.
* F code calling non-F libc: this is broken.  The caller will use hard-float instructions, while the callee will set the soft-float rounding mode.
* non-F code calling F libc: IIUC this works, we've plumbed the soft-float fenv into fcsr (at least in glibc, though maybe someone should test that)
* non-F code calling non-F libc: this works, they both use soft float

IMO we can fix this by dynamically checking for fcsr in the soft-float libraries, which shouldn't be all that hard, won't require re-compiling user code, and we can just call a bug in glibc.  We should write this down somewhere in the ABI, though, and at this point I think it's probably best to put together another thread as I'm not sure what the behavior should actually be.

aswaterman: @palmer-dabbelt yes, I was assuming that the F-calling-soft-float case would require the C library to understand and actually use the fcsr internally (or combine fcsr with an in-memory variable or whatever). Not saying we’re getting this right today, but it seems doable.

palmer-dabbelt: OK.  I'm slammed with work stuff right now, but if nobody has time to write something up before I get back to this in my inbox (it'l like a thousand deep right now) then I'll put something together and go fix the glibc bug.

riscv-elf.md
kito-cheng: @palmer-dabbelt So your are preferring keep the other 3 possible ABI combination in e_flags, I strongly prefer record `zfinx` info in `e_flags`, I think let me update for that.

aswaterman (changes requested):
As seen in the comments on this PR, some of us do not believe an additional set of ABIs is necessary.  We should resolve that question before proceeding.

riscv-elf.md
aswaterman: Thanks, @palmer-dabbelt; if you can't get to it soon, then you should probably give a brain dump of your current thinking to @kito-cheng.  He can either flesh out the idea or tell us that we're wrong.

riscv-elf.md
palmer-dabbelt: I guess I just don't see what the actual difference is in the ABI that is getting indicated by this flag.  If it's just the "fcsr can be used directly" thing from above then that's fine, we should just be explicit about what the ABI means and about having the other flags reserved.

jim-wilson:
Most of the discussion has been about how the zfinx abi calling conventions aren't really any different from the ilp32/lp64 ABIs.  But the problem is not with the ABI but how the ABI and arch interact.

Consider this testcase
float
sub (int i, int j, int k, int l, float a, float b, float c)
{
  return a + b + c;
}
compiled with riscv32-unknown-elf-gcc -march=rv32gc -mabi=ilp32 -O -S I get
sub:
	fmv.w.x	fa5,a4
	fmv.w.x	fa4,a5
	fadd.s	fa5,fa5,fa4
	fmv.w.x	fa4,a6
	fadd.s	fa5,fa5,fa4
	fmv.x.w	a0,fa5
	ret
and note that the fmv.w.x fa5,a4 will clobber a5 on a zfinx machine, causing the code to fail.  I don't have a zfinx compiler handy, but if we compile this for zfinx the compiler can emit fadd.s a5,a5,a4 which will fail on a rv32if machine, as the instruction will operate on FP regs but the arguments will be in int regs.

So f/ilp32 and zfinx/ilp32 code are not compatible in either direction.  We need to give an error if someone tries to link the two different kinds of code together.  One way to do that is to make a new ABI ilp32x and then refuse to link that with other ABIs.  Another possibly better way to do this is to check the architecture.  Except that the architecture is only recorded in the ELF attributes, and we only emit the attributes for embedded targets.  We don't emit them for linux.  So there is a backward compatible problem.  Though maybe we can solve that by saying if that if you use zfinx then you must have ELF attributes in the object file, and if you link with a file that doesn't have ELF attributes that is an error.  Another possible solution is to put a zfinx arch bit in e_flags, which serves the same purpose as Kito's proposal, we are just calling it an arch bit now instead of an abi bit.

jrtc27:
We emit the attributes unconditionally in LLVM. GCC and binutils need to do the same if they're not already, then, because this is precisely the one of the problems it's there to address.

jim-wilson:
The ELF attributes weren't part of the psABI, and still aren't.  Maybe we need to fix that before we can require them?  Plus there is the concern that there are a lot of different random linux tools from random sources that might be confused about new sections in object files.  We didn't need them for linux as badly as we needed them for embedded targets, so the safe solution was to only emit them for embedded elf at the time.

It would be reasonable to enable them for linux in gcc now, but this will need to be tested to make sure we don't break distro builds, or linux kernel builds, or something else important.

jrtc27:
Please start testing then, because if there are issues with it we need to know ASAP otherwise this will block all the upcoming extensions, though I highly doubt there will be outside of maybe the Linux kernel itself; FreeBSD uses LLVM, which always emits the attributes section, and I have not heard of there being any issues as a result of that when building thousands of packages from its ports collection for RISC-V (nor do I recall it posing issues for the kernel, either, so Linux might still Just Work).

kito-cheng:
@jrtc27 and me has a discussion on ELF attribute today in private, the issue on ELF attribute didn't enabled on Linux is because ifunc scenario are not work well for current ELF attribute scheme/implementation yet.

Consider following case [Note: *1] compile with `-march=rv64i -mabi=lp64`:
```c
int
__attribute__((target("m")))
mul_m(int x, int y) {
    return x * y;
}

int
mul_i(int x, int y) {
    return x * y;
}
```
We'll emit something like following assembly code, set the arch attribute to rv64i, and then...use M-extension instruction in `mul_m`, which can't work with gnu assembler.
```
    .attribute      5, "rv64i2p0"
...
mul_m:
...
        mulw    a0, a1, a0
        ret
mul_i:
...
        call    __muldi3@plt
...
        ret

```

I'll create another another issue to discuss this, let focus on the zfinx ABI here :)

Note:
*1: This code is compile-able with clang/LLVM, but can't compile with GCC due to target attribute unimplemented.

jim-wilson:
I tested and committed a patch to enable riscv attributes by default for all riscv targets in the FSF GCC tree.

jim-wilson:
I don't think that Kito's point about attribute target is a problem.  The architecture specified in the attributes is the minimum that the code was compiled for.  It doesn't need to handle ifunc functions compiled for an architecture higher than the minimum.

jrtc27:
> I don't think that Kito's point about attribute target is a problem. The architecture specified in the attributes is the minimum that the code was compiled for. It doesn't need to handle ifunc functions compiled for an architecture higher than the minimum.

The issue is just that `.attribute arch` sets what the assembler will allow, so if you round-trip via assembly it no longer assembles without giving errors about needing to enable extensions.

jrtc27:
(or, even if you didn't make `.attribute arch` do that and required -march= be the maximal set of extensions, that's not always possible, e.g. if you want to ifunc between F and Zfinx code)

kito-cheng:
Hi all:

Here is an alternative proposal here:
- Keep using -mabi=ilp32/-mabi=lp64
- NO new bit for e_flags.
- Disassemble and linker use ELF attribute to check the compatibility.
- Advantage:
  - Prevent too much potential ZFINX variant, all calling convention are same but ZFINX 

For this proposal, we don't need to change the document.

kito-cheng:
We use the alternative proposal:
https://github.com/riscv/riscv-elf-psabi-doc/pull/198

Issue #66, opened 2017-12-20
On RISC-V, lazily bound functions must follow the ABI


palmer-dabbelt:
See this x86 ABI bug for more details: https://sourceware.org/bugzilla/show_bug.cgi?id=21265

I vote we solve this on RISC-V by just requiring that lazily bound functions follow the standard ABI.  This lets us avoid saving the rest of the X registers, and assuming we can do the fixup without any extensions (which seems reasonable, and is what we currently do despite it not actually being enforced) we won't need to save F or V state here.

I don't know where this should go in the manual, @asb?

aswaterman:
I'd phrase it as "Functions exported from shared libraries must follow the standard calling convention.  Alternate calling conventions are permitted only for functions with internal linkage."

palmer-dabbelt:
The glibc people were strict about restricting this to lazily bound functions, as those aren't the default for shared libraries on some systems.  IDK if those actually matter, though, as they're probably pretty esoteric.

aswaterman:
Fair enough

jim-wilson:
On Tue, Dec 19, 2017 at 6:02 PM, Palmer Dabbelt <notifications@github.com>
wrote:

> The glibc people were strict about restricting this to lazily bound
> functions, as those aren't the default for shared libraries on some
> systems. IDK if those actually matter, though, as they're probably pretty
> esoteric.
>

There is a linker option -z now that turns off lazy binding at compile
time.  There is also a glibc environment variable LD_BIND_NOW=1 that turns
off lazy binding at run time.  Probably not many people other than
toolchain developers know about this stuff, though I do see some references
to LD_BIND_NOW in some FAQs found via a web search, so some people do find
it occasionally useful.

Jim

palmer-dabbelt:
`-z now` is good enough of a reason to make this distinction in the ABI document.

asb:
It had never occurred to me that people might do something _other_ than use the standard calling convention for lazily bound functions.

I'm not entirely sure where it should go. A note under "procedure calling convention" perhaps? Phrasing it as "must follow the standard calling convention" might be confusing - with RISC-V we of course have a range of standard calling conventions, and the requirement is really to conform to the specific standard calling convention used on that system.

jrtc27:
[...]

PR #190, opened 2021-05-12
Calling conventions for the lazily bound functions.

Hsiangkai:
Currently, we save/restore a0-a7 and fa0-fa7 in glibc implementation to
avoid ruining the arguments of the resolving functions. In order to
define the vector calling convention, we need to specify the behavior of
the resolver in the dynamic linker. Vector registers may have a large
size and save/restore vector argument registers may occupy a large
portion of stack space during runtime resolving. It is unreasonable
large to save/restore all these vector argument registers in the
resolver.

In this patch, we define a special symbol attribute to avoid going
through the resolver for the special purpose, i.e., non-standard calling
convention or vector calling convention. We also define an additional
dynamic tag to indicate there are symbols with the special attribute in
the dynamic symbol table of the object.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -44,6 +44,20 @@ NOTE: This is not yet a requirement as existing toolchains predating this part
 of the specification do not adhere to this, and without improvements to linker
 relaxation support doing so would regress performance and code size.
 
+== Dynamic Linking
+
+Run-time linkers that use lazy binding must preserve all argument registers
+used in the standard calling convention for the ABI in use. Any functions that
+use additional argument registers must be annotated with
+`STO_RISCV_VARIANT_CC`, as defined in <<Symbol Table>>.
+
+NOTE: Vector registers have a variable size depending on the hardware
+implementation and can be quite large. Saving/restoring all these vector
+arguments in a run-time linker's lazy resolver would use a large amount of
+stack space and hurt performance. This attribute allows vector registers to
+not be part of the standard calling convention so run-time linkers are not
+required to save/restore them and can instead eagerly bind such functions.
+
 == C++ Name Mangling
 
 {Cpp} name mangling for RISC-V follows the
@@ -134,7 +148,22 @@ There are no RISC-V specific definitions relating to ELF string tables.
 
 === Symbol Table
 
-There are no RISC-V specific definitions relating to ELF symbol tables.
+st_other:: The lower 2 bits are used to specify a symbol's visibility. The
+remaining 6 bits have no defined meaning in the ELF gABI. We use the highest
+bit to mark functions that do not follow the standard calling convention for
+the ABI in use.
++
+The defined processor-specific `st_other` flags are listed in the following
+table.
++
+[%autowidth]
+|===
+| Name                 | Mask
+
+| STO_RISCV_VARIANT_CC | 0x80
+|===
++
+See <<Dynamic Linking>> for the meaning of `STO_RISCV_VARIANT_CC`.
 
 === Relocations
 
@@ -682,9 +711,21 @@ table.
 
 There are no RISC-V specific definitions relating to ELF note sections.
 
-=== Dynamic Table
+=== Dynamic Section
 
-There are no RISC-V specific definitions relating to dynamic tables.
+The defined processor-specific dynamic array tags are listed in the following
+table.
+
+[%autowidth]
+|===
+| Name                | Value      | d_un  | Executable        | Shared Object
+
+| DT_RISCV_VARIANT_CC | 0x70000001 | d_val | Platform specific | Platform specific
+|===
+
+An object must have the dynamic tag `DT_RISCV_VARIANT_CC` if it has one or more
+`R_RISCV_JUMP_SLOT` relocations against symbols with the `STO_RISCV_VARIANT_CC`
+attribute.
 
 === Hash Table
 


kito-cheng:
Thanks for come up this , it's basis of vector calling conversion and also address #66.

- I would suggest using `lazy binding` rather than `lazily bound`, since `lazy binding` is more widely used term.
- We need a new asm directive to specify symbol is specify is `STO_RISCV_BIND_NOW`, could you send a PR to [riscv-asm-manual](https://github.com/riscv/riscv-asm-manual) to add one?

riscv-elf.md
```diff
@@ -507,7 +520,30 @@ There are no RISC-V specific definitions relating to ELF string tables.
...
+  RISC-V specific `st_other` flags
+  --------------------------
+  Name               | Mask
+  -------------------|------
+  STO_RISCV_BIND_NOW | 0x80
+
+  If `STO_RISCV_BIND_NOW` is set, the dynamic linker will resolve the symbol
+  during program loading. The resolved symbol address will be filled into GOT
+  entry regardless `LD_BIND_NOW` is set or not under dynamic linking for PIC
+  code.
```
ebahapo, line 542: `s/for PIC code//`

ebahapo:
Somewhere, not necessarily in this document, it should be stated that dynamic functions using the V instructions should get this flag.

riscv-elf.md
```diff
@@ -371,6 +372,18 @@ offset, relative to the value of the `gp` register, can be produced, referring
...
 Does not apply to the ILP32 ABIs.
 
+## <a name=dynamic-linking /> Dynamic linking
+
+Lazily bound functions must follow the standard calling conventions. The
+resolver in the dynamic linker will save/restore `a0-a7` for integer calling
+convention and save/restore `fa0-fa7` for hardware floating-point calling
+convention to avoid ruining the arguments before jumping to the resolved
+function. If there is any need to support non-standard calling convention for
+lazily bound functions, the symbol needs to be decorated using
```
ebahapo: `s/decorated/annotated/`

riscv-elf.md
```diff
@@ -371,6 +372,16 @@ offset, relative to the value of the `gp` register, can be produced, referring
...
 Does not apply to the ILP32 ABIs.
 
+## <a name=dynamic-linking /> Dynamic linking
+
+Lazy binding must follow the standard calling conventions. The resolver in the
+dynamic linker will save/restore `a0-a7` for integer calling convention and
+save/restore `fa0-fa7` for hardware floating-point calling convention to avoid
+ruining the arguments before jumping to the resolved function. If there is any
+need to support non-standard calling convention for lazy binding, the symbol
+needs to be decorated using `STO_RISCV_BIND_NOW`. With the attribute in the
```
ebahapo, line 382: `s/with the/with this/`

Hsiangkai:
> Thanks for come up this , it's basis of vector calling conversion and also address #66.
> 
> * I would suggest using `lazy binding` rather than `lazily bound`, since `lazy binding` is more widely used term.
> * We need a new asm directive to specify symbol is specify is `STO_RISCV_BIND_NOW`, could you send a PR to [riscv-asm-manual](https://github.com/riscv/riscv-asm-manual) to add one?

* I replace "lazily bound functions" with "lazy binding".
* I add a new asm directive here. https://github.com/riscv/riscv-asm-manual/pull/64

Hsiangkai:
> Somewhere, not necessarily in this document, it should be stated that dynamic functions using the V instructions should get this flag.

I think I should update #171. Should functions using the V instructions get this flag or using the V arguments get this flag? All the vector registers are caller-saved.

jrtc27:
To me this conflates implementation ("don't lazy bind") with higher-level justification ("this follows a non-standard calling convention"). We should be labelling symbols as using a non-standard calling convention and letting implementations make that imply a lack of lazy binding _if needed_, *not* being so fixed in the view of the world and calling it "bind now". Incidentally, AArch64 takes a similar approach and labels them as STO_AARCH64_VARIANT_PCS, *not* STO_AARCH64_BIND_NOW.

ebahapo:
> > Somewhere, not necessarily in this document, it should be stated that dynamic functions using the V instructions should get this flag.
> 
> I think I should update #171. 

It's probably a good place to start.

> Should functions using the V instructions get this flag or using the V arguments get this flag? All the vector registers are caller-saved.

Just those using V arguments, yes?

jrtc27:
I do also feel we'll want to separate out a vector calling convention so that those can still be lazily bound with a suitable resolver, though that can always come later (and it's still be marked STO_RISCV_VARIANT_CC or whatever, just _also_ STO_RISCV_VECTOR_CALL for implementations that want to handle that case). Otherwise you're boxed into a corner if you call it BIND_NOW and can never extend it without violating its original intent.

Hsiangkai:
> I do also feel we'll want to separate out a vector calling convention so that those can still be lazily bound with a suitable resolver, though that can always come later (and it's still be marked STO_RISCV_VARIANT_CC or whatever, just _also_ STO_RISCV_VECTOR_CALL for implementations that want to handle that case). Otherwise you're boxed into a corner if you call it BIND_NOW and can never extend it without violating its original intent.

It makes sense. STO_RISCV_VARIANT_CC is better. I will rename it. Thanks.

kito-cheng:
>> Should functions using the V instructions get this flag or using the V arguments get this flag? All the vector registers are caller-saved.
>
> Just those using V arguments, yes?

It only for those function who have V arguments.

kito-cheng:
Just note, for this changes we need to implement following component:
- assembler: New syntax for setting a symbol is STO_RISCV_VARIANT_CC.
- compiler: Must mark a symbol is STO_RISCV_VARIANT_CC if not using standard calling conversion or using parameter not passed in GPR/FPR. 
- linker: generate DT_RISCV_VARIANT_CC for dynamic section if any symbol with STO_RISCV_VARIANT_CC.
- dynamic linker: glibc and all other dynamic linker must recognize this new flag and handle this.

jim-wilson (approved):

kito-cheng (approved):
LGTM, but I would suggest we should have assembler, linker and dynamic linker PoC before merge.

riscv-elf.md
```diff
@@ -371,6 +372,16 @@ offset, relative to the value of the `gp` register, can be produced, referring
 to address literals in the GOT.  This code model is position independent.
 Does not apply to the ILP32 ABIs.
 
+## <a name=dynamic-linking /> Dynamic linking
+
+Lazy binding must follow the standard calling conventions. The resolver in the
+dynamic linker will save/restore `a0-a7` for integer calling convention and
```
palmer-dabbelt, line 378: I'd cut out the bit about save/restore, and just say that binding must follow the standard calling conventions.  IMO it's really up to the C library to determine what it needs to save/restore, and while we do currently just save/restore both X and F args in glibc that's just because we haven't figured out a reliable way to exclude the F extensions from the dynamic loader.  These are all build system concerns in glibc and other C libraries may not have those concerns so I don't want to force them to spin up the FPU if they don't need to.

riscv-elf.md
```diff
@@ -507,7 +518,30 @@ There are no RISC-V specific definitions relating to ELF string tables.
 
 ## <a name=symbol-table></a>Symbol Table
 
-There are no RISC-V specific definitions relating to ELF symbol tables.
+* st_other: The lower 2 bits are used to specify a symbol's visibility. The
+  remaining 6 bits have no defined meaning in gABI. We use the highest bit for
+  the eager binding semantic of the function call with the non-standard calling
+  convention or any other special purpose to avoid going through the resolver.
+  For example, vector registers have variant size. It may be from 128 bits to
+  4096 bits or larger. It depends on the hardware implementation. To
```
palmer-dabbelt, line 526: Much as above, I'd prefer to keep some of this extra wording out of the ABI.  We don't really have a way to split out the normative text from comments, and while having these examples is nice to explain why we're making certain decisions it could be construed as an actual ABI requirement.  From "128 bits to 4096 bits or larger" is pretty inclusive, but it's just unnecessarry.

Maybe the right answer here is to add some specific commentary decorations?  We've got a handful of these in the ABI doc right now so it'd be a nice cleanup.

palmer-dabbelt (commented):
Generally LGTM, just some wording stuff.

riscv-elf.md
jrtc27: My thinking for commentary while we aren't using asciidoc is something like:

> NOTE: Bla bla bla

jrtc27:
I don't understand why the dynamic tag needs to exist. AArch64 has DT_AARCH64_VARIANT_PCS, but as far as I can tell nothing actually consults it other than readelf-like things, according to code search.debian.net at least. Does anyone have any insight?

jrtc27:
Ah, never mind, it is checked in glibc, just obfuscated by being `DT_AARCH64 (VARIANT_PCS)` (grr why...), as a quick way to check whether it needs to scan the symbol table for variant PCS functions.

riscv-elf.md
```diff
@@ -507,7 +515,31 @@ There are no RISC-V specific definitions relating to ELF string tables.
...
+  STO_RISCV_VARIANT_CC | 0x80
+
+  If `STO_RISCV_VARIANT_CC` is set, the dynamic linker will always resolve the
+  symbol during program loading. The resolved symbol address will be filled
+  into GOT entry regardless if `LD_BIND_NOW` is set or not under dynamic
+  linking.
+
+  Static linkers must set the flag for the symbol following the eager binding
+  semantic in the dynamic symbol table and add a `DT_RISCV_VARIANT_CC` dynamic
+  tag in the Dynamic Section of the object.
```
jrtc27, line 536: This is overly prescriptive. If a dynamic linker wants to save and restore every single register, it can do so and then continue to lazily bind everything. It's likely a bad idea, but it's a correct implementation. This should only state what the requirements are; commentary about this likely implying eager binding belongs in a `> NOTE:`.

riscv-elf.md
```diff
@@ -507,7 +515,31 @@ There are no RISC-V specific definitions relating to ELF string tables.
 
 ## <a name=symbol-table></a>Symbol Table
 
-There are no RISC-V specific definitions relating to ELF symbol tables.
+* st_other: The lower 2 bits are used to specify a symbol's visibility. The
+  remaining 6 bits have no defined meaning in gABI. We use the highest bit for
+  the eager binding semantic of the function call with the non-standard calling
+  convention or any other special purpose to avoid going through the resolver.
```
jrtc27, line 521: See below

riscv-elf.md
```diff
@@ -988,9 +1020,16 @@ There are no RISC-V specific definitions relating to program header tables.
...
 There are no RISC-V specific definitions relating to ELF note sections.
 
-## <a name=dynamic-table></a>Dynamic Table
+## <a name=dynamic-section></a>Dynamic Section
+
+RISC-V specific dynamic array tags
+---------------------------------------------------------------------------------
+Name                | Value      | d_un  | Executable        | Shared Object
+--------------------|------------|-------|-------------------|-------------------
+DT_RISCV_VARIANT_CC | 0x70000001 | d_val | Platform specific | Platform specific
```
kito-cheng, line 1029: Curious why skip `0x70000000 (DT_LOPROC + 0)`? I saw some targets like Alpha, PPC, PPC64, IA64 using that value.

kito-cheng:
@jrtc27 
> I don't understand why the dynamic tag needs to exist. AArch64 has DT_AARCH64_VARIANT_PCS, but as far as I can tell nothing actually consults it other than readelf-like things, according to code search.debian.net at least. Does anyone have any insight?

Seems like it's a optimization to skip the checking st_other if DT_AARCH64_VARIANT_PCS is not set [1-2].

[1] https://github.com/riscv/riscv-glibc/commit/82bc69c012838a381c4167c156a06f4598f34227#diff-5c636a3d53a784449673f0122b405bd56513de968c2991e6ba2fd894fb34e930R109
[2] https://github.com/riscv/riscv-glibc/commit/82bc69c012838a381c4167c156a06f4598f34227#diff-5c636a3d53a784449673f0122b405bd56513de968c2991e6ba2fd894fb34e930R405

riscv-elf.md
kito-cheng: Oh, after checking with AArch64's glibc implementation[1], I realize that's an implementation optimization, we can use `sym->st_other & DT_RISCV_VARIANT_CC` to check that.

[1] https://github.com/riscv/riscv-glibc/commit/82bc69c012838a381c4167c156a06f4598f34227#diff-5c636a3d53a784449673f0122b405bd56513de968c2991e6ba2fd894fb34e930R412

Hsiangkai: No, DT_RISCV_VARIANT_CC is not used to check sym->st_other. AFAIK, there is an encoding rule for dynamic tags in gABI as follow.

"A tag whose value is an even number indicates a dynamic section entry that uses d_ptr. A tag whose value is an odd number indicates a dynamic section entry that uses d_val or that uses neither d_ptr nor d_val."

That's why it is encoded as 0x70000001.

riscv-elf.md
```diff
@@ -507,7 +515,31 @@ There are no RISC-V specific definitions relating to ELF string tables.
...
+  symbol during program loading. The resolved symbol address will be filled
+  into GOT entry regardless if `LD_BIND_NOW` is set or not under dynamic
+  linking.
+
+  Static linkers must set the flag for the symbol following the eager binding
+  semantic in the dynamic symbol table and add a `DT_RISCV_VARIANT_CC` dynamic
+  tag in the Dynamic Section of the object.
+
+> NOTE:
+  Vector registers have variant size. It depends on the hardware implementation.
```
frasercrmck, line 539: "variant size" doesn't read well to me. If it's a standard term then ignore this suggestion, but something like:

```suggestion
  Vector registers have a variable size depending on the hardware implementation.
```

riscv-elf.md
```diff
@@ -507,7 +515,31 @@ There are no RISC-V specific definitions relating to ELF string tables.
...
+  To save/restore all these vector arguments in the resolver may occupy a large
```
frasercrmck, line 540: This sentence is missing something. Either of these read better to me:
```suggestion
  To save/restore all these vector arguments, the resolver may occupy a large
```
or
```suggestion
  Saving/restoring all these vector arguments in the resolver may occupy a large
```

jrtc27:
This still talks about eager/lazy binding far too much in the normative text. Normative text should _only_ specify that the function uses a calling convention that differs from the standard calling convention specified in this document. _Only_ NOTE: text should talk about eager/lazy binding as a possible relevant thing. I'm happy to take over this PR if that would be easier to get this addressed sooner rather than later.

Hsiangkai:
> This still talks about eager/lazy binding far too much in the normative text. Normative text should _only_ specify that the function uses a calling convention that differs from the standard calling convention specified in this document. _Only_ NOTE: text should talk about eager/lazy binding as a possible relevant thing. I'm happy to take over this PR if that would be easier to get this addressed sooner rather than later.

Sure, please take over this PR to move it forward. Thanks.

riscv-elf.md
```diff
@@ -375,6 +376,21 @@ offset, relative to the value of the `gp` register, can be produced, referring
 to address literals in the GOT.  This code model is position independent.
 Does not apply to the ILP32 ABIs.
 
+## <a name=dynamic-linking></a>Dynamic Linking
+
+Run-time linkers that use lazy binding must preserve all argument registers
+used in the standard calling convention for the ABI in use. Any functions that
+use additional argument registers must be annoted with `STO_RISCV_VARIANT_CC`,
```
jim-wilson, line 383:
```suggestion
use additional argument registers must be annotated with `STO_RISCV_VARIANT_CC`,
```

jim-wilson (approved):

riscv-elf.md
jrtc27: Thanks, late night typos strike again...

jrtc27:
Rebased and fixed the typo. I believe this is just waiting on an LLVM ACK; @MaskRay?

riscv-elf.md
```diff
@@ -505,7 +521,20 @@ There are no RISC-V specific definitions relating to ELF string tables.
 
 ## <a name=symbol-table></a>Symbol Table
 
-There are no RISC-V specific definitions relating to ELF symbol tables.
+* st_other: The lower 2 bits are used to specify a symbol's visibility. The
+  remaining 6 bits have no defined meaning in the ELF gABI. We use the highest
```
MaskRay, line 525: The wording needs to be careful but the current one is fine.

https://groups.google.com/g/generic-abi/c/ytficX3WEWg/m/TfPyG4MFCQAJ

MaskRay (approved):
LGTM from my portion of LLVM viewpoint :) 

The subject needs an update. STO_AARCH64_VARIANT_PCS should be in the subject.

jrtc27:
(Rebased to resolve conflicts)

jrtc27:
> LGTM from my portion of LLVM viewpoint :)
> 
> The subject needs an update. STO_AARCH64_VARIANT_PCS should be in the subject.

Added both constant names to the commit message body. There isn't enough space in the subject to add either of them without having to make it too concise, in my opinion, unfortunately.

jrtc27:
> Just note, for this changes we need to implement following component:
> 
> * assembler: New syntax for setting a symbol is STO_RISCV_VARIANT_CC.
> * compiler: Must mark a symbol is STO_RISCV_VARIANT_CC if not using standard calling conversion or using parameter not passed in GPR/FPR.
> * linker: generate DT_RISCV_VARIANT_CC for dynamic section if any symbol with STO_RISCV_VARIANT_CC.
> * dynamic linker: glibc and all other dynamic linker must recognize this new flag and handle this.

I know the first two have patches for LLVM. Is there a patch for LLD or GNU ld yet for DT_RISCV_VARIANT_CC, and is there a patch for glibc yet (I know there isn't one for FreeBSD's rtld and that it's not something anyone at SiFive cares about)? If so we should merge this, but otherwise we're waiting on those to exist.

kito-cheng:
> I know the first two have patches for LLVM. Is there a patch for LLD or GNU ld yet for DT_RISCV_VARIANT_CC, and is there a patch for glibc yet (I know there isn't one for FreeBSD's rtld and that it's not something anyone at SiFive cares about)? If so we should merge this, but otherwise we're waiting on those to exist.

We(SiFive) have GNU ld and glibc patch, let me ask @Hsiangkai and @Nelson1225 for send that out.

jim-wilson:
The glibc patch is here
    https://sourceware.org/pipermail/libc-alpha/2021-August/129931.html
I don't see the binutils patch in the obvious places I looked.

jim-wilson:
The binutils patch is now
https://sourceware.org/pipermail/binutils/2021-August/117643.html

riscv-elf.md
```diff
@@ -1023,9 +1052,18 @@ PT_RISCV_ATTRIBUTES  | 0x70000003  | RISC-V ELF attribute section.
...
+
+The defined processor-specific dynamic array tags are listed in the following
+table.
+
+Name                | Value      | d_un  | Executable        | Shared Object
+:------------------ | :--------- | :---- | :---------------- | :-----------------
+DT_RISCV_VARIANT_CC | 0x70000001 | d_val | Platform specific | Platform specific
 
-There are no RISC-V specific definitions relating to dynamic tables.
+An object with the dynamic tag `DT_RISCV_VARIANT_CC` has one or more
```
MaskRay, line 1064: This sentence does not imply that `DT_RISCV_VARIANT_CC` must be present.

MaskRay (approved):
https://reviews.llvm.org/D107949 will add the LLVM binary utility support.

riscv-elf.md
kito-cheng: How about this?
```
An object must having  dynamic tag `DT_RISCV_VARIANT_CC` if having any `R_RISCV_JUMP_SLOT` relocations against symbols with the `STO_RISCV_VARIANT_CC` attribute.
```

jrtc27:
Rebased, converted to asciidoc and fixed dynamic tag wording to specify that it must be present if any such symbols are

jrtc27:
I think this has been open for feedback long enough at this point; if anyone has further feedback on the exact phrasing please file a new issue or PR, but let's finally move forward with this

PR #196, opened 2021-07-08
Proposal: Add mapping symbol

kito-cheng:
Revise history:
 Date | Comment
:---------------- | :--------
2021/07/22   | Remove `$d<N>`, add rule for `$x<ISA>`
2021/07/09   | Add implementation section to mention implantation related stuffs.

# What is Mapping Symbol

This proposal add a new special symbol class, called mapping symbol,
those symbol are used for assist disassembler to having better knowledge
of binary, it could be used for distinguish the code and data region.

Here is two kind of mapping symbol: data and instruction, and both
having an optional extra part in the symbol name for carry extra information.

Symbol Name | Meaning
:---------- | :-----------------------------------------------------------
$d          | Start of a sequence of data.
$x          | Start of a sequence of instructions.
$x\<ISA\>     | Start of a sequence of instructions with \<ISA\> extension.

Mapping symbol are also used by other ISA for same purpose, like ARM, AArch64,
C-sky and nds32.

# Data Mapping Symbol

Data mapping symbol could having an extra length information to present
the orignal data layout.

e.g.
```
.foo:
	.word 10
	.byte 1
	.word 20
	.byte 2
```

Without mapping symbol:
```
$ riscv64-unknown-elf-gcc foo.s -c
$ riscv64-unknown-elf-gcc foo.o -c

Disassembly of section .text:

0000000000000000 <.foo>:
   0:   000a                    c.slli  zero,0x2
   2:   0000                    unimp
   4:   1401                    addi    s0,s0,-32
   6:   0000                    unimp
   8:   0200                    addi    s0,sp,256
```

With mapping symbol:
```
Disassembly of section .text:

00000000 <.foo>:                                      # $d insert here.
   0:   0000000a        .word   0x0000000a
   4:   00001401        .word   0x00001401
   8:   Address 0x0000000000000008 is out of bounds.
```

# Instruction Mapping Symbol

Instruction mapping symbol with extra ISA info could also used for
ifunc, e.g. library are built with `rv64gc`, but few functions
like memcpy provide two version, one built with `rv64gc`, and one built with
`rv64gcv`, and select by ifunc mechanism at run-time; however the arch
attribute is recording for minimal execution environment requirement, so the ISA
information from arch attribute isn't enough for disassembler to disassemble the
`rv64gcv` version correctly.

# Implementation

Here is implementation note for each component of toolchain:
- Mapping symbol are automatically emit by assembler.
- Disassembler take mapping symbol as hint to having better better knowledge about original layout.
- Compiler don't need to aware the existence of mapping symbol.
- Linker don't need extra mechanism for handing mapping symbol, but linker are permit to merge the adjacency mapping symbol if they are exactly same type.
- Strip tool are permit to strip mapping symbol.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1164,6 +1164,44 @@ Merge policy:::
 The linker should report errors if object files of different privileged
 specification versions are merged.
 
+
+=== Mapping Symbol
+
+The section can have a mixture of code and data or code with different ISAs.
+A number of symbols, named mapping symbols, describe the boundaries.
+
+[%autowidth]
+|===
+| Symbol Name | Meaning
+| $d          | Start of a sequence of data.
+| $x          | Start of a sequence of instructions.
+| $x<ISA>     | Start of a sequence of instructions with <ISA> extension.
+|===
+
+The mapping symbol should set the type to `STT_NOTYPE`, binding to `STB_LOCAL`,
+and the size of symbol to zero.
+
+The mapping symbol for data(`$d`) indicates the start of a sequence of data bytes.
+
+The mapping symbol for instruction(`$x`) indicates the start of a sequence of
+instructions.
+and it has an optional ISA string, which means the following code regions are
+using ISA is different than the ISA recorded in the arch attribute;
+the ISA information will used until the next instruction mapping symbol;
+an instruction mapping symbol without ISA string means using ISA configuration
+from ELF attribute.
+
+Format and rule of the optional ISA string are same as `Tag_RISCV_arch`, must
+having explicit version, more detailed rule please refer to <<Attributes>>.
+
+NOTE: The use case for mapping symbol for instruction(`$x`) with ISA information
+is used with ifunc, e.g. libraries are built with `rv64gc`, but few functions
+like memcpy provides two versions, one built with `rv64gc`, and one built with
+`rv64gcv`, and select by ifunc mechanism at run-time; however, the arch
+attribute is recording for minimal execution environment requirements, so the
+ISA information from arch attribute is not enough for the disassembler to
+disassemble the `rv64gcv` version correctly.
+
 == Linker Relaxation
 
 At link time, when all the memory objects have been resolved, the code sequence


Nelson1225:
Hi Guys,

This PR LGTM, thanks for Kito's help.

The v2 proposed patch in binutils is as follows,
https://sourceware.org/pipermail/binutils/2021-July/117316.html

However, compared to the current implementation, there are some issues that might be worth discussing.

* $d<N>, N is data length.

The length is only attached when the data is added by cons_worker.  List all possible cases as follows,
`.byte`, `.hword`, `.int`, `.long`, `.octa`, `.quad`, `.short`, `.word`, `.2byte`, `.4byte`, `.8byte`, `.dc`, `.dc.a`, `.dc.b`, `.dc.l`, `.dc.w`, `.rva`.

For the .fill data directive, or rs_align, I still use a $d without data length to mark the region, since it is hard to estimate the length when adding the mapping symbols.  Besides, the initial purpose is to display the code which are written by users more closely, the .fill data are originally a directive to represent a data region, so add the data length to the $d seems redundant.

Therefore, it would be great if we can add - `$d<N>, N is 1, 2, 4, 8`.

* $a1 for odd byte code alignment, $a for a region of code alignment.

Consider the following case, which is compiled by the old toolchain without the mapping symbols.
$ cat tmp.s
        .text
        .option norvc
        .option norelax
        .byte   1
        .align  3
        nop
$ riscv64-unknown-elf-as tmp.s -o tmp.o
$ riscv64-unknown-elf-objdump -d tmp.o
0000000000000000 <.text>:
   0:   0001                    nop
   2:   0001                    nop
   4:   00000013            nop
   8:   00000013            nop
   c:   00000013            nop

There are two problems here,
1. The `.byte` + part of `.align 3` is recognized to a c.nop instruction.
2. c.nop is illegal since `.option norvc` is used.

But if we add mapping symbols for the case, then we can get the more friendly result,
0000000000000000 <.text>:
   0:   01                      .byte   0x01
   1:   00                      align.byte
   2:   0001                  align.nop
   4:   00000013          align.nop
   8:   00000013          nop
   c:   00000013          align.nop

$a1 is added at address 0x1, which is the start of the `.align 3`.  And then $a is added at 0x1 + 0x1 = 0x2.  If we don't add the $a1, then we will get the following result, 
0000000000000000 <.text>:
   0:   01                      .byte   0x01
   1:   0100                   align.addi      s0,sp,128
   3:   1300                   align.addi      s0,sp,416
   5:   0000                   align.unimp
   7:   1300                   align.addi      s0,sp,416
   9:   0000                   unimp
   b:   1300                   addi    s0,sp,416
   d:   0000                   align.unimp

Umm seems like objdump decode the `.align.byte` as a part of instruction `addi`.  However, we also can improve the objdump, so maybe it could dump the correct results without the $a1 mapping symbol.  I believe this is feasible, but use $a1 is quit easy, and no harm, at least to me.

* The documentation for -D says that on ARM platforms -D should disassemble instructions.  Therefore, they dump data by searching mapping symbols only when the DISASSEMBLE_DATA is set.  That causes the mapping symbols are useless for -D.  I am not sure if any riscv document had mentioned this, so I prefer to keep them useful, but only for text sections.  Otherwise, dump other sections as data.

jrtc27:
align.byte and align.nop are not valid mnemonics/directives. There is no need for special markers for alignment bytes, just fix the disassembler to correctly handle:
```
$d:
    .byte 1
$x:
    .byte half_of_a_c_nop(?)
    c.nop
```
i.e. don't try to disassemble instructions that are at odd addresses after data regions (or anything other than 0 mod 4 if RVC isn't present).

riscv-elf.md
```diff
@@ -1086,6 +1087,37 @@ Tag_RISCV_priv_spec contains the major/minor/revision version information of
...
+Mapping symbol is a special symbol class used for assist disassembler to
+having better knowledge of binary, it could be used for distinguish the
+code and data region.
+
+Symbol Name | Meaning
+:---------- | :-----------------------------------------------------------
+$d          | Start of a sequence of data.
+$d<N>       | Start of a sequence of N-byte data.
+$x          | Start of a sequence of instructions.
+$x<ISA>     | Start of a sequence of instructions with <ISA> extension.
```
jrtc27, line 1101: You don't need length information for data; it doesn't exist for real `.data` sections, it doesn't need to exist for `.text` either, just put a `$d` at the start and a `$x` at the end.

You also need to allow the `$d.foo`/`$x<ISA>.foo` form for cases where manual mapping symbols are being added to assembly files, as otherwise you get duplicate symbol definitions. The `.foo` is used to make them all unique, but is completely ignored.

riscv-elf.md
```diff
@@ -1086,6 +1087,37 @@ Tag_RISCV_priv_spec contains the major/minor/revision version information of
...
+
+Mapping symbol for data(`$d`) means following region are data,
+and it has an optional length information, for more precisely present
+the original data layout.
+
+Mapping symbol for instruction(`$x`) means following region are instructions,
+and it has an optional ISA string, which means following code region are using
```
jrtc27, line 1108: You need to state what the rules are for the ISA string; is it anything valid for -march= (ie default versions, implied extensions, etc all supported) or is it like the .RISCV.attributes section where it must be the full exploded string?

riscv-elf.md
kito-cheng: > You also need to allow the $d.foo/$x<ISA>.foo form for cases where manual mapping symbols are being added to assembly files, as otherwise you get duplicate symbol definitions. The .foo is used to make them all unique, but is completely ignored.

Suffix is not needed since mapping symbol local symbol, that could verified by AArch64 or ARM.

jrtc27: You can't specify the same symbol twice in the same assembly file. That's what the suffix is for.

kito-cheng: > You don't need length information for data; it doesn't exist for real .data sections, it doesn't need to exist for .text either, just put a $d at the start and a $x at the end.

That was designed for text section, used for mixing data and instruction, here is an example[1] from linux kernel, there is few data are put in text, and data having `.word`(32 bits) and `.dword`(64 bits), length information can provide better disassemble output here. 

[1] https://github.com/torvalds/linux/blob/master/arch/riscv/kernel/head.S#L52

kito-cheng:
> The objudmp -D make mapping symbols useless for ARM/AARCH64.
The documentation for -D says that on ARM platforms -D should disassemble instructions. Therefore, they dump data by searching mapping symbols only when the DISASSEMBLE_DATA is set. That causes the mapping symbols are useless for -D. I am not sure if any riscv document had mentioned this, so I prefer to keep them useful, but only for text sections. Otherwise, dump other sections as data.

I would prefer don't write down the behavior, it's a hit, how to interpreter that is dependent on implementation.

riscv-elf.md
jrtc27: But if I do:
```
.data
    .dword 0
    .byte 1
```
then there is no length information there. Having mapping symbols to switch between code and data in .text is sufficient to achieve parity with how .data gets dumped. This adds unnecessary complexity.

MaskRay:
I do see a use case for `$d<N>` yet.

I have some experience with the llvm-objdump and llvm-symbolizer. llvm-symbolizer just excludes mapping symbols. llvm-objdump adds mapping symbols of the current section into a sorted vector and performs a binary search. `$d<N>` will be ignored anyway. If you don't ignore `$d<N>`, you will need additional verification complexity I don't know will be justified.

---

I cannot find `$d.<suffix>` `$x.<suffix>` `$t.<suffix>` usage in a hand-written assembly file other than lld/gas testsuites. I think it is fine not to define it for now.

Nelson1225:
Thanks for your suggestions, @jrtc27 @MaskRay and @kito-cheng, I have sent v3 series of patches as follows,
https://sourceware.org/pipermail/binutils/2021-July/117348.html

Compared to v2,

* Removed the data mapping symbols with data size.

* Removed the alignment mapping symbols, $a and $a1.

* If the alignment have odd bytes spaces and relaxation
  is disable, we usually remain an odd byte 0x00 at the
  start of alignment, and then fill nops for the remianing
  spaces.  Therefore, add a $d mapping symbol for the odd
  byte, and then add a $x at $d + 0x1.  This behavior is
  same as Arm and Aarch64.

Thanks
Nelson

riscv-elf.md
kito-cheng: `$d<N>` is removed, I agree that it's unnecessary complexity, and might increase symbol table.

kito-cheng:
Updates:
- Remove `$d`
- add rule for `$x<ISA>`

riscv-elf.md
```diff
@@ -1086,6 +1087,38 @@ Tag_RISCV_priv_spec contains the major/minor/revision version information of
...
+$x          | Start of a sequence of instructions.
+$x<ISA>     | Start of a sequence of instructions with <ISA> extension.
+
+Mapping symbol for data(`$d`) means following region are data.
+
+Mapping symbol for instruction(`$x`) means following region are instructions,
+and it has an optional ISA string, which means following code region are using
+ISA different than the ISA recorded in arch attribute, the ISA information will
+used until next instruction mapping symobl; an instruction mapping symobl
+without without ISA string means using ISA configuration from ELF attribute.
```
jim-wilson, line 1108:
```suggestion
without ISA string means using ISA configuration from ELF attribute.
```

riscv-elf.md
jim-wilson: Below it says that the string must be same as Tag_RISCV_arch.

jim-wilson (commented):
Looks OK to me.  Could use a little English improvement, but that can wait until enough people are OK with it.  No need to rewrite the English multiple times.

kito-cheng:
Changes:
- Mark use case part as `NOTE`.
- Add one more note about toolchain implementation, allow linker to merge mapping symbol and allow strip tool to strip mapping symbol.

kito-cheng:
Changes:
- Rebase to `master`

kito-cheng:
Changes:
- Rebase to master

kito-cheng:
@jrtc27 @MaskRay could you take a look? :)

riscv-elf.adoc
```diff
@@ -760,6 +760,43 @@ Tag_RISCV_priv_spec contains the major/minor/revision version information of
 the privileged specification. It will report errors if object files of different
 privileged specification versions are merged.
 
+=== Mapping Symbol
+
+Mapping symbol is a special symbol class used for assist disassembler to
```
MaskRay, line 765: I'd say that a section can have a mixture of code and data. Mapping symbols describe the boundaries.

riscv-elf.adoc
```diff
@@ -760,6 +760,43 @@ Tag_RISCV_priv_spec contains the major/minor/revision version information of
...
+having better knowledge of binary, it could be used for distinguish the
+code and data region.
+
+[%autowidth]
+|===
+| Symbol Name | Meaning
+| $d          | Start of a sequence of data.
+| $x          | Start of a sequence of instructions.
+| $x<ISA>     | Start of a sequence of instructions with <ISA> extension.
```
MaskRay, line 774: Better to follow arm's lead by requiring a bot: `$x.<any>`

riscv-elf.adoc
```diff
@@ -760,6 +760,43 @@ Tag_RISCV_priv_spec contains the major/minor/revision version information of
...
+
+Mapping symbol for data(`$d`) means following region are data.
+
+Mapping symbol for instruction(`$x`) means following region are instructions,
+and it has an optional ISA string, which means following code region are using
+ISA different than the ISA recorded in arch attribute, the ISA information will
+used until next instruction mapping symobl; an instruction mapping symobl
+without ISA string means using ISA configuration from ELF attribute.
+
+Format and rule of the optional ISA string is same as Tag_RISCV_arch, must
```
MaskRay, line 785: Reserving `$x.<any>` is more useful than defining it just for ISA.
I am not sure specifying an ISA is useful, or needed to be defined now.

`$x` and `$d` have clear usefulness and can be added today.

riscv-elf.adoc
```diff
@@ -760,6 +760,43 @@ Tag_RISCV_priv_spec contains the major/minor/revision version information of
...
+
+NOTE: The use case for mapping symbol for instruction(`$x`) with ISA information is
+used with ifunc, e.g. library are built with `rv64gc`, but few functions
+like memcpy provide two version, one built with `rv64gc`, and one built with
+`rv64gcv`, and select by ifunc mechanism at run-time; however the arch
+attribute is recording for minimal execution environment requirement, so the ISA
+information from arch attribute isn't enough for disassembler to disassemble the
+`rv64gcv` version correctly.
+
+NOTE: For toolchain implementation, linker are permit to merge the adjacency
```
MaskRay, line 796: adjacent mapping symbols if they are of

MaskRay (changes requested):

riscv-elf.adoc
jrtc27: Changing the ISA for disassembly is the primary reason why this is needed on RISC-V (nobody really cares about constant pools that much otherwise we'd see mapping symbols on more architectures). We need a way to have, say, an RV64GC binary that has snippets of RV64GCV in it, and whilst you could introduce some kind of "max ISA" attribute to go alongside the current ISA attribute in .riscv.attributes (which is the "I need this to be able to run" attribute), that doesn't work once you have conflicting extensions (e.g. you might want F and Zfinx in the same binary, or the ratified V spec with XOldVSpecBecauseVendorImplementedAnEarlyDraftDespiteClearDocumentationEverywhereThisShouldNotBeDone).

MaskRay: Your description makes sense. The original wording based on ifunc scared me.

Still worth thinking how if we ever need `<any>`, where to place it. `$x<isa>.<any>`?

jrtc27: Yeah, the `.<any>` is a meaningless uniquifier, so would go on the end after the actually-significant ISA.

riscv-elf.adoc
kito-cheng: Updated, thanks :)

riscv-elf.adoc
kito-cheng: Added one sentence to allow an optional suffix uniquifier string.

kito-cheng:
Changes:
- Rebase
- Allow an optional suffix uniquifier string for mapping symbol name.
- Revise wording per @MaskRay's suggestion

kito-cheng:
Changes:
- Rebase
- Revise description of mapping symbol

kito-cheng:
Changes:
- Rebase
-  Wording and grammar tweaks.

kito-cheng:
We have landed mapping symbol on binutils site and having patch under review on LLVM land (https://reviews.llvm.org/D137417).

@MaskRay Do you mind give a blessing to this PR? :)

riscv-elf.adoc
```diff
@@ -1134,6 +1134,45 @@ Merge policy:::
 The linker should report errors if object files of different privileged
 specification versions are merged.
 
+
+=== Mapping Symbol
+
+The section can have a mixture of code and data or code with different ISAs;
+mapping symbol is a special symbol class used to describe the boundaries.
```
MaskRay, line 1141: I haven't seen a term named "symbol class". Avoid it?

`; mapping symbol ...` => `. A number of symbols, named mapping symbols, describe the boundaries`.

Mention that they have type `STT_NOTYPE`, binding `STB_LOCAL`, and size zero.

riscv-elf.adoc
```diff
@@ -1134,6 +1134,45 @@ Merge policy:::
...
+
+[%autowidth]
+|===
+| Symbol Name | Meaning
+| $d          | Start of a sequence of data.
+| $x          | Start of a sequence of instructions.
+| $x<ISA>     | Start of a sequence of instructions with <ISA> extension.
+|===
+
+The mapping symbol for data(`$d`) means the following region is data.
```
MaskRay, line 1151: indicates the start of a sequence of data bytes.

riscv-elf.adoc
```diff
@@ -1134,6 +1134,45 @@ Merge policy:::
...
+
+The mapping symbol for instruction(`$x`) means the following region are
```
MaskRay, line 1153: indicates the start of a sequence of instructions.

riscv-elf.adoc
```diff
@@ -1134,6 +1134,45 @@ Merge policy:::
...
+instructions,
+and it has an optional ISA string, which means the following code regions are
+using ISA is different than the ISA recorded in the arch attribute;
+the ISA information will used until the next instruction mapping symbol;
+an instruction mapping symbol without ISA string means using ISA configuration
+from ELF attribute.
+
+Format and rule of the optional ISA string are same as Tag_RISCV_arch, must
```
MaskRay, line 1161: backticks around `Tag_RISCV_arch`

The format of the optional ISA string

riscv-elf.adoc
```diff
@@ -1134,6 +1134,45 @@ Merge policy:::
...
+
+NOTE: The use case for mapping symbol for instruction(`$x`) with ISA information
+is used with ifunc, e.g. libraries are built with `rv64gc`, but few functions
+like memcpy provides two versions, one built with `rv64gc`, and one built with
+`rv64gcv`, and select by ifunc mechanism at run-time; however, the arch
+attribute is recording for minimal execution environment requirements, so the
+ISA information from arch attribute is not enough for the disassembler to
+disassemble the `rv64gcv` version correctly.
+
+NOTE: For toolchain implementation, the linker is permitted to merge the
```
MaskRay, line 1172: Does GNU ld combine adjacent mapping symbols? How GNU strip removes mapping symbols is complex. 

That would be a weird thing to do. Consider dropping this sentence.

MaskRay (approved):

riscv-elf.adoc
MaskRay: `$x.<ISA>`

jrtc27: The <any> is a uniquifier. The <ISA> is the ISA to use. $x.<ISA> would then be ambiguous if we permit Arm's uniquifier approach where you can add anything after a . to work around assemblers that don't like duplicate symbols.

riscv-elf.adoc
kito-cheng: Thanks, added one more sentence to describe the type, binding and size.

riscv-elf.adoc
kito-cheng: Removed, currently binutils didn't implement both.

riscv-elf.adoc
kito-cheng: Yeah, I would prefer keep same since we have slightly difference between ARM's mapping symbol, or maybe we could add `$x.<any>` and `$x<ISA>.<any>` to allow we have uniquifier like ARM?

kito-cheng:
Changes:
- Rebase.
- Mention mapping symbol has type STT_NOTYPE, binding STB_LOCAL, and size zero.
- Wording tweaks.

kito-cheng:
Got approve from LLD and binutils maintainer, gonna merge this :)

riscv-elf.adoc
jrtc27: We should IMO

kito-cheng: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/372

PR #198, opened 2021-07-09
Add a note to call out that Zfinx always follows the "soft-float" ABI

jrtc27:
Closes: #166

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -322,6 +322,11 @@ operating mode.
 The \*F ABIs require the \*F ISA extension, the \*D ABIs require the \*D ISA
 extension, and the LP64Q ABI requires the Q ISA extension.
 
+> NOTE:
+> This means code targeting the Zfinx extension always uses the ILP32, ILP32E
+> or LP64 integer calling-convention only ABIs as there is no dedicated
+> hardware floating-point register file.
+
 ## <a name=default-abis></a>Default ABIs
 
 While various different ABIs are technically possible, for software


jim-wilson (approved):

riscv-elf.md
```diff
@@ -319,6 +319,11 @@ operating mode.
 The \*F ABIs require the \*F ISA extension, the \*D ABIs require the \*D ISA
 extension, and the LP64Q ABI requires the Q ISA extension.
 
+> NOTE:
+> This means code targeting the Zfinx extension always uses the ILP32, ILP32E
+> or LP64 integer calling-convention only ABIs as there is no dedicated
```
luismarques, line 324: "integer calling-convention only ABIs" -> "integer only calling-convention ABIs"?

jrtc27: That does seem like a better way round, but I copied the description used for ILP32 and LP64 above, which have it this way round

jrtc27: Currently the document is written as "there is an integer calling convention and you can optionally add a floating point calling convention to it", which is a slightly weird way of thinking of it that doesn't quite match what is actually happening

kito-cheng (approved):
LGTM.

kito-cheng:
But let us wait one more LGTM from LLVM community.

asb:
> But let us wait one more LGTM from LLVM community.

LGTM from the LLVM side.

jrtc27:
Rebased to resolve merge conflict. Ok to merge?

Issue #160, opened 2020-08-14
GPREL and TPREL register use


sorear:
The psABI has this to say about GPREL_I and GPREL_S:

---

**GP-Relative Relocations**: If `symbol` is within the range of a signed 12-bit immediate offset from `__global_pointer$`, then the address can be loaded with a single instruction which has one relocation, a `R_RISCV_GPREL_I` or `R_RISCV_GPREL_S`.

The instruction is an I-Type instruction (add immediate or load) with an `R_RISCV_GPREL_I` or an S-type instruction (store) with an `R_RISCV_GPREL_S` relocation. The following assembly show loading a gp-relative address:

```
     addi a0, gp, 0          # R_RISCV_GPREL_I (symbol)
```

This relies on the value of `__global_pointer$` being loaded into `gp` (aka `x3`). This can be used by linker relaxation to delete the `lui` instruction.

---

Note that nothing in there specifies that the gp register must be used until after a code sample is given that uses gp.  The simplest interpretation of the text in the psABI is that any register may be used with `R_RISCV_GPREL_I` and `R_RISCV_GPREL_S` relocations, as long as that register contains `__global_pointer$` prior to the relocated instruction.

ld.bfd implements something different:

https://github.com/riscv/riscv-elf-psabi-doc/blame/master/riscv-elf.md#L641-L655

ld.bfd has an undocumented feature: gprel relocations can be used to access the zero page (±2 KiB).  However, the way this is implemented overwrites `rs1` in any instruction which uses a gprel relocation, so `R_RISCV_GPREL_I` could not be used with `__global_pointer$` materialized in any other register.  TPREL functions similarly and overwrites rs1 with tp.

lld does not implement GPREL or TPREL at all.

---

Historically, conflicts between the psABI and binutils have been resolved in favor of binutils to avoid compatibility problems.  In this case, I think the compatibility concern is negligible; neither gcc nor clang generate GPREL_I, GPREL_S, TPREL_I, or TPREL_S relocations, and there is no assembly syntax for them, so the only way they could exist in object files is if someone else implemented object file writing against the ld.bfd source code while ignoring the psABI.

**I would like to add a sentence to the psABI stating that any register can be used with the short GPREL and TPREL relocations as long as that register holds the appropriate base value, and modify ld.bfd to support this.**  This is simpler to describe, simpler to implement, and provides some additional flexibility (the compact code model as currently proposed has a use case for this).

Yes/No?

[...]

PR #199, opened 2021-07-09
Remove [GT]PREL_[IS] internal binutils relocations

jrtc27:
These have never been emitted by GCC or Clang, nor has there ever been
assembly syntax in either toolchain to create them. These are fake
internal relocations inside binutils used as part of its implementation
of linker relaxation, and so should not be part of the psABI. They
should be free to be reallocated for psABI uses (which can include
reintroducing them, if there is a need to expose them to assembly and
code generation).

Closes: #160


aswaterman:
Sounds right to me.

sorear:
LGTM

kito-cheng:
LGTM

aswaterman:
I think that's enough LGTMs!

Issue #157, opened 2020-07-08
Reserve a subset of e_flags for non-standard extensions


jrtc27:
Currently the specification says:

> Until such a time that the Reserved bits (0xffffffe0) are allocated by future versions of this specification, they shall not be set by standard software.

This leaves no bits free for use by non-standard extensions. On [CHERI-RISC-V] we currently allocate two bits (although we'd like to make one of those redundant at some point), and those were chosen to be 0x00010000 and 0x00020000 in order to leave space for new bits allocated in the base ABI, whilst also leaving space for us to allocate new bits for CHERI-RISC-V if/when needed (and being at a nice round boundary). The RISC-V attributes section helps alleviate contention on the ELF flags, but when defining drastically-different ABIs it is useful to have bits in e_flags for that, as that is generally what dynamic loaders have to hand, with RISC-V attributes parsing a whole pile of RISC-V-specific complexity that you don't want in such crucial low-level code.

Please define a range of bits reserved for use by non-standard extensions. Ideally this would overlap with our allocated bits, but if it doesn't make sense to do so then we can perform a (painful) migration to newer standards-conforming flag bits.

[CHERI-RISC-V]: https://github.com/CTSRD-CHERI/cheri-elf-psabi/blob/master/riscv.md#file-header

lenary:
It seems a proactive measure to do this, and would allow developers to work on prototypes of other additions to the psABI (such as the compact code model, or FDPIC).

I'm unsure how many bits to reserve for custom users - it would be good to work out what is reasonable but I don't have the time to quickly scan other psABIs now. I suggest either the highest 8 or 16 bits to be used, as those seem like 'round' numbers, but as I said, I'm unsure.

I would like to hear @jim-wilson and @ebahapo's opinions.

jrtc27:
Yeah, 16 seems quite generous, but 8 could be a bit tight depending on what you're doing. I went with the first two bits of the higher 16 bits because, at the time, RISC-V attributes support didn't exist in LLVM so I thought I might need rather more, and it was a "nice" way to split it. I think it depends how many more bits are earmarked for future _standard_ extensions. I anticipate wanting about 5 bits, so 8 would be fine for us, just annoying to shift from bit 16+ to bit 24+.

aswaterman:
I think 16 is too many.  We can always turn more reserved bits into custom bits in the future, but we can't go backwards.  Bits 24-31 seems OK to me, if possibly on the overly generous side.

[...]

PR #200, opened 2021-07-09
Reserve e_flags bits 24 through 31 for custom use

jrtc27:
Whilst here, correctly pluralise some headers and use consistent
alignment.

Closes: #157

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -469,9 +469,9 @@ rules about 2&times;XLEN aligned arguments being passed in "aligned" register pa
 * e_flags: Describes the format of this ELF file.  These flags are used by the
   linker to disallow linking ELF files with incompatible ABIs together.
 
-   Bit 0 | Bit  1 - 2 | Bit 3 | Bit 4 | Bit 5 - 31
-  -------|------------|-------|-------|------------
-   RVC   | Float ABI  |  RVE  |  TSO  | *Reserved*
+   Bit 0 | Bits 1 - 2 | Bit 3 | Bit 4 | Bits 5 - 23 | Bits 24 - 31
+  -------|------------|-------|-------|-------------|---------------------------
+   RVC   | Float ABI  | RVE   | TSO   | *Reserved*  | *Non-standard extensions*
 
   * EF_RISCV_RVC (0x0001): This bit is set when the binary targets the C ABI,
     which allows instructions to be aligned to 16-bit boundaries (the base RV32
@@ -497,9 +497,18 @@ rules about 2&times;XLEN aligned arguments being passed in "aligned" register pa
   * EF_RISCV_TSO (0x0010): This bit is set when the binary requires the RVTSO
     memory consistency model.
 
-  Until such a time that the *Reserved* bits (0xffffffe0) are allocated by
+  Until such a time that the *Reserved* bits (0x00ffffe0) are allocated by
   future versions of this specification, they shall not be set by standard
-  software.
+  software. Non-standard extensions are free to use bits 24-31 for any purpose.
+  This many conflict with other non-standard extensions.
+
+  > NOTE:
+  > There is no provision for compatibility between conflicting uses of the
+  > e_flags bits reserved for non-standard extensions, and many standard RISC-V
+  > tools will ignore them. Do not use them unless you control both the
+  > toolchain and the operating system, and the ABI differences are so
+  > significant they cannot be done with a .RISCV.attributes tag nor an ELF
+  > note, such as using a different syscall ABI.
 
 ## <a name=sections></a>Sections
 


riscv-elf.md
```diff
@@ -469,9 +469,9 @@ rules about 2&times;XLEN aligned arguments being passed in "aligned" register pa
 * e_flags: Describes the format of this ELF file.  These flags are used by the
   linker to disallow linking ELF files with incompatible ABIs together.
 
-   Bit 0 | Bit  1 - 2 | Bit 3 | Bit 4 | Bit 5 - 31
-  -------|------------|-------|-------|------------
-   RVC   | Float ABI  |  RVE  |  TSO  | *Reserved*
+   Bit 0 | Bits 1 - 2 | Bit 3 | Bit 4 | Bits 5 - 23 | Bits 24 - 31
+  -------|------------|-------|-------|-------------|---------------------------
+   RVC   | Float ABI  | RVE   | TSO   | *Reserved*  | *Non-standard extensions*
```
jrtc27, line 474: I dropped the "Reserved for" as otherwise the table gets rather wide

jim-wilson:
I think that there are practical issues with this suggestion.  The ELF Header flags are one of the first things checked when trying to figure out if we can run this binary, but if the ELF Header flags contain bits that can only be parsed if something else is read first then the bits aren't usable for their normal purpose.  It isn't clear why anyone would want to put vendor specific bits in eflags.  Is there any other target that does this?  Just checking binutils, I don't see an obvious one that is allowing unknown bits in eflags.

Something that could work is if the bits were dependent on the EI_OSABI byte in the ident string, e.g. if ELFOSABI_LINUX and ELFOSABI_FREEBSD used the bits for different purposes then that would be OK, as then we can still parse the bits without reading anything other than the elf header.  Or if we had two different machine numbers, e.g. EM_RISCV and another one, and the value of the bits depended on which machine number is being used, but a second machine number is probably best avoided if possible.

Anyways, I think bits that can't be parsed without reading something outside the elf header is a mistake.  There are lots of other ways to add vendor specific support, like vendor specific program headers, or sections, or relocations, or vendor specific extensions to the elf attribute which is designed to be extensible.  We don't need vendor specific bits in eflags.

jrtc27:
This proposal is what was discussed in #157. We use them downstream for CHERI to distinguish between plain RISC-V binaries and pure-capability CHERI-RISC-V binaries. Just like the float ABI, different binaries are totally incompatible, but even more so; the code is totally different, even if it does resemble RISC-V a lot. Even Arm, after dragging their heels and not using an ELF flag for Morello for over a year, relented and reserved one in https://github.com/ARM-software/abi-aa/pull/101. In FreeBSD, which ELF image activator (32-bit, 64-bit and, in our case, 64-bit CHERI pure-capability) gets used is decided very early on based solely on the ELF header, and it's only once it's decided on an image activator that it then starts to look at program headers, so this information _has_ to be in the ELF header. Not to mention that so many tools need to know the information that having an ELF note or similar is just a complete pain for no gain.

Whatever happens in the RISC-V psABI, we will just continue to use an ELF flag downstream because that is by far the best solution for our needs. I'd just like to use a bit that I know won't be used by the psABI in future (without abusing my power to stop people allocating the bits we're using and thereby giving my downstream special treatment). I appreciate the point that discovering what the bits mean isn't really possible, but having them be meaningful in that kind of context just isn't a goal. This is for "I have a system where I absolutely need an ELF flag and control both the OS and toolchain producing them, and have no need to support any other such non-standard extensions". To be honest, I don't _want_ people to use these bits, because the situations in which you should be using them are exceedingly rare, only for something as radical as CHERI (I see these more as "research" than "vendor" bits...), but it is a part of the psABI that currently does not have a provision for custom extensions that need to reserve bits/encodings and so I feel there should be _something_ there for such cases like ours.

I could add a big blob of "Do not use these unless you absolutely need to. There is no provision for compatibility between conflicting uses, and many standard RISC-V tools will ignore them. Only use these if you control the operating system and the toolchain, and the ABI differences are so significant that they cannot be done with an ELF note, such as if you require use of a different syscall ABI." kind of text if that would help?

jim-wilson:
In the ARM case it is a Morello specific bit, not a reserved extension bit.  The GNU toolchain doesn't support this Morello bit yet.

Yes, I think adding a note to recommend not using the bits except in special cases would help.

jrtc27:
> In the ARM case it is a Morello specific bit, not a reserved extension bit. The GNU toolchain doesn't support this Morello bit yet.

Well, there is a commitment from them not to use the bit in the main AAELF64 spec, which they can do because both are controlled solely by Arm. I can't just reserve a bit for CHERI-RISC-V and commit to not allocating it in the psABI because that's an abuse of power, hence why I want to go through the officials channels to ensure there is a bit we can freely use without reserving it specifically for CHERI-RISC-V (I mean, if people would _rather_ I do that instead then I can but I rather doubt that's more palatable...).

And yes, the Morello GNU toolchain lags LLVM in many respects, it's just not a particularly useful artefact at this point given the distinct lack of a working GCC (which they are working on, but is a large undertaking to replicate all the work we did to the internals of LLVM for CHERI and is a way away from being usable, let alone efficient), there's zero reason why you'd choose to use LLVM with GNU binutils, the basic opcodes BFD parts are primarily there for GDB at the moment, with the GNU as and ld parts really only starting to be useful for the GCC work (I don't know of anything else using them), thus all the development happens on the LLVM toolchain with their binutils playing catch-up when people notice (which makes a change to have it that way round for once!).

kito-cheng:
Although I am not flavor to reserve e_flags for custom/vendor use, but we have `.riscv.attribute` which has lots space to extend, so LGTM once it add note.

jim-wilson (approved):

kito-cheng (approved):

Issue #126, opened 2019-12-18
Clarity on weak, undefined symbols in non-PIC code


bsdjhb:
The FreeBSD kernel uses weak symbols in its uart(4) driver as well as in numerous other cases for linker sets.  These symbols are not functions invoked via call, but are pointers to data.  In some cases these weak symbols may be undefined when linking the kernel which FreeBSD assumes will result in a value of NULL.  This has been true with all other architectures FreeBSD has run on to date.  The bfd linker also seemingly enforces this.

When linking the kernel with lld, I ran into linker errors due to the resolved symbol being out of range for the PC-relative relocations being used.  I proposed a patch to lld at https://reviews.llvm.org/D71100 but there are some questions of if the semantics in that change are the desired semantics.

From comments in #98 it seems that it is assumed (but not stated) that R_RISCV_CALL relocations against undefined weak symbols should be resolved to an absolute address of 0 rather than a PC-relative address.

However, for data addresses, it is conceivable that a data address should be indirected through a GOT entry to permit overriding (e.g. if a symbol is provided at runtime via LD_PRELOAD or a newer version of a library than was available at the time a binary was linked) while still permitting a value of zero to be used if the binary is running at a "high" virtual address.

jim-wilson:
Given the testcase
extern int i __attribute__ ((weak));
int main (void) { return i; }
compiling with gcc -O I get
000000000001043e <main>:
   1043e:	000007b7          	lui	a5,0x0
   10442:	0007a503          	lw	a0,0(a5) # 0 <_PROCEDURE_LINKAGE
_TABLE_-0x10360>
   10446:	8082                	ret
compiling with gcc -O -fpie I get
000000000001046e <main>:
   1046e:	00002797          	auipc	a5,0x2
   10472:	bb27b783          	ld	a5,-1102(a5) # 12020 <i>
   10476:	4388                	lw	a0,0(a5)
   10478:	8082                	ret
So it is indirecting through the got.  This fails with a segfault when I run it.  But if I preload a shared library that defines the variable i, then it returns that value.
hifiveu017:1073$ ./a.out
Segmentation fault (core dumped)
hifiveu017:1074$ cat tmp2.c
int i = 10;
hifiveu017:1075$ gcc --shared -fpic tmp2.c -o libtmp.so
hifiveu017:1076$ LD_PRELOAD=`pwd`/libtmp.so ./a.out
hifiveu017:1077$ echo $?
10
[...]

PR #201, opened 2021-07-09
Advise, but don't yet require, use of the GOT for medany weak undef

jrtc27:
Closes: #126

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -371,6 +371,17 @@ a signed 32-bit offset, relative to the value of the `pc` register,
 can be produced.
 This code model is position independent.
 
+As a special edge-case, undefined weak symbols must still be supported, whose
+addresses will be 0 and may be out of range depending on the address at which
+the code is linked. Any references to possibly-undefined weak symbols should be
+made indirectly through the GOT as is used for position-independent code. Not
+doing so is deprecated and a future version of this specification will require
+using the GOT, not just advise.
+
+> NOTE: This is not yet a requirement as existing toolchains predating this
+> part of the specification do not adhere to this, and without improvements to
+> linker relaxation support doing so would regress performance and code size.
+
 # <a name=c-types></a>C/C++ type details
 
 ## <a name=c-type-sizes></a>C/C++ type sizes and alignments


jrtc27:
Cc @MaskRay (can't request a review from you as you're not a member of the repo... might be worth granting you explicit read access, I think that lets you be added as a reviewer?)

riscv-elf.md
```diff
@@ -365,6 +365,17 @@ a signed 32-bit offset, relative to the value of the `pc` register,
 can be produced.
 This code model is position independent.
 
+As a special edge-case, weak undefined symbols must still be supported, whose
```
MaskRay, line 368: Nit: while both word orders are used, AIUI the order `undefined weak` appears to be more common among specifications (at least: ELF abi, Oracle's document, aarch64 abi)

MaskRay (approved):
LG.

I think "undefined weak" should not be on a performance bottleneck of any applications, so using GOT for compiler codegen and leaving the optional instruction rewriting for the linker (even if it is implemented in no implementation right now) is perfectly fine. We just can't help users who want to shoot themselves in the foot...

riscv-elf.md
```diff
@@ -365,6 +365,17 @@ a signed 32-bit offset, relative to the value of the `pc` register,
...
+addresses will be 0 and may be out of range depending on the address at which
+the code is linked. Any references to possibly-undefined weak symbols should be
+made indirectly through the GOT as is used for position-independent code. Not
+doing so is deprecated and a future version of this specification will require
+using the GOT, not just advise.
+
+> NOTE: This is not yet a requirement as existing toolchains predating this
+> part of the specification do not adhere to this, and without improvements to
+> linker relaxation support doing so would regress performance and code size.
```
MaskRay, line 377: I see that the "regress performance and code size" part is for @jim-wilson ....
I still hope that we can just remove this sentence as I don't see how usage of undefined weak symbols become a bottleneck of anything...

aswaterman: Loop-invariant code motion is essentially irrelevant here.  Jim's contention is about code size, not dynamic instruction count.  Jim's concern remains relevant.

jim-wilson: The newlib crt0.S contains an undefined weak.  This means that anyone building a riscv64-elf compiler from riscv-gnu-toolchain and compiling the official RISC-V embench testsuite will see code size increases.  That is a big problem in the embedded world.

There is no problem in the Linux world.  Anyone that cares about code size or performance of undefined weak is probably confused.

The GNU toolchain has been supporting weak for static linking since the 1990s, and this is commonly used in embedded programming.  I don't know if clang supports this.  But in the GNU world this is a feature that needs to work efficiently.

Having GNU ld relaxations to optimize the code will mitigate the code size and performance regressions, but then there is the problem of who will write the code and when.  There is a shortage of people actively working upstream in the GNU toolchain.  So until that is done, we can't mandate the change, and there is no way to know when the work will be done.

jim-wilson:
This is only OK to me if the change remains optional until someone writes the gnu ld relaxation support to optimize the result back to what it was before the ABI change.

riscv-elf.md
jrtc27: Hang on, are we talking about a single reference to a single variable? I.e. at worst a code size increase of 6 bytes (2-byte compressed to 2*4-byte uncompressed) and data size increase of 4 bytes on RV32 and 8 bytes on RV64?

riscv-elf.md
jrtc27: Fixed

kito-cheng:
I think we all agree that would be better way for non-embedded world part like Linux, but I am in the same line with Jim for the embedded world part - few bytes is matter in the embedded world, I am OK with this only for optional and no deprecated in the ABI.

MaskRay:
Still LGTM.

jrtc27:
LLVM patch: https://reviews.llvm.org/D107280

jrtc27:
@jim-wilson Are you happy with the current wording?

jim-wilson (approved):

PR #202, opened 2021-07-20
Remove stub compact code model section

jrtc27:
This section is currently a draft, so should not feature in any official publication (and does not currently meet the requirements in the pull request policy so would never have been added were those in place at the time). Thus, until a concrete specification for it exists and is deemed to be a worthwhile addition, we should not talk about it and risk causing confusion. Especially since, as currently described, it's a new ABI, not just a new code model, due to the requirement that gp always be able to be used to access the GOT, which isn't necessarily the case currently for executables, and is never the case currently for shared libraries as far as linkers are concerned.

### DIFF
--- a/riscv-elf.md
+++ b/riscv-elf.md
@@ -365,16 +365,6 @@ a signed 32-bit offset, relative to the value of the `pc` register,
 can be produced.
 This code model is position independent.
 
-### Compact
-
-The compact code model allows the code to address the whole 64-bit address space,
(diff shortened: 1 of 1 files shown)


kito-cheng (approved):
Although we (SiFive) have PoC for compact code model but we didn't open most source yet, and I suppose it need more discussion, but we didn't have enough bandwidth on that yet, so I am OK to remove that now, and added back once that become more mature.

jim-wilson (approved):

Issue #183, opened 2021-03-27
R_RISCV_ALIGN friendly to linkers not supporting relaxation (riscv_relax_delete_bytes)


MaskRay:
In ELF, many relocation types capable of linker optimization are pure optional features: R_AVR_CALL, R_PPC64_PCREL_OPT, R_X86_64_GOTPCRELX, R_X86_64_REX_GOTPCRELX, TLS relocation types supporting GD->LE/GD->IE/LD->LE/etc optimizations. A linker not understanding the advanced feature can provide the basic feature and move on. However, `R_RISCV_ALIGN` is different.

Currently `R_RISCV_ALIGN` just encodes the expected alignment. Can it encode the actual bytes of NOPs beside the expected alignment?
Then linkers not supporting linker relaxation can happily accept `-mrelax` object files.

There are several schemes

* Encode the actual bytes of NOPs in the symbol part of the r_info field. The associated symbol can be absolute (`st_shndx==SHN_ABS`).
* Encode the actual bytes of NOPs in the high bits of the r_addend field.
* Introduce a new relocation type used together with `R_RISCV_ALIGN`.

The third approach is probably least favored. The first one should be compatible with existing implementations.

There would add additional complexity to the only linker implementation today: GNU ld.
Skimming `bfd/elfnn-riscv.c:riscv_relax_delete_bytes`, this probably won't be too difficult.

https://maskray.me/blog/2021-03-14-the-dark-side-of-riscv-linker-relaxation (Apologies for the title:) Someone said I probably should have postponed it to 05-04 😹)

kito-cheng:
@Nelson1225 @jrtc27 @PkmX @kuanlinchentw

kito-cheng:
I would suggest add new relocation for a relax-free-safe, the reason is I think it would require different behavior of NOP padding, we are always padding N - 2 byte of NOP before the `.align` or `.p2align`, no matter it's already align or not, so that means we might not satisfy the alignment requirement if there is no NOP deletion.

e.g.
```
$ cat foo.s
foo:
        add a0, a0, a0
        add a0, a0, a0
        .align 3 # align to 8 byte
        sub a0, a0, a0
        ret
```
But currently we'll padding 6 byte NOP before the align point, the `sub` isn't align to 8 byte if we don't delete any NOP!
```
$ riscv64-unknown-elf-as -mrelax foo.s -o foo.o -march=rv64gc
[...]

PR #210, opened 2021-08-25
Document the behavior of R_RISCV_ALIGN

kito-cheng:
Describe the behavior of R_RISCV_ALIGN relocation.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -683,6 +683,68 @@ label:
     sw t2, t0, %pcrel_lo(label)   # R_RISCV_PCREL_LO12_S (label)
 ----
 
+==== Relocation for Alignment
+
+The relocation type `R_RISCV_ALIGN` marks a location that must be aligned to
+`N`-bytes, where `N` is the smallest power of two that is greater than the value
+of the addend field, e.g. `R_RISCV_ALIGN` with addend value 2 means align to 4
+bytes, `R_RISCV_ALIGN` with addend value 4 means align to 8 bytes; this
+relocation is only required if the containing section has any `R_RISCV_RELAX`
+relocations, `R_RISCV_ALIGN` points to the beginning of the padding bytes,
+and the instruction that actually needs to be aligned is located at the point
+of `R_RISCV_ALIGN` plus its addend.
+
+To ensure the linker can always satisfy the required alignment solely by
+deleting bytes, the compiler or assembler must emit a `R_RISCV_ALIGN` relocation
+and then insert `N` - <<IALIGN>> padding bytes before the location where we need to
+align, it could be mark by an alignment directive like `.align`, `.p2align` or
+`.balign` or emit by compiler directly, the addend value of that relocation
+is the number of padding bytes.
+
+The compiler and assembler must ensure padding bytes are valid instructions
+without any side-effect like `nop` or `c.nop`, and make sure those instructions
+are aligned to IALIGN if possible.
+
+The linker may remove part of the padding bytes at the linking process to meet
+the alignment requirement, and must make sure those padding bytes still are
+valid instructions and each instruction is aligned to at least IALIGN byte.
+
+Here is example to showing how `R_RISCV_ALIGN` is used:
+[,asm]
+----
+
+0x0    c.nop           # R_RISCV_ALIGN with addend 2
+0x2    add t1, t2, t3  # This instruction must align to 4 byte.
+
+----
+
+
+NOTE: `R_RISCV_ALIGN` relocation is needed because linker relaxation can shrink
+preceding code during the linking process, which may cause an aligned location
+to become mis-aligned.
+
+NOTE: IALIGN[[IALIGN]] means the instruction-address alignment constraint. IALIGN is 4
+bytes in the base ISA, but some ISA extensions, including the compressed ISA
+extension, relax IALIGN to 2 bytes. IALIGN may not take on any value other than
+4 or 2. This term is also defined in `The RISC-V Instruction Set Manual` with a
+similar meaning, the only difference being it is specified in terms of the number
+of bits instead of the number of bytes.
+
+NOTE: Here is pseudocode to decide the alignment of `R_RISCV_ALIGN` relocation:
+[,python]
+----
+# input:
+#   addend: addend value of relocation with R_RISCV_ALIGN type.
+# output:
+#   Alignment of this relocation.
+
+def align(addend):
+  ALIGN = 1
+  while addend >= ALIGN:
+    ALIGN *= 2
+  return ALIGN
+----
+
 === Thread Local Storage
 
 RISC-V adopts the ELF Thread Local Storage Model in which ELF objects define


kito-cheng:
@MaskRay

kito-cheng:
Changes:

- Apply @jim-wilson's and @MaskRay's suggestion. 
- Rebase to master to fit new structure of document.

kito-cheng:
Changes:
- Minor fix typo.

riscv-elf.adoc
```diff
@@ -482,6 +482,24 @@ label:
     sw t2, t0, %pcrel_lo(label)   # R_RISCV_PCREL_LO12_S (label)
 ----
 
+==== Relocation for Alignment
+
+The relocation type `R_RISCV_ALIGN` marks a location that must be aligned to
+`N`-bytes, where `N` is the value of the addend field; this relocation is only
```
jrtc27, line 488: The addend is currently the number of bytes of NOPs emitted, which you can then use to infer what N is.

jrtc27 (commented):
So, what happens with things like:
```
.byte 0
.balign 16
```
Is that currently handled correctly? What does it even emit? We need to make sure all edge cases are covered before we commit to this being an official part of the ABI, and I suspect currently things break if you do something like that.

riscv-elf.adoc
jrtc27: (and the relocation is to the start of the NOP sequence, not the instruction you want to align)

kito-cheng: Fixed... I should check binutils implementation before I written :(

kito-cheng:
> So, what happens with things like:
> 
> ```
> .byte 0
> .balign 16
> ```
> 
> Is that currently handled correctly? What does it even emit? We need to make sure all edge cases are covered before we commit to this being an official part of the ABI, and I suspect currently things break if you do something like that.

Currently binutils will result linking error, but I think it should padding with a NULL byte at first and then 14 byte `NOPs`, I've add the behavior into spec.

kito-cheng:
@jim-wilson Could you review the updated version :)
@Nelson1225 Could you review that to make sure that is consist with binutils except the handling for 1-byte alignment (`.byte 0 + .balign 16`).

Nelson1225:
> > So, what happens with things like:
> > ```
> > .byte 0
> > .balign 16
> > ```
> > Is that currently handled correctly? What does it even emit? We need to make sure all edge cases are covered before we commit to this being an official part of the ABI, and I suspect currently things break if you do something like that.
>
> Currently binutils will result linking error, but I think it should padding with a NULL byte at first and then 14 byte `NOPs`, I've add the behavior into spec.

What I expected is that we probably need to make sure the boundary (from data to text) align to the text alignment.  And we now have mapping symbols in GNU binutils (may need llvm guys to make sure if it is also supported in llvm), so it shouldn't be hard to find the boundaries that need extra alignment padding.  For example,

.byte 0 -> data, `$d`
.balign16 -> fill nops in text section, so this is part of text, `$x`

If the rvc is enabled/disabled (text alignment is 2/4 byte), then we can do the 2/4-byte alignment after the `.byte 0' first, then the following alignments should work as expected.  And we probably need to do the same thing for the boundary from rvc to norvc, and probably also need to consider the .option relax/norelax.  Ideally, we can handle these boundaries as the above idea, but not sure if it can work for every cases.  Considering data/text, rvc/norvc and relax/norelax at the same time makes things complicated.  However, I have discussed this with Palmer and Kito many times, and Palmer also discussed this with Jim and Andrew before, maybe we should continue to recommend users not to put their data in text until we have a good solution.

Nelson1225:
>Currently binutils will result linking error, but I think it should padding with a NULL byte at first and then 14 byte NOPs, I've add the behavior into spec.

kito-cheng:
> Yes, the NULL byte padding is what I expected - align the boundary from data to text, if rvc is enable, then we need 1 byte padding, and 14 bytes NOPS. Otherwise we need 3-byte padding, and 12 bytes NOPS.

1 byte padding + 14 bytes NOPS or 3-byte padding + 12 bytes NOPS are not too much difference on no-RVC case, since both are not valid instruction sequence if CPU not support RVC, I think we could using some more general term like padding byte instead of NULL byte here to let implementation has flexibility.

Nelson1225:
LGTM, thanks.

kito-cheng:
Changes:
- Revise the description for the padding bytes.

jrtc27:
> > > So, what happens with things like:
> > > ```
> > > .byte 0
> > > .balign 16
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Is that currently handled correctly? What does it even emit? We need to make sure all edge cases are covered before we commit to this being an official part of the ABI, and I suspect currently things break if you do something like that.
> > 
> > 
> > Currently binutils will result linking error, but I think it should padding with a NULL byte at first and then 14 byte `NOPs`, I've add the behavior into spec.
> 
> What I expected is that we probably need to make sure the boundary (from data to text) align to the text alignment. And we now have mapping symbols in GNU binutils (may need llvm guys to make sure if it is also supported in llvm), so it shouldn't be hard to find the boundaries that need extra alignment padding. For example,

Uh, I hope you're potentially willing to change the implementation, because that pull request has _not_ been merged yet. I don't foresee it changing, but it is _not_ a finalised spec.

kito-cheng:
Changes:
- Remove description about padding byte for location with 1-byte align

It's not implement anywhere, let add back once we have PoC to make sure it work-able after first release.

riscv-elf.adoc
```diff
@@ -482,6 +482,30 @@ label:
...
+of the addend field, e.g. `R_RISCV_ALIGN` with addend value 2 means align to 4
+bytes; this relocation is only required if the containing section
+has any `R_RISCV_RELAX` relocations.
+
+The reason why we need a special relocation for handing alignment is because
+code relaxation might shrink the code during the linking process, that may
+cause an aligned point become mis-aligned after code relaxation.
+
+In order to let the linker have enough space to make sure the alignment
+requirement can be met after code relaxation, the assembler must insert a
```
jrtc27, line 498:
```suggestion
To ensure the linker can always satisfy the required alignment solely by deleting bytes, the assembler must emit a
```
(the key thing is that this means the linker only ever needs to _delete_ bytes, which allows various invariants to hold and make implementing relaxation much more feasible)
(and s/insert/emit/)

riscv-elf.adoc
```diff
@@ -482,6 +482,30 @@ label:
...
+`R_RISCV_ALIGN` relocation and then insert `N` - 4 padding bytes or `N` - 2
+padding bytes if the C extension is enabled, the padding bytes insert before
+the location of the alignment diretive like `.align`, `.p2align` or `.balign`,
```
jrtc27, line 501: Too assembly-specific (LLVM doesn't have alignment directives in any of its code by default because it never gets turned into assembly, for example)

riscv-elf.adoc
```diff
@@ -482,6 +482,30 @@ label:
...
+the addend value of that relocation is the number of padding bytes;
+linker may remove arbitrary numbers of the padding bytes at the linking process
+to meet the alignment requirement.
+
+The assembler and linker must ensure padding bytes are valid instructions like
+`nop` or `c.nop` if possible.
```
jrtc27, line 507: Stronger than that, they must have no side-effects, and don't you have to be careful about having a 4-byte NOP where you end up only deleting half of it?

riscv-elf.adoc
```diff
@@ -482,6 +482,30 @@ label:
...
+has any `R_RISCV_RELAX` relocations.
+
+The reason why we need a special relocation for handing alignment is because
+code relaxation might shrink the code during the linking process, that may
+cause an aligned point become mis-aligned after code relaxation.
+
+In order to let the linker have enough space to make sure the alignment
+requirement can be met after code relaxation, the assembler must insert a
+`R_RISCV_ALIGN` relocation and then insert `N` - 4 padding bytes or `N` - 2
+padding bytes if the C extension is enabled, the padding bytes insert before
```
jrtc27, line 500:
```suggestion
`R_RISCV_ALIGN` relocation and then insert `N - IALIGN` padding bytes before
```
though this is kind of the wrong order to define it, the alignment directive points at the code to align with the padding preceding it, so it would make sense for the spec here to talk about the padding bytes and then the relocation in that order

riscv-elf.adoc
```diff
@@ -482,6 +482,30 @@ label:
...
+bytes; this relocation is only required if the containing section
+has any `R_RISCV_RELAX` relocations.
+
+The reason why we need a special relocation for handing alignment is because
+code relaxation might shrink the code during the linking process, that may
+cause an aligned point become mis-aligned after code relaxation.
+
+In order to let the linker have enough space to make sure the alignment
+requirement can be met after code relaxation, the assembler must insert a
+`R_RISCV_ALIGN` relocation and then insert `N` - 4 padding bytes or `N` - 2
```
jrtc27, line 499: I don't see it defined anywhere that the addend is how many bytes of padding have been inserted?

riscv-elf.adoc
```diff
@@ -482,6 +482,30 @@ label:
...
+padding bytes if the C extension is enabled, the padding bytes insert before
+the location of the alignment diretive like `.align`, `.p2align` or `.balign`,
+the addend value of that relocation is the number of padding bytes;
+linker may remove arbitrary numbers of the padding bytes at the linking process
```
jrtc27, line 503: Arbitrary is not true, cannot remove more than the amount of padding there is otherwise you break anything. Otherwise per this spec a legal implementation of R_RISCV_ALIGN would be to delete everything before it so it starts at 0.

riscv-elf.adoc
kito-cheng: Use compiler and/or assembler term here, and add sentence `or emit by compiler directly`.

riscv-elf.adoc
kito-cheng: Good point, added more description to say linker must ensure those instructions are no side-effect and aligned to `IALIGN`.

kito-cheng:
ping

riscv-elf.adoc
```diff
@@ -482,6 +482,40 @@ label:
...
+`.balign` or emit by compiler directly, the addend value of that relocation
+is the number of padding bytes.
+
+The compiler and assembler must ensure padding bytes are valid instructions
+without any side-effect like `nop` or `c.nop`, and make sure those instructions
+are aligned to IALIGN if possible.
+
+The linker may remove part of the padding bytes at the linking process to meet
+the alignment requirement, and make sure those padding bytes still are valid
+instructions and each instruction is aligned to at least IALIGN byte.
```
rui314, line 506: In the first sentence, it is not required to always use `nop` or `c.nop`, so the compiler or the assembler can emit arbitrary instructions as long as they have no side effects.

The second sentence says that it is linker's responsibility to keep instructions legal after removing bytes. I think its hard to satisfy, as the linker is usually agnostic about machine instructions. If we really ensure that, we need to teach the linker about the all possible valid nop instructions.

So, I think we should move the responsibility from the linker to the compiler/assembler. The linker should always remove bytes from the end of padding, and the compiler/assembler has to guarantee that the instructions are valid after removing bytes. This is actually what we are doing right now; we should just clarify that.

MaskRay: The number of NOP bytes modulo 4 may be 0 or 2. R_RISCV_ALIGN relaxation may change the state from 0 to 2, or from 2 to 0. So the linker has to rewrite the code sequence, regardless of what compiler/assembler does (since we are not going to do #183).

rui314: If the linker has to rewrite NOPs to fill the padding, what was the point of writing NOPs in object files? We could have wrote no NOPs to object files and instead just say "align this instruction at 2^n" by R_RISC_ALIGN, and the linker would align that location to a specified boundary and write NOPs to the padding caused by the alignment specification. It looks like we didn't have to emit bunch of NOPs after R_RISCV_ALIGN.

rui314: On second thought, is that actually a problem? I mean R_RISCV_ALIGN with 2^1 alignment requirement doesn't make much sense, as it is always satisfied. So we always remove a number of bytes multiple of 4.

kito-cheng: > So we always remove a number of bytes multiple of 4.

We might remove byte with non-multiple-of-4 in some case:

Consider following case:
```
foo:
        call    bar
        c.nop
.balign 8
        add t1, t2, t3
```

`$ riscv64-unknown-elf-as -c x.s  -o x.o -mrelax -march=rv32gc`
```
00000000 <foo>:
   0:   00000097                auipc   ra,0x0
                        0: R_RISCV_CALL bar
                        0: R_RISCV_RELAX        *ABS*
   4:   000080e7                jalr    ra # 0 <foo>
   8:   0001                    nop
   a:   0001                    nop
                        a: R_RISCV_ALIGN        *ABS*+0x6
   c:   00000013                nop
  10:   01c38333                add     t1,t2,t3
```

And that might result 3 different results:

- auipc + jalr can't relax: no padding removed.

```
00000000 <foo>:
   0:   00000097                auipc   ra,0x0
   4:   000080e7                jalr    ra # 0 <foo>
   8:   0001                    c.nop
   a:   0001                    c.nop
   c:   00000013                nop
  10:   01c38333                add     t1,t2,t3
```

- auipc + jalr relaxed to jal: 4 byte padding removed.
```
00000000 <foo>:
   0:   000080e7                jal     <foo>
   4:   0001                    c.nop
   6:   0001                    c.nop # Delete first 4 byte and rewrite last 2 byte to c.nop
   8:   01c38333                add     t1,t2,t3
```

- auipc + jalr relaxed to c.jal: 2 byte padding removed.

```
00000000 <foo>:
   0:   0000                c.jal    ra # 0 <foo>
   2:   0001                    nop
   4:   00000013                nop # Delete 2 byte
  8:   01c38333                add     t1,t2,t3
```

rui314: Ah, good point. Then maybe we have no choice other than making it a responsibility of the static linker to fill gaps with appropriate number of c.nop and nop operations.

kito-cheng:
Changes:
- Rebase

kito-cheng:
@rui314 @MaskRay I would like to gather LGTM from you guys, since we already got two LGTM from GNU developer but no LGTM from non-GNU developer  :)

rui314:
> The relocation type `R_RISCV_ALIGN` marks a location that must be aligned to `N`-bytes

To me, this sounded like `R_RISCV_ALIGN` appears past-the-end of the padding nop instructions, as that's the location that will be aligned.

riscv-elf.adoc
```diff
@@ -675,6 +675,40 @@ label:
     sw t2, t0, %pcrel_lo(label)   # R_RISCV_PCREL_LO12_S (label)
 ----
 
+==== Relocation for Alignment
+
+The relocation type `R_RISCV_ALIGN` marks a location that must be aligned to
+`N`-bytes, where `N` is the smallest power of two that greater than the value
```
luismarques, line 681:
```suggestion
`N`-bytes, where `N` is the smallest power of two that is greater than the value
```

riscv-elf.adoc
```diff
@@ -675,6 +675,40 @@ label:
...
+of the addend field, e.g. `R_RISCV_ALIGN` with addend value 2 means align to 4
+bytes; this relocation is only required if the containing section has any
```
luismarques, line 683: "the smallest power of two that is greater than the value of the addend field" -> `2^2 = 4` is equal, not greater, so this becomes a bit confusing to read. Also, it's not immediately obvious what is the relationship between `N` and `addend`. Regardless of those specific issues, having two or more examples would help disambiguate between possible interpretations that the reader might have.

riscv-elf.adoc
luismarques: > If the linker has to rewrite NOPs to fill the padding, what was the point of writing NOPs in object files?

Good question. Even if the relaxation design doesn't really require it, I think it's still a good idea to do that because that's a lot clearer when you're looking at the unlinked object files. I can't think of a scenario where the compiler would benefit from having the flexibility to choose other instructions but maybe you can think of something?

riscv-elf.adoc
```diff
@@ -675,6 +675,40 @@ label:
...
+NOTE: `R_RISCV_ALIGN` relocation is needed because code relaxation can shrink
+preceding code during the linking process, which may cause an aligned location
+to become mis-aligned.
+
+NOTE: IALIGN means the instruction-address alignment constraint. IALIGN is 4
+bytes in the base ISA, but some ISA extensions, including the compressed ISA
+extension, relax IALIGN to 2 bytes. IALIGN may not take on any value other than
+4 or 2, this term is also defined in `The RISC-V Instruction Set Manual` in
+similar meaning, the only difference is the psABI defined that in byte and
+the ISA manual defined that in bits.
```
luismarques, line 710: Can you link to this note when `IALIGN` is first mentioned?

luismarques, line 710:
```suggestion
4 or 2. This term is also defined in `The RISC-V Instruction Set Manual` with a
similar meaning, the only difference being it is specified in terms of the number
of bits instead of the number of bytes.
```

riscv-elf.adoc
kito-cheng: > If the linker has to rewrite NOPs to fill the padding, what was the point of writing NOPs in object files? We could have wrote no NOPs to object files and instead just say "align this instruction at 2^n" by R_RISC_ALIGN, and the linker would align that location to a specified boundary and write NOPs to the padding caused by the alignment specification. It looks like we didn't have to emit bunch of NOPs after R_RISCV_ALIGN.

We are trying do only code shrink, and do not insert new code during the linker relaxation process, that could prevent we made any branch/jump become out-of-range, RISC-V ISA has relative short branch and jump range* compare to other architecture, so once jump is out-of-range we need using indirect-jump to resolve, but it's almost impossible at the linker stage.

* +-4k for branch, and +-1M for jump.

kito-cheng:
Changes:
- Add example to show how R_RISCV_ALIGN is used.
- Add pseudocode to show how to compute alignment of R_RISCV_ALIGN.
- Add refer to IALIGN when it is first mentioned.
- Add more word to clarify the location of R_RISCV_ALIGN is point to beginning of padding bytes.

kito-cheng:
Ping :)

rui314:
What about my previous comment?

kito-cheng:
>> To me, this sounded like R_RISCV_ALIGN appears past-the-end of the padding nop instructions, as that's the location that will be aligned. Maybe we should make it more clear that the relocation refers the beginning of the nops?
> 
> What about my previous comment?

Does this clear or its still not clear enough? https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/210/commits/cb8a12e99efb618f1ab82dfe1cca2e6b6db14621
```diff
 The relocation type `R_RISCV_ALIGN` marks a location that must be aligned to
 `N`-bytes, where `N` is the smallest power of two that is greater than the value
 of the addend field, e.g. `R_RISCV_ALIGN` with addend value 2 means align to 4
-bytes; this relocation is only required if the containing section has any
-`R_RISCV_RELAX` relocations.
+bytes, `R_RISCV_ALIGN` with addend value 4 means align to 8 bytes; this
+relocation is only required if the containing section has any `R_RISCV_RELAX`
+relocations, `R_RISCV_ALIGN` points to the beginning of the padding bytes,
+and the instruction that actually needs to be aligned is located at the point
+of `R_RISCV_ALIGN` plus its addend.
```

rui314:
LGTM

nit: there's a typo: diretive → directive

Nelson1225:
> > > So, what happens with things like:
> > > ```
> > > .byte 0
> > > .balign 16
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Is that currently handled correctly? What does it even emit? We need to make sure all edge cases are covered before we commit to this being an official part of the ABI, and I suspect currently things break if you do something like that.
> > 
> > 
> > Currently binutils will result linking error, but I think it should padding with a NULL byte at first and then 14 byte `NOPs`, I've add the behavior into spec.
> 
> What I expected is that we probably need to make sure the boundary (from data to text) align to the text alignment. And we now have mapping symbols in GNU binutils (may need llvm guys to make sure if it is also supported in llvm), so it shouldn't be hard to find the boundaries that need extra alignment padding. For example,
> 
> .byte 0 -> data, `$d` .balign16 -> fill nops in text section, so this is part of text, `$x`
> 
> If the rvc is enabled/disabled (text alignment is 2/4 byte), then we can do the 2/4-byte alignment after the `.byte 0' first, then the following alignments should work as expected. And we probably need to do the same thing for the boundary from rvc to norvc, and probably also need to consider the .option relax/norelax. Ideally, we can handle these boundaries as the above idea, but not sure if it can work for every cases. Considering data/text, rvc/norvc and relax/norelax at the same time makes things complicated. However, I have discussed this with Palmer and Kito many times, and Palmer also discussed this with Jim and Andrew before, maybe we should continue to recommend users not to put their data in text until we have a good solution.

Just mentioned that there is a patch from Jan Beulich, who is one of the binutils global maintainers, seems trying to resolve the above rvc to norvc alignment issue, [here](https://inbox.sourceware.org/binutils/e00cbdad-ba25-4ead-86f2-dd0fa5b01ac6@suse.com/).

PR #214, opened 2021-09-24
Fix incorrect description of call pseudoinstruction expansion

### LINKED ISSUE #206: Fix incorrect description of call pseudoinstruction expansion
The following is currently specified:

> `R_RISCV_CALL` and `R_RISCV_CALL_PLT` relocations are associated with
> pairs of instructions (`AUIPC+JALR`) generated by the `CALL` or `TAIL`
> pseudoinstructions.  Originally, these relocations had slightly different
> behavior, but that has turned out to be unnecessary, and they are now
> interchangeable.
> 
> In position dependent code (`-fno-pic`), with relaxation enabled,
> the `AUIPC` instruction in the
> `AUIPC+JALR` pair has both a `R_RISCV_CALL` relocation and a `R_RISCV_RELAX`
> relocation indicating the instruction sequence can be relaxed during linking.
> 
> In position independent code (`-fPIC`, `-fpic` or `-fpie`),
> with relaxation enabled, the `AUIPC`
> instruction in the `AUIPC+JALR` pair has both a `R_RISCV_CALL_PLT` relocation
> and a `R_RISCV_RELAX` relocation indicating the instruction sequence can be
> relaxed during linking.
> 
> Procedure call linker relaxation allows the `AUIPC+JALR` pair to be relaxed
> to the `JAL` instruction when the procedure or PLT entry is within (-1MiB to
> +1MiB-2) of the instruction pair.
> 
> The pseudoinstruction:
> 
> ```
>     call symbol
> ```
> 
> expands to the following assembly and relocation:
> 
> ```
>     auipc ra, 0           # R_RISCV_CALL (symbol), R_RISCV_RELAX (symbol)
>     jalr  ra, ra, 0
> ```
> 
> and when `-fpic` is enabled it expands to:
> 
> ```
>     auipc ra, 0           # R_RISCV_CALL_PLT (symbol), R_RISCV_RELAX (symbol)
>     jalr  ra, ra, 0
> ```

This is incorrect for both LLVM and GNU as. Which relocation is picked depends on whether or not a `@plt` suffix is present on the symbol. Whether or not _that suffix is emitted by the compiler_ is governed by `-fpic` etc. We should also untangle the R_RISCV_RELAX from here; we don't specify that for other code sequences in the spec, it's the same for both PLT and non-PLT relocations so should not be duplicated, and we don't actually currently specify what happens if relaxations are _disabled_. Also, this is likely too assembly-focused and duplicates riscv-asm-manual material.

kito-cheng:
- Fix #206

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -396,16 +396,9 @@ pseudoinstructions.  Originally, these relocations had slightly different
 behavior, but that has turned out to be unnecessary, and they are now
 interchangeable.
 
-In position dependent code (`-fno-pic`), with relaxation enabled,
-the `AUIPC` instruction in the
-`AUIPC+JALR` pair has both a `R_RISCV_CALL` relocation and a `R_RISCV_RELAX`
+With relaxation enabled, the `AUIPC` instruction in the `AUIPC+JALR` pair has
+both a `R_RISCV_CALL` or `R_RISCV_CALL_PLT` relocation and an `R_RISCV_RELAX`
 relocation indicating the instruction sequence can be relaxed during linking.
-
-In position independent code (`-fPIC`, `-fpic` or `-fpie`),
-with relaxation enabled, the `AUIPC`
-instruction in the `AUIPC+JALR` pair has both a `R_RISCV_CALL_PLT` relocation
-and a `R_RISCV_RELAX` relocation indicating the instruction sequence can be
-relaxed during linking.
 
 Procedure call linker relaxation allows the `AUIPC+JALR` pair to be relaxed
 to the `JAL` instruction when the procedure or PLT entry is within (-1MiB to
@@ -416,6 +409,7 @@ The pseudoinstruction:
 [,asm]
 ----
     call symbol
+    call symbol@plt
 ----
 
 expands to the following assembly and relocation:
@@ -426,7 +420,7 @@ expands to the following assembly and relocation:
     jalr  ra, ra, 0
 ----
 
-and when `-fpic` is enabled it expands to:
+and when symbol has an `@plt` suffix it expands to:
 
 [,asm]
 ----


riscv-elf.adoc
```diff
@@ -384,17 +384,10 @@ pseudoinstructions.  Originally, these relocations had slightly different
 behavior, but that has turned out to be unnecessary, and they are now
 interchangeable.
 
-In position dependent code (`-fno-pic`), with relaxation enabled,
-the `AUIPC` instruction in the
-`AUIPC+JALR` pair has both a `R_RISCV_CALL` relocation and a `R_RISCV_RELAX`
+With relaxation enabled, the `AUIPC` instruction in the `AUIPC+JALR` pair has
+both a `R_RISCV_CALL` or `R_RISCV_CALL_PLT` relocation and a `R_RISCV_RELAX`
```
jrtc27, line 388:
```suggestion
both a `R_RISCV_CALL` or `R_RISCV_CALL_PLT` relocation and an `R_RISCV_RELAX`
```

riscv-elf.adoc
```diff
@@ -414,7 +408,7 @@ expands to the following assembly and relocation:
     jalr  ra, ra, 0
 ----
 
-and when `-fpic` is enabled it expands to:
+and when symbol with `@PLT` suffix it expands to:
```
jrtc27, line 411:
```suggestion
and when symbol has an `@plt` suffix it expands to:
```
(yes it is actually case-sensitive in both assemblers)

riscv-elf.adoc
```diff
@@ -404,6 +397,7 @@ The pseudoinstruction:
 [,asm]
 ----
     call symbol
+    call symbol@PLT
```
jrtc27, line 400:
```suggestion
    call symbol@plt
```

kito-cheng:
Changes:
- Apply @jrtc27's suggestion.

jrtc27 (approved):
Though please squash the code review commit into the main one before merging

kito-cheng:
Squashed :)

PR #215, opened 2021-09-24
Document we must export  to dynamic symbol table

kito-cheng:


kito-cheng:
@Nelson1225 could you review that?

riscv-elf.adoc
```diff
@@ -134,7 +134,8 @@ There are no RISC-V specific definitions relating to ELF string tables.
 
 === Symbol Table
 
-There are no RISC-V specific definitions relating to ELF symbol tables.
+`__global_pointer$` must export to dynamic symbol table for dynamically-linked
```
Nelson1225, line 137: Maybe use "position dependency executable (PDE)" to replace the "dynamically-linked executable"?, since we won't output gp when generating PIE.  Besides, should we mention the details that this is used to resolve the ifunc problem?

The ifunc problem is that - when the ifunc resolver is in the executable, we may relax the variables to gp-relative access instruction in the ifunc resolver, or in other functions that called by the ifunc resolver.  But this will cause the uninitialized gp problem since the ifunc need to be resolved at the early runtime, that is at the pre-load stage, but we set the gp until the startup code.  Therefore, we make sure `__global_pointer$' is output as a dynamic symbol when we are generating pde (we only do the relaxation for pde), so ld.so can init the gp register early, before the pre-load stage.

kito-cheng: Thanks, updated :)

kito-cheng:
Changes:
- Using position dependency executable rather than dynamically-linked executable.

kito-cheng:
@MaskRay

riscv-elf.adoc
jrtc27: We should use gp for PIEs. The fact we don't is a binutils deficiency. Only DSOs cannot use gp.

kito-cheng: @jrtc27 what do you think if we keep using "dynamically-linked executable" ?

jrtc27: That's what I would use

kito-cheng:
Changes:
- Revert last change, use `dynamically-linked executable`.

kito-cheng:
Changes:
- Rebase.

riscv-elf.adoc
```diff
@@ -165,6 +165,9 @@ table.
 +
 See <<Dynamic Linking>> for the meaning of `STO_RISCV_VARIANT_CC`.
 
+`__global_pointer$` must export to dynamic symbol table for dynamically-linked
+executable if there is any reference to that symbol.
```
jrtc27, line 169:
```suggestion
`__global_pointer$` must be exported in the dynamic symbol table of dynamically-linked
executables if there are any GP-relative accesses present in the executable.
```
or something like that? Not sure what the exact requirements are, but the uses are implicit via gp.

kito-cheng: Yeah, but I think it give some room to optimize that then restrict export that, like @Nelson1225's suggestion, his is PIE does not use any GP-relative accesses so we don't need to export that.

jim-wilson (approved):

kito-cheng:
@jrtc27 let me know if it's good to you now :)

jrtc27 (approved):

Issue #132, opened 2020-02-21
do we support DT_INIT?


jim-wilson:
Our policy is that we don't support .init or .fini sections.  However, the GNU toolchain is still creating binaries and shared libraries with DT_INIT set which appears to conflict with that.  That leads to several questions.  Should we have DT_INIT support?  Given that we already have linux distros built with DT_INIT support, is it possible to drop support for it, or would that be an unacceptable ABI break?  There is also a similar issue with DT_FINI.  Though DT_INIT seems to be the more immediate problem as a glibc libc.so has DT_INIT but not DT_FINI.

This came up as glibc is adding better support for systems without ini/fini sections, and those patches unexpectedly broke RISC-V support.  Leading to questions about whether RISC-V really dropped support for init/fini sections or not.

A search of Fedora RISC-V libraries shows about half a dozen projects have libraries with DT_INIT/DT_FINi.  One of those is GNU libc which is already under discussion.  One is numactl.  I looked at that one, and I see that the Makefile.am has -Wl,-init,numa_init -Wl,-fini,numa_fini so this is at least not the toolchain's fault.  But then this does bring up another question as to what the toolchain should do in this case.  Give an error when -Wl,-init is used?  Or try to convert a linker -init option into an init_array entry?  I'd rather give a linker error and fix the package.

jim-wilson:
Here is a pointer to the glibc thread
https://sourceware.org/ml/libc-alpha/2020-02/msg00896.html

aswaterman:
“Fix the package” sounds cleanest, but I don’t have a clear idea of how pervasive those changes would be. Given the relatively small number of packages with DT_INIT (assuming I understood the post on libc-alpha correctly), perhaps the scope is narrow enough.

fweimer:
> I looked at that one, and I see that the Makefile.am has -Wl,-init,numa_init -Wl,-fini,numa_fini so this is at least not the toolchain's fault.

[...]

PR #217, opened 2021-09-27
riscv-elf: Document that DT_INIT/DT_FINI should not be relied upon

jrtc27:
These are horrible, legacy features that have had better alternatives for many years. Binutils and glibc do not support them on RISC-V and never will.

Closes: #132


MaskRay:
LG

jim-wilson (approved):

PR #218, opened 2021-09-27
Add vector register convention

kito-cheng:
We didn't have enough engineer resource to explore the complete vector register convention yet, so we only add minimal description for vector register convention, and we intend to extend by adding new calling convention variant in future.


cmuellner:
Note, that this was discussed in the chairs meeting on Sept 22, 2021, and this is a proposal, which would allow us to explicitly handle the vector extension, but don't commit to a vector calling convention yet. Of course, this allows that we can define a vector calling convention for passing vector arguments in the future.

The definitions within this proposal are actually Appendix B of the Vector specification (which is currently in public review).

riscv-cc.adoc
```diff
@@ -52,6 +52,22 @@ The Floating-Point Control and Status Register (fcsr) must have thread storage
...
+| vl      |              | Vector length                | No
+| vtype   |              | Vector data type register    | No
+|===
+
+Vector registers are not used for passing argument or return values, we
+intend to define new calling convention variant to allow that for optimizing
+software in future.
+
+The `vxrm` and `vxsat` fields of `vcsr` have thread storage duration.
+
```
cmuellner, line 70: I am missing two aspects from the vector specification:

1) Handling of `vstart`
2) Behaviour of context switches

```
Procedures may assume that vstart is zero upon entry. Procedures may assume that vstart is zero upon return from a
procedure call.

Note that application software should normally not write vstart explicitly. Any procedure that does explicitly write vstart to a nonzero value must zero vstart before either returning or calling another procedure.
```

I would simply add this as well.

2) Behaviour during context switches

```
Executing a system call causes all caller-saved vector registers (v0-v31, vl, vtype) and vstart to become unspecified.
Note that this scheme allows system calls that cause context switches to avoid saving and later restoring the vector registers.
Note that most OSes will choose to either leave these registers intact or reset them to their initial state to avoid leaking information across process boundaries.
```

Although this is not strictly needed for a calling convention, it is needed to avoid performance penalties (for simple syscalls).
However, it does not mean that all context switches can ignore vector registers. E.g. if the scheduler moves a running thread (that uses vector registers) off the CPU and puts it into the runable queue, because another thread is getting CPU time and that other thread is also using vector registers.

We could add this as an informal note:
`Note, that the registers named above may also become unspecified in case of environment calls (SYSTEM instruction, e.g. system calls)`.

jrtc27: Non-voluntary context switches clearly must preserve state for everything otherwise things break.

Voluntary context switches are OS-specific and do not belong in the psABI, that is up to that kernel's system call ABI to define.

jrtc27: (With the exception of `syscall(2)`, which is inherently extremely OS-specific and thus totally out of scope, all other system calls are performed via function calls anyway)

cmuellner: Yes, therefore I suggested an informal note.
Not to address what is already addressed or does not need to be addressed, but to save developers from surprising behavior (for them).

fweimer-rh: @jrtc27 For lazy binding, the dynamic loader effectively has to perform a context switch, so that the incoming argument list is not clobbered by the caller. In glibc, some targets actually perform a fairly complete context switch, other targets have ELF markers that disable lazy binding automatically for variant procedure call standards. I don't know what is best for RISC-V here, but some decision has to be made (even inaction implies one in this case).

jrtc27: Several points:
1. This is a minimal "calling" convention that does not say how vector arguments are passed and how vectors are returned (#171 does that, but there is debate over the exact argument split); really it's just enough so that code can start to use vector registers within functions
2. #190 addresses that, replicating the AArch64 approach, so that when the time comes to pass vectors between functions in registers we can rely on runtimes supporting it already

jim-wilson:
Looks OK to me.

The vstart text looks like a useful addition.

The syscall/context switch comments look confused.  Making vector regs undefined across a syscall does not make context switches faster.  These are different things.  There may be a point here though.  The kernel is compiled soft-float so that it doesn't have to save/restore FP registers when a syscall happens.  Maybe they want to allow the kernel to use vector registers though?  In which case the syscall handler either needs to save/restore them or clobber them, and clobber is faster.  I'm not sure if this makes any sense though.  I think the kernel guys need to talk to the vector guys and work something out.  Meanwhile, I think we should stay out of this.  Syscall handling is out of scope for this document.  Though if syscalls do clobber vector registers, then the compiler will likely need to be aware of that and do something sensible like save/restore them itself (e.g. caller save) if necessary.

cmuellner:
> The syscall/context switch comments look confused.

Then let's skip this in psABI and bring the discussion to the vector spec authors so they can either clarify or drop that as well.

jim-wilson:
Asking around, I was pointed to this
    https://www.kernel.org/doc/Documentation/arm64/sve.txt
which mentions that vector regs can be clobbered by syscalls.  I find that surprising, but it means that the current text in the V spec is consistent with the SVE linux kernel docs.

kito-cheng:
I am in favor on do not document anything about content switch, we didn't document that for FPR too, that's OS specific part.

kito-cheng:
Changes:
- Add description of vstart

jim-wilson (approved):

riscv-cc.adoc
```diff
@@ -52,6 +52,22 @@ The Floating-Point Control and Status Register (fcsr) must have thread storage
...
+|===
+| Name    | ABI Mnemonic | Meaning                      | Preserved across calls?
+| v0-v31  |              | Temporary registers          | No
+| vl      |              | Vector length                | No
+| vtype   |              | Vector data type register    | No
+|===
+
+Vector registers are not used for passing argument or return values, we
+intend to define new calling convention variant to allow that for optimizing
+software in future.
```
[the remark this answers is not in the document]
jrtc27: And, to nit-pick, I'd change the comma to a semicolon

kito-cheng:
Changes:
- Apply Jim's and Jessica's comment.
- Rebase.

riscv-cc.adoc
```diff
@@ -52,6 +52,29 @@ The Floating-Point Control and Status Register (fcsr) must have thread storage
...
+|===
+| Name    | ABI Mnemonic | Meaning                      | Preserved across calls?
+| v0-v31  |              | Temporary registers          | No
+| vl      |              | Vector length                | No
+| vtype   |              | Vector data type register    | No
+|===
+
+Vector registers are not used for passing argument or return values; we
+intend to define new calling convention variant to allow that for optimizing
+software in future.
```
jrtc27, line 67: You've lost Jim's suggestions and only kept the semicolon

kito-cheng: Ooops...add back

kito-cheng:
Changes:
- Correctly apply Jim's and Jessica's comment.
- Adjust format

kito-cheng:
Add description for system call in another PR: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/222

PR #219, opened 2021-09-27
Adjust README to reflect new structure

kito-cheng:


jim-wilson:
You switched the order of ELF and DWARF.  I prefer the old order.  I would suggest moving the dwarf line to the bottom, since it does make sense to keep the elf subsections together.

"ELF Object files" is actually an == subtitle inside the riscv-elf.adoc file, so using it for the entire file looks wrong.  The file itself says "ELF Specification" and similarly the riscv-dwarf.adoc file says "DWARF Specification".  I suggest using those names for the entire file.  riscv-cc.adoc is just "Calling Conventions" without the Procedure.

Inside the riscv-elf.adoc file, we have 5 "==" subtitles.  We might as well list them all if we are going to list any.  Relocations is not one of them.
rohan:2258$ grep "^== " riscv-elf.adoc 
== Code models
== Dynamic Linking
== C++ Name Mangling
== ELF Object Files
== Code relaxation

Relocations is a === subtitle.inside the == ELF Object Files section.  I think that there are too many === substitles to list them all.  I don't think we really need this one.

If we are listing the == subtitles in the riscv-elf.adoc file, do we also want to do the same for the other files?  The dwarf file has only one, "Dwarf Register Numbers".  The riscv-cc.adoc file has four.  Keep in mind the more links we put in the README.md file, the more stuff we have to worry about later when we change the docs.  If we aren't listing the subsections in the other files, do we still want to do that for the ELF file?

jrtc27:
I agree, we should only link to the documents themselves. We can still sign-post people if we want to by giving examples of important parts of the documents, but those shouldn't be linked, nor tied to specific titles, just informal names for the parts so that they don't need updating as the document evolves. Though I'm not sure that's really needed, there are only three documents to search through and the names should make it obvious what's what.

kito-cheng:
My intention is to having some link to the actually  document, but manually sync that sound make no sense, so let close this PR for now, maybe we could provide link to generated html version document, but I didn't figure out how to do that in right way with github page, let me investigate little bit more about github page and github action to make after all release stuff settle down :p

PR #220, opened 2021-10-01
Update author list

kito-cheng:
List is generated by `git log --format='%an <%ae>' |sort |uniq` with few manually post process.

### DIFF
--- a/preamble.adoc
+++ b/preamble.adoc
@@ -1,15 +1,29 @@
-This RISC-V ELF psABI specification document is
+Contributors to all versions of the spec in alphabetical order:
 
-(C) 2016 Palmer Dabbelt <palmer@dabbelt.com> +
-(C) 2016 Stefan O'Rear <sorear2@gmail.com> +
-(C) 2016 Kito Cheng <kito.cheng@gmail.com> +
-(C) 2016-2017 Andrew Waterman <aswaterman@gmail.com> +
-(C) 2016-2017 Michael Clark <michaeljclark@mac.com> +
-(C) 2017-2019 Alex Bradbury <asb@asbradbury.org> +
-(C) 2017 David Horner <ds2horner@gmail.com> +
-(C) 2017 Max Nordlund <max.nordlund@gmail.com> +
-(C) 2017 Karsten Merker <merker@debian.org> +
-(C) 2019 Sam Elliott <selliott@lowrisc.org>
+Alex Bradbury,
+Andrew Burgess,
+Chih-Mao Chen,
+Zakk Chen,
+Kito Cheng,
+Nelson Chu,
+Michael Clark,
+Jessica Clarke,
+Palmer Dabbelt,
+Sam Elliott,
+Gonzalo Brito Gadeschi,
+Sebastian Huber,
+Roger Ferrer Ibanez,
+Quey-Liang Kao,
+Nick Knight,
+Luís Marques,
+Evandro Menezes,
+Max Nordlund,
+Stefan O'Rear,
+Konrad Schwarz,
+Fangrui Song,
+Hsiangkai Wang,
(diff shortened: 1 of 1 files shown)


jim-wilson:
Some of the names are fake.  Rincewind for instance is a character in Terry Pratchett novels.  At least with the email addresses we can vaguely tell who they are, but one has no email address.

Do we want to make a distinction between people who made non-trivial contributions, and people who made trivial contributions like fixing a simple typo?  That would probably weed out all of the fake names if we didn't credit people for just fixing a typo.

jrtc27 (commented):
> Some of the names are fake. Rincewind for instance is a character in Terry Pratchett novels. At least with the email addresses we can vaguely tell who they are, but one has no email address.
> 
> Do we want to make a distinction between people who made non-trivial contributions, and people who made trivial contributions like fixing a simple typo? That would probably weed out all of the fake names if we didn't credit people for just fixing a typo.

Yeah, I don't want to have to list every single contributor over the lifetime of the document. "Substantive" contributions get you on the list, which is always going to be fuzzy, but some of these contributors definitely don't meet that bar.

preamble.adoc
```diff
@@ -1,15 +1,33 @@
...
-(C) 2016 Kito Cheng <kito.cheng@gmail.com> +
-(C) 2016-2017 Andrew Waterman <aswaterman@gmail.com> +
-(C) 2016-2017 Michael Clark <michaeljclark@mac.com> +
-(C) 2017-2019 Alex Bradbury <asb@asbradbury.org> +
-(C) 2017 David Horner <ds2horner@gmail.com> +
-(C) 2017 Max Nordlund <max.nordlund@gmail.com> +
-(C) 2017 Karsten Merker <merker@debian.org> +
-(C) 2019 Sam Elliott <selliott@lowrisc.org>
+Alex Bradbury <asb@lowrisc.org> +
+Andrew <apb@armyofcats.co.uk> +
```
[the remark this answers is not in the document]
kito-cheng: Update with his Embecosm email

kito-cheng:
Changes:
- Update the email and name list per @jrtc27's comment
- Remove people who made trivial contributions from the list

kito-cheng:
Ping

kito-cheng:
In case, we want to find the email address to contact, we still can found that from git log.

Changes:
- Only keep name in the author list.

preamble.adoc
```diff
@@ -1,15 +1,11 @@
...
-(C) 2016 Stefan O'Rear <sorear2@gmail.com> +
-(C) 2016 Kito Cheng <kito.cheng@gmail.com> +
-(C) 2016-2017 Andrew Waterman <aswaterman@gmail.com> +
-(C) 2016-2017 Michael Clark <michaeljclark@mac.com> +
-(C) 2017-2019 Alex Bradbury <asb@asbradbury.org> +
-(C) 2017 David Horner <ds2horner@gmail.com> +
-(C) 2017 Max Nordlund <max.nordlund@gmail.com> +
-(C) 2017 Karsten Merker <merker@debian.org> +
-(C) 2019 Sam Elliott <selliott@lowrisc.org>
+Alex Bradbury, Andrew Burgess, Andrew Waterman, Chih-Mao Chen,
```
jrtc27, line 3: Missed this in earlier versions but this should be sorted based on last name not first name

kito-cheng:
Changes:
- Sort by last name
- One name per name

Preview PDF: [riscv-abi.pdf](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/files/7324058/riscv-abi.pdf)

preamble.adoc
kito-cheng: Updated :)

PR #222, opened 2021-10-06
Add calling convention for system calls

kito-cheng:
Per discussion in last psABI, I added description to system call, I tried to find system call document for linux, but I can only find that on a [header file](https://github.com/torvalds/linux/blob/master/arch/riscv/include/asm/syscall.h)...

Actually we didn't specify anything about that, just a section to describe that's scope of respective OSs, and should find that from
the OS document.

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -320,6 +320,13 @@ NOTE: Although RV64GQ systems can technically use <<abi-lp64q,LP64Q>>, it is
 strongly recommended to use LP64D on general-purpose RV64GQ systems for
 compatibility with standard RV64G software.
 
+== Calling Convention for System Calls
+
+The calling convention for system calls does not fall within the scope of this
+document.
+Please refer to the documentation of the RISC-V execution environment interface
+(e.g OS kernel ABI, SBI).
+
 == C/C++ type details
 
 === C/C++ type sizes and alignments


kito-cheng:
@cmuellner

riscv-cc.adoc
```diff
@@ -295,6 +295,11 @@ NOTE: Although RV64GQ systems can technically use <<abi-lp64q,LP64Q>>, it is
 strongly recommended to use LP64D on general-purpose RV64GQ systems for
 compatibility with standard RV64G software.
 
+== Calling Convention for System Call
+
+Calling convention for system call is not scope of psABI, please reference the
```
cmuellner, line 300: The calling convention for system calls does not fall within the scope of this document.
Please refer to the documentation of the RISC-V execution environment interface
(e.g OS kernel ABI, SBI).

cmuellner:
In the call, we were talking about an informal note that suggests reading the OS documentation in the context of vector instructions (it is likely that the vector register contents won't survive a syscall, but a programmer might not be aware of this).

Putting this hint into its own section is fine for me as well.

I've rephrased your text a bit to use RISC-V terminology ("RISC-V execution environment interface") as defined in the ISA spec.

kito-cheng:
I didn't find a proper place to put the note since I think that's also applied for GPR and FPR, so...I decide put in a new section.

riscv-cc.adoc
kito-cheng: Updated :)

kito-cheng:
Changes:
- Apply  @cmuellner's suggestions from code review

riscv-cc.adoc
```diff
@@ -295,6 +295,13 @@ NOTE: Although RV64GQ systems can technically use <<abi-lp64q,LP64Q>>, it is
 strongly recommended to use LP64D on general-purpose RV64GQ systems for
 compatibility with standard RV64G software.
 
+== Calling Convention for System Call
```
jrtc27, line 298:
```suggestion
== Calling Convention for System Calls
```

kito-cheng:
Changes:
- Update per @jrtc27's suggestion.

PR #223, opened 2021-10-28
bibtex entry

samuelgruetter:
Is it ok to cite this document using the following bibtex entry? (it might useful to have it somewhere in this repo or in the readme, I thought)


cmuellner:
The description of a preferred citation form is highly appreciated.
This has also been mentioned in the review comments.

The unpriv spec solved this by the following text:
```
Please cite as: “The RISC-V Instruction Set Manual, Volume I: User-Level ISA, Document
Version 20191214-draft”, Editors Andrew Waterman and Krste Asanović, RISC-V Foundation,
December 2019.
```

I would prefer to use the "editors" ("Kito Cheng and Jessica Clarke") approach as well, instead of adding a list of 20+ authors.

Slightly off-topic, but as it has been mentioned:
My solution to the versioning issue is to use a `-wip` postfix to the version number.
So while working on a document with version `0.9.0`, I use the version `0.9.0-wip`.
Eventually, version `0.9.0-wip` becomes `0.9.0` and gets published (dropping the `-wip` can be scripted).
The next change then increases the version number to a `-wip` version (e.g. `0.9.1-wip`) again.
If it helps you might want to use something similar.

jrtc27:
Fixed in e8f6b84096bb1f52aeda269cd648351631fcee05

PR #225, opened 2021-11-01
Revise for ELF File Header section

kito-cheng:
Address those review comment from @cmuellner

* 8.1.: "We don't let users..." -> Who is "We"
   Possible solution: "It is not possible to link RV32 and RV64 code together"
* 8.1: Missing description
  Possible solution: "The section below lists the defined RISC-V-specific values for several ELF header fields."
* 8.1.: The selection of header fields looks arbitrary.
  If only RISC-V specific amendments to ELF header are specified, then why
  having e_type there (with "Nothing RISC-V specific")?
  Possible solution: Either remove e_type or name all untouched header fields
  (e_type, e_version, e_entry, e_phoff, e_shoff, e_ehsize, e_phentsize, e_phnum, e_shentsize, e_shnum, e_shstrndx)

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -72,6 +72,9 @@ for more detail on how to mangle types.
 
 === File Header
 
+The section below lists the defined RISC-V-specific values for several ELF
+header fields.
+
 e_ident::
   EI_CLASS::: Specifies the base ISA, either RV32 or RV64.
   Linking RV32 and RV64 code together is not supported.
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -73,8 +73,8 @@ for more detail on how to mangle types.
 === File Header
 
 e_ident::
-  EI_CLASS::: Specifies the base ISA, either RV32 or RV64.  We don't let users
-  link RV32 and RV64 code together.
+  EI_CLASS::: Specifies the base ISA, either RV32 or RV64.
+  Linking RV32 and RV64 code together is not supported.
 +
 --
 [horizontal]
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -85,6 +85,14 @@ e_ident::
 ELFCLASS64:::: ELF-64 Object File
 ELFCLASS32:::: ELF-32 Object File
 --
+  EI_DATA::: Specifies the endianness; either big-endian or little-endian.
+  Linking big-endian and little-endian code together is not supported.
++
+--
+[horizontal]
+ELFDATA2LSB:::: Little-endian Object File
+ELFDATA2MSB:::: Big-endian Object File
+--
 
 e_machine:: Identifies the machine this ELF file targets.  Always contains
 EM_RISCV (243) for RISC-V ELF files.  We only support RISC-V v2 family ISAs,
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -73,7 +73,8 @@ for more detail on how to mangle types.
 === File Header
 
 The section below lists the defined RISC-V-specific values for several ELF
-header fields.
+header fields; any fields not listed in this section have no RISC-V-specific
+values.
 
 e_ident::
   EI_CLASS::: Specifies the base ISA, either RV32 or RV64.
@@ -84,8 +85,6 @@ e_ident::
 ELFCLASS64:::: ELF-64 Object File
 ELFCLASS32:::: ELF-32 Object File
 --
-
-e_type:: Nothing RISC-V specific.
 
 e_machine:: Identifies the machine this ELF file targets.  Always contains
 EM_RISCV (243) for RISC-V ELF files.  We only support RISC-V v2 family ISAs,


riscv-elf.adoc
```diff
@@ -72,17 +72,34 @@ for more detail on how to mangle types.
...
+--
+[horizontal]
+ELFDATA2LSB:::: Little-endian Object File
+ELFDATA2MSB:::: Big-endian Object File
+--
++
+NOTE: RISC-V was only support little-endian, but RISC-V spec added big-endian
+since https://github.com/riscv/riscv-isa-manual/commit/2151617320bc7d27d3ce8415be306ab51c362508[Jun, 2019].
++
+  EI_ABIVERSION::: Specifies the ABI version, this field have to be set to 0,
```
jrtc27, line 100: From the gABI:

> EI_ABIVERSION
Byte e_ident[EI_ABIVERSION] identifies the version of the ABI to which the object is targeted. This field is used to distinguish among incompatible versions of an ABI. The interpretation of this version number is dependent on the ABI identified by the EI_OSABI field. If no values are specified for the EI_OSABI field by the processor supplement or no version values are specified for the ABI determined by a particular value of the EI_OSABI byte, the value 0 shall be used for the EI_ABIVERSION byte; it indicates unspecified.

We therefore neither need to specify it nor should we as it can be OS-dependent

riscv-elf.adoc
```diff
@@ -72,17 +72,34 @@ for more detail on how to mangle types.
...
 +
 --
 [horizontal]
 ELFCLASS64:::: ELF-64 Object File
 ELFCLASS32:::: ELF-32 Object File
 --
-
-e_type:: Nothing RISC-V specific.
+  EI_DATA::: Specifies the endian, either big-endian or little-endian.
+  The linker disallow linking big-endian and little code together.
```
jrtc27, line 89:
```suggestion
  EI_DATA::: Specifies the endianness; either big-endian or little-endian.
  Linking big-endian and little-endian code together is not supported.
```

riscv-elf.adoc
```diff
@@ -72,17 +72,34 @@ for more detail on how to mangle types.
...
++
+--
+[horizontal]
+ELFDATA2LSB:::: Little-endian Object File
+ELFDATA2MSB:::: Big-endian Object File
+--
++
+NOTE: RISC-V was only support little-endian, but RISC-V spec added big-endian
```
jrtc27, line 97: Does this commentary really belong here?

riscv-elf.adoc
```diff
@@ -72,17 +72,34 @@ for more detail on how to mangle types.
 
 === File Header
 
+The section below lists the defined RISC-V-specific values for several ELF
+header fields, field not listed in this section means those field didn't have
+RISC-V specific values.
```
jrtc27, line 77:
```suggestion
The section below lists the defined RISC-V-specific values for several ELF
header fields; any fields not listed in this section have no RISC-V-specific
values.
```

riscv-elf.adoc
```diff
@@ -72,17 +72,34 @@ for more detail on how to mangle types.
...
+
 e_ident::
-  EI_CLASS::: Specifies the base ISA, either RV32 or RV64.  We don't let users
-  link RV32 and RV64 code together.
+  EI_CLASS::: Specifies the base ISA, either RV32 or RV64.  The linker disallow
+  linking RV32 and RV64 code together.
```
jrtc27, line 81:
```suggestion
  EI_CLASS::: Specifies the base ISA, either RV32 or RV64.
  Linking RV32 and RV64 code together is not supported.
```

riscv-elf.adoc
kito-cheng: Yeah, that seems should be bundle with `EI_OSABI`, but we are trying to avoid include OS-depended part in here, so let remove that.

riscv-elf.adoc
kito-cheng: My original intention is to explain why it should not fixed to `ELFDATA2LSB`, but actually it's kind of weird the describe here.

kito-cheng:
Changes:
- Address @jrtc27's comment

Preview PDF:
[riscv-abi.pdf](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/files/7457482/riscv-abi.pdf)

cmuellner:
LGTM

riscv-elf.adoc
```diff
@@ -72,17 +72,28 @@ for more detail on how to mangle types.
...
-e_type:: Nothing RISC-V specific.
+  EI_DATA::: Specifies the endianness; either big-endian or little-endian.
+  Linking big-endian and little-endian code together is not supported.
++
+--
+[horizontal]
+ELFDATA2LSB:::: Little-endian Object File
+ELFDATA2MSB:::: Big-endian Object File
+--
++
```
jrtc27, line 96: Is this needed? Seems odd to me, but not an AsciiDoc expert.

kito-cheng: Updated, seems can be removed without changing layout

kito-cheng:
Changes:
- Minor clean up.

jrtc27 (approved):

PR #226, opened 2021-11-01
Revise dwarf chapter

kito-cheng:
Changes:
- Add new chapter to describe RISC-V using std DWARF spec. 
- Add introduction sentence to DWARF register number mapping table

### DIFF
--- a/riscv-dwarf.adoc
+++ b/riscv-dwarf.adoc
@@ -4,6 +4,12 @@ ifeval::["{docname}" == "riscv-dwarf"]
 include::prelude.adoc[]
 endif::[]
 
+== DWARF Debugging Format
+
+The DWARF debugging format for RISC-V follows the
+https://dwarfstd.org/[standard DWARF specification]; this specification only
+describes RISC-V-specific definitions.
+
 == DWARF Register Numbers
 
 [cols="2,2,4"]
--- a/riscv-dwarf.adoc
+++ b/riscv-dwarf.adoc
@@ -12,6 +12,9 @@ describes RISC-V-specific definitions.
 
 == DWARF Register Numbers
 
+The table below lists the mapping from DWARF register numbers to machine
+registers.
+
 [cols="2,2,4"]
 [width=80%]
 |===


kito-cheng:
@cmuellner

riscv-dwarf.adoc
```diff
@@ -4,8 +4,17 @@ ifeval::["{docname}" == "riscv-dwarf"]
 include::prelude.adoc[]
 endif::[]
 
+== DWARF Debugging Format
+
+DWARF debugging format for RISC-V follows the
+https://dwarfstd.org/[standard DWARF specification], this specification only
+describe RISC-V specific definitions.
```
jrtc27, line 11:
```suggestion
The DWARF debugging format for RISC-V follows the
https://dwarfstd.org/[standard DWARF specification]; this specification only
describes RISC-V-specific definitions.
```
though do we need to use a better way of referencing it than an inline link?

kito-cheng:
@jrtc27 

> though do we need to use a better way of referencing it than an inline link?

Maybe using footnote to doing that? I could create a separated PR for that, and also convert all other inline link.

cmuellner:
> > though do we need to use a better way of referencing it than an inline link?
> 
> Maybe using footnote to doing that? I could create a separated PR for that, and also convert all other inline link.

As a rule of thumb, we can follow the unpriv spec or do something better.
And that document has a `Bibliography` at the end.
What could be done better: use [MEANINGFUL NAME] instead of [NUMBER] as an identifier.

The rest looks good.

jrtc27:
> > > though do we need to use a better way of referencing it than an inline link?
> > 
> > 
> > Maybe using footnote to doing that? I could create a separated PR for that, and also convert all other inline link.
> 
> As a rule of thumb, we can follow the unpriv spec or do something better. And that document has a `Bibliography` at the end. What could be done better: use [MEANINGFUL NAME] instead of [NUMBER] as an identifier.

I guess we should be using https://docs.asciidoctor.org/asciidoc/latest/sections/bibliography/ rather than hand-rolling things. The numbers are the default from bibtex, unlikely a conscious choice on the unprivileged spec authors' parts (and documents that _don't_ use numbers always look odd to me as it's unusual in our field...).

cmuellner:
> > > > though do we need to use a better way of referencing it than an inline link?
> > > 
> > > 
> > > Maybe using footnote to doing that? I could create a separated PR for that, and also convert all other inline link.
> > 
> > 
> > As a rule of thumb, we can follow the unpriv spec or do something better. And that document has a `Bibliography` at the end. What could be done better: use [MEANINGFUL NAME] instead of [NUMBER] as an identifier.
> 
> I guess we should be using https://docs.asciidoctor.org/asciidoc/latest/sections/bibliography/ rather than hand-rolling things. The numbers are the default from bibtex, unlikely a conscious choice on the unprivileged spec authors' parts (and documents that _don't_ use numbers always look odd to me as it's unusual in our field...).

Certainly a matter of taste (and thus not worth to argue about).
But I would see both widely used in IT (non is unusual).
E.g. the link you provided uses named indices.

jrtc27:
> > > > > though do we need to use a better way of referencing it than an inline link?
> > > > 
> > > > 
> > > > Maybe using footnote to doing that? I could create a separated PR for that, and also convert all other inline link.
> > > 
> > > 
> > > As a rule of thumb, we can follow the unpriv spec or do something better. And that document has a `Bibliography` at the end. What could be done better: use [MEANINGFUL NAME] instead of [NUMBER] as an identifier.
> > 
> > 
> > I guess we should be using https://docs.asciidoctor.org/asciidoc/latest/sections/bibliography/ rather than hand-rolling things. The numbers are the default from bibtex, unlikely a conscious choice on the unprivileged spec authors' parts (and documents that _don't_ use numbers always look odd to me as it's unusual in our field...).
> 
> Certainly a matter of taste (and thus not worth to argue about). But I would see both widely used in IT (non is unusual). E.g. the link you provided uses named indices.

Yeah, I don't particularly mind, if the AsciiDoc default is to use the names then we should just use names. That seems to be standard practice outside the LaTeX world.

kito-cheng:
https://docs.asciidoctor.org/asciidoc/latest/sections/bibliography/ I tried this and seems work fine, but I would prefer to having a separated PR for convert those bibliography entries, other than that, is this PR ok for you?

jrtc27 (approved):
(but please squash before merging, and you'll also have to rebase the bibliography PR afterwards)

PR #227, opened 2021-11-02
Add table titles

kito-cheng:
Add title for tables

Preview PDF:
[riscv-abi.pdf](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/files/7459713/riscv-abi.pdf)

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -8,6 +8,7 @@ endif::[]
 
 === Integer Register Convention
 
+.Integer register convention
 [cols="1,1,2,2"]
 |===
 | Name      | ABI Mnemonic | Meaning                | Preserved across calls?
@@ -32,6 +33,7 @@ it must reside in x8 (s0), the register remains callee-saved.
 
 === Floating-point Register Convention
 
+.Floating-point register convention
 [cols="1,1,2,2"]
 |===
 | Name      | ABI Mnemonic | Meaning                | Preserved across calls?
@@ -54,6 +56,7 @@ duration in accordance with C11 section 7.6 "Floating-point environment
 
 === Vector Register Convention
 
+.Vector register convention
 [%autowidth]
 |===
 | Name    | ABI Mnemonic | Meaning                      | Preserved across calls?
@@ -336,6 +339,7 @@ There are two conventions for C/C++ type sizes and alignments.
 ILP32, ILP32F, ILP32D, and ILP32E:: Use the following type sizes and
 alignments (based on the ILP32 convention):
 +
+.C/C++ type sizes and alignments for RV32
 [cols="4,>2,>3"]
 [width=60%]
 |===
@@ -360,6 +364,7 @@ alignments (based on the ILP32 convention):
 LP64, LP64F, LP64D, and LP64Q:: Use the following type sizes and
 alignments (based on the LP64 convention):
 +
+.C/C++ type sizes and alignments for RV64
 [cols="4,>2,>3"]
 [width=60%]
 |===
@@ -429,6 +434,7 @@ systems. These are noted in this section.
 The following definitions apply for all ABIs defined in this document. Here
 there is no differentiation between ILP32 and LP64 ABIs.
 
+.Linux-specific C type sizes and alignments
 [cols="2,>1,>1"]
 [width=80%]
 |===
--- a/riscv-dwarf.adoc
+++ b/riscv-dwarf.adoc
@@ -15,6 +15,7 @@ describes RISC-V-specific definitions.
 The table below lists the mapping from DWARF register numbers to machine
 registers.
 
+.DWARF register number encodings
 [cols="2,2,4"]
 [width=80%]
 |===
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -142,9 +142,11 @@ EM_RISCV (243) for RISC-V ELF files.  We only support RISC-V v2 family ISAs,
 this support is implicit.
 
 e_flags:: Describes the format of this ELF file.  These flags are used by the
-linker to disallow linking ELF files with incompatible ABIs together.
+linker to disallow linking ELF files with incompatible ABIs together,
+<<e-flags-layout>> shows the layout of e_flags, and flag details are listed
+below.
 +
-
+[[e-flags-layout]]
 .Layout of e_flags
 [cols="1,2,1,1,3,5"]
 [width=80%]
@@ -154,6 +156,8 @@ linker to disallow linking ELF files with incompatible ABIs together.
 | RVC   | Float ABI  | RVE   | TSO   | *Reserved*  | *Non-standard extensions*
 |===
 
++
+--
   EF_RISCV_RVC (0x0001)::: This bit is set when the binary targets the C ABI,
   which allows instructions to be aligned to 16-bit boundaries (the base RV32
   and RV64 ISAs only allow 32-bit instruction alignment).  When linking
@@ -181,18 +185,19 @@ linker to disallow linking ELF files with incompatible ABIs together.
 
   EF_RISCV_TSO (0x0010)::: This bit is set when the binary requires the RVTSO
   memory consistency model.
-+
+
 Until such a time that the *Reserved* bits (0x00ffffe0) are allocated by future
 versions of this specification, they shall not be set by standard software.
 Non-standard extensions are free to use bits 24-31 for any purpose. This many
 conflict with other non-standard extensions.
-+
+
 NOTE: There is no provision for compatibility between conflicting uses of the
 e_flags bits reserved for non-standard extensions, and many standard RISC-V
 tools will ignore them. Do not use them unless you control both the toolchain
 and the operating system, and the ABI differences are so significant they
 cannot be done with a .RISCV.attributes tag nor an ELF note, such as using a
 different syscall ABI.
+--
 
 === Sections
 
@@ -211,7 +216,6 @@ the ABI in use.
 +
 The defined processor-specific `st_other` flags are listed in <<rv-st-other>>.
 +
-
 [[rv-st-other]]
 .RISC-V-specific `st_other` flags
 [cols="3,1"]
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -144,6 +144,8 @@ this support is implicit.
 e_flags:: Describes the format of this ELF file.  These flags are used by the
 linker to disallow linking ELF files with incompatible ABIs together.
 +
+
+.Layout of e_flags
 [cols="1,2,1,1,3,5"]
 [width=80%]
 |===
@@ -207,9 +209,11 @@ remaining 6 bits have no defined meaning in the ELF gABI. We use the highest
 bit to mark functions that do not follow the standard calling convention for
 the ABI in use.
 +
-The defined processor-specific `st_other` flags are listed in the following
-table.
+The defined processor-specific `st_other` flags are listed in <<rv-st-other>>.
 +
+
+[[rv-st-other]]
+.RISC-V-specific `st_other` flags
 [cols="3,1"]
 [width=60%]
 |===
@@ -257,6 +261,7 @@ Calculation:: Formula for how to resolve the relocation value; definitions of th
 Description:: Additional information about the relocation
 
 [[reloc-table]]
+.Relocation types
 [cols=">2,6,3,6,11"]
 [width=100%]
 |===
@@ -381,9 +386,11 @@ _RISC-V Assembly Programmer's Manual_ <<rv-asm>>.
 
 ==== Calculation Symbols
 
-The following table provides details on the variables used in relocation
+<<var-reloc-calc>> provides details on the variables used in relocation
 calculation:
 
+[[var-reloc-calc]]
+.Variables used in relocation calculation
 [%autowidth]
 |===
 | Variable  | Description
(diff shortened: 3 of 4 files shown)


kito-cheng:
@cmuellner

riscv-cc.adoc
```diff
@@ -8,6 +8,7 @@ endif::[]
 
 === Integer Register Convention
 
+.Integer register convention
```
jrtc27, line 11: These might look better under the table? Which I believe needs `table: caption_side: bottom` in the theme.

cmuellner:
Two comments:
* Table captions are typically below the table.
* Tables are often considered free floating objects in a document (in LaTex you can't even predict where they will end up). Therefore you have text, which references the tables (i.e. every table gets referenced at least once in the text). E.g. "Table 1 shows the RISC-V integer register convention.". I prefer this separation of text and tables. If you prefer to embed the tables into the text, then that's fine for me. But in this case the captions are pointless.

jrtc27:
> Two comments:
> 
> * Table captions are typically below the table.

Yeah I don't know why this is the default placement...

> * Tables are often considered free floating objects in a document (in LaTex you can't even predict where they will end up). Therefore you have text, which references the tables (i.e. every table gets referenced at least once in the text). E.g. "Table 1 shows the RISC-V integer register convention.". I prefer this separation of text and tables. If you prefer to embed the tables into the text, then that's fine for me. But in this case the captions are pointless.

Clearly you've never used float's H placement specifier :) I don't think AsciiDoc has the notion of floats in the same way, looks like it has some rudimentary support for left and right floats in HTML output but nothing for PDFs. And of course it came from the even more basic Markdown where inline tables was definitely all we had.

cmuellner:
> Clearly you've never used float's H placement specifier :)

Isn't this called the "heinous" specifier? ;)

I just checked what is being done in the `convert2adoc` branch of the RV ISA spec. E.g.:
https://github.com/riscv/riscv-isa-manual/blob/convert2adoc/src/naming.adoc (scroll to "Subset Naming Convention").

There they continued to pretend that tables are free floating objects (most likely to be as close to the tex version as possible).
So it would be reasonable to do the same.

kito-cheng:
Changes:
- Change table caption to button 
- Reference to table name instead of just `following table`

Preview PDF:
[riscv-abi.pdf](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/files/7532158/riscv-abi.pdf)

riscv-elf.adoc
```diff
@@ -154,9 +156,11 @@ remaining 6 bits have no defined meaning in the ELF gABI. We use the highest
 bit to mark functions that do not follow the standard calling convention for
 the ABI in use.
 +
-The defined processor-specific `st_other` flags are listed in the following
-table.
+The defined processor-specific `st_other` flags are listed in <<rv-st-other>>.
 +
+
+[[rv-st-other]]
+.RISC-V specific value for `st_other` flags.
```
jrtc27, line 163:
```suggestion
.RISC-V-specific `st_other` flags
```

riscv-elf.adoc
```diff
@@ -257,9 +263,11 @@ https://github.com/riscv/riscv-asm-manual[RISC-V Assembly Programmer's Manual].
 
 ==== Calculation Symbols
 
-The following table provides details on the variables used in relocation
+<<var-reloc-calc>> provides details on the variables used in relocation
 calculation:
 
+[[var-reloc-calc]]
+.Variable used in relocation calculation
```
jrtc27, line 270:
```suggestion
.Variables used in relocation calculation
```

riscv-elf.adoc
```diff
@@ -280,8 +288,10 @@ of the `__global_pointer$` symbol into register `gp` (aka `x3`).
 
 ==== Field Symbols
 
-The following table provides details on the variables used in relocation fields:
+<<var-reloc-field>> provides details on the variables used in relocation fields:
 
+[[var-reloc-field]]
+.Variable used in relocation fields
```
jrtc27, line 294:
```suggestion
.Variables used in relocation fields
```
Though variable(s) seems like the wrong word, these aren't variables, they're just names

riscv-elf.adoc
```diff
@@ -512,9 +524,11 @@ described by the ELF TLS specification, with `tp` containing the address one
 past the end of the TCB.
 
 There are various thread local storage models for statically allocated or
-dynamically allocated thread local storage. The following table lists the
+dynamically allocated thread local storage. <<tls-model>> lists the
 thread local storage models:
 
+[[tls-model]]
+.TLS model
```
jrtc27, line 531:
```suggestion
.TLS models
```

riscv-elf.adoc
```diff
@@ -656,8 +670,10 @@ typedef struct
 
 ==== Section Types
 
-The defined processor-specific section types are listed in following table.
+The defined processor-specific section types are listed in <<rv-section-type>>.
 
+[[rv-section-type]]
+.RISC-V specific section types
```
jrtc27, line 676:
```suggestion
.RISC-V-specific section types
```

riscv-elf.adoc
```diff
@@ -668,8 +684,10 @@ The defined processor-specific section types are listed in following table.
 
 ==== Special Sections
 
-The following table lists the special sections defined by this ABI.
+<<rv-section>> lists the special sections defined by this ABI.
 
+[[rv-section]]
+.RISC-V specific sections
```
jrtc27, line 690:
```suggestion
.RISC-V-specific sections
```

riscv-elf.adoc
```diff
@@ -682,9 +700,10 @@ The following table lists the special sections defined by this ABI.
 
 === Program Header Table
 
-The defined processor-specific segment types are listed in following
-table.
+The defined processor-specific segment types are listed in <<rv-seg-type>>.
 
+[[rv-seg-type]]
+.RISC-V specific segment types
```
jrtc27, line 706:
```suggestion
.RISC-V-specific segment types
```

riscv-elf.adoc
```diff
@@ -701,9 +720,10 @@ There are no RISC-V specific definitions relating to ELF note sections.
 
 === Dynamic Section
 
-The defined processor-specific dynamic array tags are listed in the following
-table.
+The defined processor-specific dynamic array tags are listed in <<rv-dyn-tag>>.
 
+[[rv-dyn-tag]]
+.RISC-V specific dynamic array tags
```
jrtc27, line 726:
```suggestion
.RISC-V-specific dynamic array tags
```

riscv-elf.adoc
```diff
@@ -91,6 +91,8 @@ this support is implicit.
 e_flags:: Describes the format of this ELF file.  These flags are used by the
 linker to disallow linking ELF files with incompatible ABIs together.
 +
+
```
jrtc27, line 94: This seems like it'd cause issues with the table no longer being sufficiently indented? The `+` is to tie the table as a continuation of the nested list item.

riscv-dwarf.adoc
```diff
@@ -6,6 +6,7 @@ endif::[]
 
 == DWARF Register Numbers
 
+.Mapping table between DWARF register numbers and machine register
```
jrtc27, line 9:
```suggestion
.DWARF register number encodings
```
perhaps?

jrtc27:
Still concerned about the two tables not being glued to the previous paragraph by `+`

kito-cheng:
@jrtc27 

I've adjust a little bit about the e_flags part and move the location of table, Is it better for you?

Preview PDF:
[riscv-abi.pdf](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/files/7594523/riscv-abi.pdf)

jrtc27:
Just apply
```
diff --git a/riscv-elf.adoc b/riscv-elf.adoc
index db7af2b..4563813 100644
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -135,7 +135,7 @@ Until such a time that the *Reserved* bits (0x00ffffe0) are allocated by future
 versions of this specification, they shall not be set by standard software.
 Non-standard extensions are free to use bits 24-31 for any purpose. This many
 conflict with other non-standard extensions.
-
++
 [[e-flags-layout]]
 .Layout of e_flags
 [cols="1,2,1,1,3,5"]
@@ -145,7 +145,7 @@ conflict with other non-standard extensions.
 
 | RVC   | Float ABI  | RVE   | TSO   | *Reserved*  | *Non-standard extensions*
 |===
-
++
 NOTE: There is no provision for compatibility between conflicting uses of the
 e_flags bits reserved for non-standard extensions, and many standard RISC-V
 tools will ignore them. Do not use them unless you control both the toolchain
@@ -170,7 +170,6 @@ the ABI in use.
 +
 The defined processor-specific `st_other` flags are listed in <<rv-st-other>>.
 +
-
 [[rv-st-other]]
 .RISC-V-specific `st_other` flags
 [cols="3,1"]
```
it works just fine and I don't understand why you keep breaking the list continuations and losing the indentation

riscv-elf.adoc
```diff
@@ -99,16 +99,11 @@ EM_RISCV (243) for RISC-V ELF files.  We only support RISC-V v2 family ISAs,
 this support is implicit.
 
 e_flags:: Describes the format of this ELF file.  These flags are used by the
-linker to disallow linking ELF files with incompatible ABIs together.
+linker to disallow linking ELF files with incompatible ABIs together,
+<<e-flags-layout>> show the layout of e_flags, and detail flags are listed
```
jrtc27, line 103:
```suggestion
<<e-flags-layout>> shows the layout of e_flags, and flag details are listed
```

riscv-elf.adoc
```diff
@@ -136,18 +131,31 @@ linker to disallow linking ELF files with incompatible ABIs together.
 
   EF_RISCV_TSO (0x0010)::: This bit is set when the binary requires the RVTSO
   memory consistency model.
+--
 +
+--
```
jrtc27, line 136:
```suggestion

```
should work, otherwise only getting limited benefit from the use of an open block

riscv-elf.adoc
```diff
@@ -136,18 +131,31 @@ linker to disallow linking ELF files with incompatible ABIs together.
...
 Until such a time that the *Reserved* bits (0x00ffffe0) are allocated by future
 versions of this specification, they shall not be set by standard software.
 Non-standard extensions are free to use bits 24-31 for any purpose. This many
 conflict with other non-standard extensions.
-+
+
+[[e-flags-layout]]
```
jrtc27, line 142: Actually, why'd this table even get moved at all? It should at least be above the reserved and non-standard extensions bits paragraph above, but having it as the first thing was useful so you could see the summary and then dive into the details below if desired, rather than having to scroll through all the details. Was this originally just due to the missing `+`s messing up the indentation?

riscv-elf.adoc
kito-cheng: Seems like remove those 3 lines not work well, that make next paragraph combine to EF_RISCV_TSO like this:
Selection_057

jrtc27: GitHub screwed up my suggestion... was meant to be a single blank line

riscv-elf.adoc
kito-cheng: My bad, I didn't understand this comment https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/227#discussion_r754201167 well so made this change :(

riscv-elf.adoc
kito-cheng: Oh, cool, single blank line work.

riscv-elf.adoc
kito-cheng: Moved back.

riscv-elf.adoc
```diff
@@ -109,6 +113,8 @@ linker to disallow linking ELF files with incompatible ABIs together.
 | RVC   | Float ABI  | RVE   | TSO   | *Reserved*  | *Non-standard extensions*
 |===
 
++
+--
```
jrtc27, line 117: I'd just keep the old couple of +'s, the use of `--` doesn't save you anything any more (I believe you could ditch the 145/150 ones, but two +'s vs two `--`'s doesn't really make a difference)

kito-cheng: I would prefer keep current layout, that made `Reserved` paragraph belong `e_flags` rather than `EF_RISCV_TSO`

Current layout:
Selection_058

And the layout of the old couple of +'s:
Selection_059

jrtc27: Oh I see. The current layout is what I intended. Though I think you can drop the `+`s within the `--`? Possibly also the `+` before the `--`?

kito-cheng: > Possibly also the + before the --?

Dropping that will cause `EF_RISCV_*` become same indention level as `e_flags`.

kito-cheng:
Just rebase.

riscv-elf.adoc
jrtc27: Weird. I'm the same, the documentation only covers the simple cases...

jrtc27:
If it ends up with the right formatting then sure let's just go with it. Maybe I'll play around locally at some point to see if it can be simplified, if I care enough...

So, ok to merge after squashing

kito-cheng:
Rebase and squashing.

Preview PDF:
[riscv-abi.pdf](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/files/7619872/riscv-abi.pdf)

jrtc27 (approved):

kito-cheng:
Cool, the remain issue is sync with [docs-templates](https://github.com/riscv/docs-templates), but I think it's not rush on this moment.

kito-cheng:
@jrtc27 thanks for your reviewing!!!

PR #228, opened 2021-11-03
Add 'cite as' entry

kito-cheng:
Preview PDF:
[riscv-abi.pdf](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/files/7468557/riscv-abi.pdf)


cmuellner:
See also #223

jrtc27 (approved):
WFM if Christoph is happy

cmuellner:
LGTM

PR #230, opened 2021-11-13
Add example and description of out-of-range for code model 

kito-cheng:


riscv-elf.adoc
```diff
@@ -14,6 +14,10 @@ via its literal.  And, when not, other data structures are used to help the
 code to address the memory space.  The coding conventions governing their use
 are known as code models.
 
+However some code models can't access whole addressing space, it will result
+linker error during linking stage, and user should consider to using larger code
+model to rebuild program.
+
```
cmuellner, line 20: Proposal:
However, some code models can't access the whole address space. The linker may raise an error if it cannot emit instructions to access the target address in the current code model.

riscv-elf.adoc
```diff
@@ -23,6 +27,25 @@ By using the instructions `lui` and `ld` or `st`, when referring to an object, o
 a 32-bit address literal can be produced.
 This code model is not position independent.
 
+The following assembly show how to loading value, storing value or calculating
```
cmuellner, line 30: The following instructions show how to load a value, store a value, or calculate an address in the `medlow` code model.

riscv-elf.adoc
```diff
@@ -44,6 +67,28 @@ NOTE: This is not yet a requirement as existing toolchains predating this part
 of the specification do not adhere to this, and without improvements to linker
 relaxation support doing so would regress performance and code size.
 
+The following assembly show how to loading value, storing value or calculating
+an address under medium code model.
+
```
cmuellner, line 72: The following instructions show how to load a value, store a value, or calculate an address in the medany code model.

kito-cheng:
Changes:
- Revise wording per @cmuellner's comment.

riscv-elf.adoc
```diff
@@ -23,6 +27,25 @@ By using the instructions `lui` and `ld` or `st`, when referring to an object, o
...
 This code model is not position independent.
 
+The following instructions show how to load a value, store a value, or calculate
+an address in the `medlow` code model.
+
+[,asm]
+----
+    # Load address
+    lui  a0, %hi(symbol)
+    addi a0, a0, %lo(symbol)
```
jrtc27, line 37: I'd put this last and change the comment to "Calculate address" to more closely match the above paragraph

riscv-elf.adoc
```diff
@@ -23,6 +27,25 @@ By using the instructions `lui` and `ld` or `st`, when referring to an object, o
...
+
+    # Load value from a symbol
+    lui  a0, %hi(symbol)
+    lw   a0, a0, %lo(symbol)
```
jrtc27, line 41:
```suggestion
    lw   a0, %lo(symbol)(a0)
```

riscv-elf.adoc
```diff
@@ -23,6 +27,25 @@ By using the instructions `lui` and `ld` or `st`, when referring to an object, o
...
+
+    # Store value to a symbol
+    lui  a0, %hi(symbol)
+    sw   a1, a0, %lo(symbol)
```
jrtc27, line 45:
```suggestion
    sw   a1, %lo(symbol)(a0)
```

riscv-elf.adoc
```diff
@@ -23,6 +27,25 @@ By using the instructions `lui` and `ld` or `st`, when referring to an object, o
...
+----
+
+
```
jrtc27, line 48:
```suggestion
```

riscv-elf.adoc
```diff
@@ -44,6 +67,28 @@ NOTE: This is not yet a requirement as existing toolchains predating this part
 of the specification do not adhere to this, and without improvements to linker
 relaxation support doing so would regress performance and code size.
 
+The following instructions show how to load a value, store a value, or calculate
+an address in the medany code model.
+
+[,asm]
+----
+    # Load address
```
jrtc27, line 75: Ditto re order and comment

riscv-elf.adoc
```diff
@@ -44,6 +67,28 @@ NOTE: This is not yet a requirement as existing toolchains predating this part
...
+.LA0
```
jrtc27, line 76: Labels need colons. Also elsewhere we've used `1: ... 1b`, I'd stick to that rather than use .LAn (which is a GCCism; LLVM uses the less-terse/opaque .Lpcrel_hiN).

riscv-elf.adoc
```diff
@@ -44,6 +67,28 @@ NOTE: This is not yet a requirement as existing toolchains predating this part
...
+    auipc a0, %hi(symbol)
+    lw    a0, a0, %lo(.LA1)
+
+    # Store value to a symbol
+.LA2
+    auipc a0, %hi(symbol)
+    sw    a1, a0, %lo(.LA2)
+----
+
+
```
jrtc27, line 91:
```suggestion
```

riscv-elf.adoc
```diff
@@ -44,6 +67,28 @@ NOTE: This is not yet a requirement as existing toolchains predating this part
...
+----
+    # Load address
+.LA0
+    auipc a0, %pcrel_hi(symbol)
+    addi  a0, a0, %pcrel_lo(.LA0)
+
+    # Load value from a symbol
+.LA1
+    auipc a0, %hi(symbol)
+    lw    a0, a0, %lo(.LA1)
```
jrtc27, line 83: Same as before re load/store syntax (technically yes this does parse fine but it's not canonical)

riscv-elf.adoc
```diff
@@ -14,6 +14,10 @@ via its literal.  And, when not, other data structures are used to help the
 code to address the memory space.  The coding conventions governing their use
 are known as code models.
 
+However, some code models can't access the whole address space. The linker may
+raise an error if it cannot emit instructions to access the target address in
```
jrtc27, line 18: "emit" isn't quite right, it's just changing the immediates for already-existing instructions (ignoring binutils's PC-relative -> absolute relaxation that happens to make some out-of-range addresses work, and all other relaxations that change the instruction stream but not the set of addresses reachable), but not sure I have a better way to express this. I think this is fine as is, but suggestions welcome.

cmuellner: Suggestion:
"The linker may raise an error if it cannot adjust the instructions to access the target address in the current code model."

jrtc27: Works for me

kito-cheng:
Changes:
- Address @jrtc27's comment.

riscv-elf.adoc
```diff
@@ -44,6 +66,27 @@ NOTE: This is not yet a requirement as existing toolchains predating this part
 of the specification do not adhere to this, and without improvements to linker
 relaxation support doing so would regress performance and code size.
 
+The following instructions show how to load a value, store a value, or calculate
+an address in the medany code model.
+
+[,asm]
+----
+    # Load value from a symbol
+0:
```
jrtc27, line 75: I'd put the auipc on the same line as the label, it makes it clearer it refers to the specific instruction, and saves a bit of space

kito-cheng:
Changes:
- Tweak example code for medany code model

Preview PDF:
[riscv-abi.pdf](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/files/7619347/riscv-abi.pdf)

riscv-elf.adoc
```diff
@@ -44,6 +66,24 @@ NOTE: This is not yet a requirement as existing toolchains predating this part
...
+----
+    # Load value from a symbol
+0:  auipc a0, %pcrel_hi(symbol)
+    lw    a0, %pcrel_lo(0b)(a0)
+
+    # Store value to a symbol
+1:  auipc a0, %pcrel_hi(symbol)
+    sw    a1, %pcrel_lo(1b)(a0)
+
+    # Load address
```
jrtc27, line 82: s/Load/Calculate/

jrtc27 (commented):
One typo fix, then good for squashing and merging

kito-cheng:
Updated for fixing typo.

jrtc27 (approved):

PR #240, opened 2022-01-12
riscv-elf: Use simpler equivalent expression for LO12 relocations

jrtc27:
Since LO12 only uses the low 12 bits of the value, subtracting a value
that's left-shifted by 12 cannot possibly have any effect. All this did
was sign-extend the value of lo12 from bit 12 upwards, which isn't
particularly useful, and just leads to confusion, since this unnecessary
(and backwards, given hi20's computation is based on lo12's immediate
being sign-extended, so feeding hi20 back into lo12's calculation is
somewhat circular) complexity has made its way into implementations in
toolchains and runtimes, which is bad.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -468,8 +468,8 @@ is followed by an I-Type instruction (add immediate or load) with an
 calculated like this:
 
 [horizontal]
-HI20:: `((symbol_address + 0x800) >> 12)`
-LO12:: `symbol_address - (hi20 << 12)`
+HI20:: `(symbol_address + 0x800) >> 12`
+LO12:: `symbol_address`
 
 The following assembly and relocations show loading an absolute address:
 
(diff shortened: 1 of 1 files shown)


riscv-elf.adoc
```diff
@@ -468,8 +468,8 @@ is followed by an I-Type instruction (add immediate or load) with an
 calculated like this:
 
 [horizontal]
-HI20:: `((symbol_address + 0x800) >> 12)`
-LO12:: `symbol_address - (hi20 << 12)`
+HI20:: `(symbol_address + 0x800) >> 12`
+LO12:: `symbol_address`
```
MaskRay, line 472: Omitted bitmask?

jrtc27: They’ve always been implicit here. If we wanted one for LO12 then there should also be one for HI20, as it’s currently an XLEN-bit value with an unspecified signedness.

asb (approved):
This change looks good to me. As MaskRay suggests, making the bitmask explicit might be a good further improvement.

kito-cheng (approved):
LGTM, checked with binutils implementation, that easier to understanding, and OK to me without bitmask.

PR #241, opened 2022-01-13
Clarify the range of medlow code model on rv64

kito-cheng:
Range of medlow code model is highest 2GiB and lowest 2GiB, but the spec
only say lowest 2GiB.

And also put the detail to show how those range come from.

Ref:
[1] https://www.sifive.com/blog/all-aboard-part-4-risc-v-code-models

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -21,7 +21,8 @@ in the current code model.
 === Small code model
 
 The small code model, or `medlow`, allows the code to address the whole RV32
-address space or the lower 2 GiB of the RV64 address space.
+address space or the lower 2 GiB and highest 2GiB of the RV64 address space
+(`0xFFFFFFFF7FFFF800` ~ `0xFFFFFFFFFFFFFFFF` and `0x0` ~ `0x000000007FFFF7FF`).
 By using the instructions `lui` and `ld` or `st`, when referring to an object, or
 `addi`, when calculating an address literal, for example,
 a 32-bit address literal can be produced.
@@ -42,6 +43,20 @@ an address in the `medlow` code model.
     # Calculate address
     lui  a0, %hi(symbol)
     addi a0, a0, %lo(symbol)
+----
+
+NOTE: The ranges on RV64 are not `0x0` ~ `0x000000007FFFFFFF` and
+`0xFFFFFFFF80000000` ~ `0xFFFFFFFFFFFFFFFF` due to RISC-V's sign-extension of
+immediates; the following code fragments show where the ranges come from:
+[,asm]
+----
+# Largest postive number:
+lui a0, 0x7ffff # a0 = 0x7ffff000
+addi a0, 0x7ff # a0 = a0 + 2047 = 0x000000007FFFF7FF
+
+# Smallest negative number:
+lui a0, 0x80000 # a0 = 0xffffffff80000000
+addi a0, a0, -0x800 # a0 = a0 + -2048 = 0xFFFFFFFF7FFFF800
 ----
 
 === Medium code model


kito-cheng:
@palmer-dabbelt

nick-knight:
Looks like this was introduced in https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/153.

@jim-wilson pointed it out in his review https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/153#discussion_r443149571, which apparently wasn't addressed.

kito-cheng:
@nick-knight thanks for dig out the history :)

kito-cheng:
Changes:
- Rebase

riscv-elf.adoc
```diff
@@ -44,6 +45,20 @@ an address in the `medlow` code model.
...
+immediates; the following code fragments show where the ranges come from:
+[,asm]
+----
+# Largest postive number:
+lui a0, 0x7ffff # a0 = 0x7ffff
+addi a0, 0x7ff # a0 = a0 + 2047 = 0x000000007FFFF7FF
+
+# Smallest negative number:
+lui a0, 0x80000 # a0 = 0xffffffff80000000
+addi, a0, a0, 0xfffffffffffff800 # a0 = a0 + -2048 = 0xFFFFFFFF7FFFF800
```
jrtc27, line 59: Should this be written -0x800? Both LLVM and binutils accept this syntax but it's not so clear and a bit misleading.

jrtc27 (commented):
Went ahead and fixed the grammar myself; please check you still agree with what's written

riscv-elf.adoc
kito-cheng: Yeah, `-0x800` is much clear than `0xfffffffffffff800` and verified with GNU and LLVM that produce same result.

kito-cheng:
Changes:
- Apply @jrtc27's suggestion

riscv-elf.adoc
```diff
@@ -44,6 +45,20 @@ an address in the `medlow` code model.
     addi a0, a0, %lo(symbol)
 ----
 
+NOTE: The ranges on RV64 are not `0x0` ~ `0x000000007FFFFFFF` and
+`0xFFFFFFFF7FFFF800` ~ `0xFFFFFFFFFFFFFFFF` due to RISC-V's sign-extension of
+immediates; the following code fragments show where the ranges come from:
+[,asm]
+----
+# Largest postive number:
+lui a0, 0x7ffff # a0 = 0x7ffff
```
nick-knight, line 54: I think you're missing a couple trailing zeros in the comment.

nick-knight (changes requested):

kito-cheng:
Changes:
- Fix wrong comment on calculate the medlow range for rv64

riscv-elf.adoc
```diff
@@ -44,6 +45,20 @@ an address in the `medlow` code model.
     addi a0, a0, %lo(symbol)
 ----
 
+NOTE: The ranges on RV64 are not `0x0` ~ `0x000000007FFFFFFF` and
+`0xFFFFFFFF7FFFF800` ~ `0xFFFFFFFFFFFFFFFF` due to RISC-V's sign-extension of
```
nick-knight, line 49: Um, the second range _is_ correct, though, right? (Copy/paste error?)

nick-knight (changes requested):

riscv-elf.adoc
kito-cheng: `0xFFFFFFFF7FFFF800` is what I want since `0xFFFFFFFF7FFFF800` lowest address in the highest 2GiB.

See
```
# Smallest negative number:
lui a0, 0x80000 # a0 = 0xffffffff80000000
addi a0, a0, -0x800 # a0 = a0 + -2048 = 0xFFFFFFFF7FFFF800
```

jrtc27: Compare with how 0x7fffffff is the number given here for the low range

jrtc27 (approved):

riscv-elf.adoc
```diff
@@ -44,6 +45,20 @@ an address in the `medlow` code model.
     addi a0, a0, %lo(symbol)
 ----
 
+NOTE: The ranges on RV64 are not `0x0` ~ `0x000000007FFFFFFF` and
+`0xFFFFFFFF80000000` ~ `0xFFFFFFFFFFFFFFFF` due to RISC-V's sign-extension of
+immediates; the following code fragments show where the ranges come from:
+[,asm]
+----
+# Largest postive number:
```
luismarques, line 53: Sorry for not commenting on time on this but here's a typo to fix in a follow-up PR: "postive".

PR #242, opened 2022-01-13
Define XLEN and FLEN in `Terms and Abbreviations`

kito-cheng:
We use XLEN and FLEN without define that, although that's widely used in
RISC-V world, but it would be great if we have explicitly define that.

Address review comment come from LLVM community[1].

[1] https://lists.llvm.org/pipermail/llvm-dev/2022-January/154600.html

### DIFF
--- a/introduction.adoc
+++ b/introduction.adoc
@@ -27,4 +27,6 @@ This specification uses the following terms and abbreviations:
 | PC    | Program Counter
 | TLS   | Thread-Local Storage
 | NTBS  | Null-Terminated Byte String
+| XLEN  | The width of an integer register in bits
+| FLEN  | The width of a floating-point register in bits
 |===


introduction.adoc
```diff
@@ -27,4 +27,6 @@ This specification uses the following terms and abbreviations:
 | PC    | Program Counter
 | TLS   | Thread-Local Storage
 | NTBS  | Null-Terminated Byte String
+| XLEN  | The width of an integer register in bits.
+| FLEN  | The width of an floating-point register in bits.
```
jrtc27, line 31:
```suggestion
| FLEN  | The width of a floating-point register in bits.
```

kito-cheng:
Changes:
- Updated with @jrtc27's suggestion.

kito-cheng:
Updated again :P

jrtc27 (approved):

PR #244, opened 2022-01-13
Clairfy address of struct return value is not required hold on return register

### LINKED ISSUE #238: Should the called function keep the address of memory allocated by the caller for the return value?
It seems that the document does not  describe whether the address of memory should be kept if  it is allocated by the caller for the return value and passed by as an implicit first parameter. For now, GCC will keep the value and this also assumed by GDB but LLVM not.

https://godbolt.org/z/YPe7GT5dT
https://github.com/riscv-collab/riscv-binutils-gdb/blob/riscv-binutils-2.37/gdb/riscv-tdep.c#L3206

kito-cheng:
psABI isn't specify the address should hold on return register(`a0`) or not,
GCC and LLVM are implement on different way here, GCC will hold the
address in return register, but LLVM won't.

It's an potential ABI incompatible change for GCC, fortunately the GCC
isn't rely that behavior, use return address rather than the rematerialization.

Following code can demonstrate this issue:

```
struct int4 {
   int   values[4];
};

struct int4 i4a;

struct int4 foo (struct int4 arg1)
{
  asm volatile ("li a0, 100 #force clobber a0":::"a0");
  return arg1;
}
```

Fix #238

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -153,7 +153,8 @@ to the aligned register pair rule).
 Values are returned in the same manner as a first named argument of the same
 type would be passed.  If such an argument would have been passed by
 reference, the caller allocates memory for the return value, and passes the
-address as an implicit first parameter.
+address as an implicit first parameter, the address is not required hold on
+return register (a0).
 
 The stack grows downwards (towards lower addresses) and the stack pointer shall
 be aligned to a 128-bit boundary upon procedure entry.


riscv-cc.adoc
```diff
@@ -153,7 +153,8 @@ to the aligned register pair rule).
 Values are returned in the same manner as a first named argument of the same
 type would be passed.  If such an argument would have been passed by
 reference, the caller allocates memory for the return value, and passes the
-address as an implicit first parameter.
+address as an implicit first parameter, the address is not required hold on
+return register (a0).
```
jrtc27, line 157: This should be non-normative text IMO

jrtc27:
Converted to non-normative text and tweaked the phrasing

PR #248, opened 2022-01-13
preamble: Clarify what frozen etc mean in this context

jrtc27:
This specification is rather unique in the terms being mostly
meaningless from an implementor's perspective.

### DIFF
--- a/preamble.adoc
+++ b/preamble.adoc
@@ -35,3 +35,17 @@ Kito Cheng and Jessica Clarke, RISC-V International, {monthyear}.
 
 The latest version of this document can be found here:
 https://github.com/riscv-non-isa/riscv-elf-psabi-doc.
+
+This specification is written in collaboration with the development communities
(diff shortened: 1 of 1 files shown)


preamble.adoc
```diff
@@ -35,3 +35,17 @@ Kito Cheng and Jessica Clarke, RISC-V International, {monthyear}.
 
 The latest version of this document can be found here:
 https://github.com/riscv-non-isa/riscv-elf-psabi-doc.
+
+This specification is written in collaboration with the development communities
+of the major open-source toolchain and operating system communities, and as
+such specifies what has been agreed upon and implemented. As a result, any
```
jrtc27, line 41: This isn't entirely accurate, as we only require a PoC, not an upstreamed implementation (which would lead to a circular dependency) in order for things to be merged, but it's close enough IMO to make the point

kito-cheng (approved):
LGTM, let wait few more days make sure no further feedback.

Issue #249, opened 2022-01-13
Disclaim that ILP32E is not stable


jrtc27:
RV32E is not ratified, so it is not possible for us to promise that the ILP32E ABI is stable. We should clearly indicate that it is not stable yet despite what the rest of the document may say.

aswaterman:
In case anyone is wondering, ratifying RV32E is something we're hoping to do in the next few months.  I wouldn't expect any changes (but all of the usual caveats apply).

PR #250, opened 2022-01-13
riscv-cc: Document that ILP32E is not and cannot be stable

jrtc27:
Closes: #249

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -244,6 +244,11 @@ provided they hold values no more than FLEN bits wide.
 
 === ILP32E Calling Convention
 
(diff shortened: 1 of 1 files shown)


kito-cheng (approved):
Agree with that, although `rv32e*`/`ilp32e` is used by some vendors on production as I know, so personally I don't like to made any incompatible change if possible, but the real problem is `rv32e` is not stable as you describe, so we can't guarantee anything about that, adding such disclaimer is good to me.

kito-cheng:
I believe it's harmless even rv32e is ratified soon, we can remove this disclaimer later.

PR #251, opened 2022-01-14
The medany code model is not PIC

### LINKED ISSUE #245: GCC generates non-position-independent code under -cmodel=medany
The objects we generate from GCC for the medium code model, aka -cmmodel=medany, are not position independent unless -fPIC is also passed.  This violates the psABI spec, which says under the medium code model section that "This code model is position independent."

aswaterman:
Looks like this incorrect description snuck in in 991c1b5f0bac58388921bf9dc64ecb6b06824115

Resolves #245

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -53,7 +53,7 @@ and `ld` or `st`, when referring to an object, or
 `addi`, when calculating an address literal, for example,
 a signed 32-bit offset, relative to the value of the `pc` register,
 can be produced.
-This code model is position independent.
+This code model is not position independent.
 
 As a special edge-case, undefined weak symbols must still be supported, whose
 addresses will be 0 and may be out of range depending on the address at which


jrtc27:
I'm struggling to think of a situation where medany isn't position-independent? It obviously doesn't follow the symbol preemption rules for -fPIC, but the difference between medany and -fPIE is almost non-existent. The only difference I can think of is -fPIE will avoid PC-relative addressing for symbols not defined in the current translation unit, since those might require copy relocations, but that doesn't make it any more position independent, it just makes it more dynamic linking friendly, which is orthogonal.

So on a technical level yes you didn't ask for -fPIE, you've asked for position-dependent code that uses PC-relative addressing, but I think the point it's trying to convey is that, ignoring what you ask the linker to do (after all, you can link -fPIC code as position dependent), it is position independent?

aswaterman:
@jrtc27 AFAICS, the main problem is static data initialized to addresses _within_ the translation unit.  These get filled in at link time, and unlike in PIE they aren't fixed up with dynamic relocs at program start.

aswaterman:
You're right, I'm conflating the code model with the linking model (or whatever the correct term for the latter is).

I guess I'll close this and move the discussion to #245

PR #252, opened 2022-01-14
Add `Medium position independent code model` to code model

kito-cheng:
And remove `This code model is position independent.` and
`This code model is not position independent.` from small and medium
code model to prevent confusion.

Try to address #245 .


riscv-elf.adoc
```diff
@@ -84,6 +82,34 @@ an address in the medany code model.
     addi  a0, a0, %pcrel_lo(2b)
 ----
 
+NOTE: Note, that although the generated code is position independent, but
+it's not suitable for building position-independent ELF objects (e.g. shared
```
jrtc27, line 86: The model is totally fine for PIEs, it's just shared objects that are the issue. PIEs currently prefer the GOT like shared objects, but they technically don't _have_ to, they just do to avoid copy relocations.

kito-cheng: Yeah, let me think how to tweak that...maybe `it's not suitable for building ELF shared object.`?

kito-cheng:
Changes:
- Update note for medany code model

jrtc27:
Will make an editorial pass for grammar over this but otherwise ready

jrtc27:
(now done)

PR #253, opened 2022-01-28
Fix typos for CALL, CALL_PLT and GOT_HI20

rui314:
`CALL` and `CALL_PLT` relocations refer `auipc` and `jalr` instruction pairs.
`jalr` is not a J-type instruction; it's I-type.

As to `GOT_HI20`, `G` means the offset inside the .got in the other ELF specs.
So the correct expression is not `G+A-P` but `G+GOT+A-P` where `GOT` means
the address of the .got.

I found these errata when I was trying to implement a RISC-V support to my linker.
https://github.com/rui314/mold/commit/e76f7c0d474e667cc0eccac5281709664539b714

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -299,12 +299,12 @@ Description:: Additional information about the relocation
                                             <| S + A - P
 .2+| 17      .2+| JAL           .2+| Static  | _J-Type_          .2+| 20-bit PC-relative jump offset
                                             <| S + A - P
-.2+| 18      .2+| CALL          .2+| Static  | _U+J-Type_        .2+| 32-bit PC-relative function call, macros `call`, `tail`
+.2+| 18      .2+| CALL          .2+| Static  | _U+I-Type_        .2+| 32-bit PC-relative function call, macros `call`, `tail`
                                             <| S + A - P
-.2+| 19      .2+| CALL_PLT      .2+| Static  | _U+J-Type_        .2+| 32-bit PC-relative function call, macros `call`, `tail` (PIC)
+.2+| 19      .2+| CALL_PLT      .2+| Static  | _U+I-Type_        .2+| 32-bit PC-relative function call, macros `call`, `tail` (PIC)
                                             <| S + A - P
 .2+| 20      .2+| GOT_HI20      .2+| Static  | _U-Type_          .2+| High 20 bits of 32-bit PC-relative GOT access, `%got_pcrel_hi(symbol)`
-                                            <| G + A - P
+                                            <| G + GOT + A - P
 .2+| 21      .2+| TLS_GOT_HI20  .2+| Static  | _U-Type_          .2+| High 20 bits of 32-bit PC-relative TLS IE GOT access, macro `la.tls.ie`
                                             <|
 .2+| 22      .2+| TLS_GD_HI20   .2+| Static  | _U-Type_          .2+| High 20 bits of 32-bit PC-relative TLS GD GOT reference, macro `la.tls.gd`
@@ -402,6 +402,7 @@ calculation:
 | A         | Addend field in the relocation entry associated with the symbol
 | B         | Base address of a shared object loaded into memory
 | G         | Offset of the symbol into the GOT (Global Offset Table)
+| GOT       | Address of the GOT (Global Offset Table)
 | P         | Position of the relocation
 | S         | Value of the symbol in the symbol table
 | V         | Value at the position of the relocation
@@ -437,7 +438,7 @@ of the `__global_pointer$` symbol into register `gp` (aka `x3`).
 | _S-Type_    | Specifies a field as the immediate field in an S-type instruction
 | _U-Type_    | Specifies a field as the immediate field in an U-type instruction
 | _J-Type_    | Specifies a field as the immediate field in a J-type instruction
-| _U+J-Type_  | Specifies a field as the immediate fields in a U-type and J-type instruction pair
+| _U+I-Type_  | Specifies a field as the immediate fields in a U-type and I-type instruction pair
 |===
 
 ==== Constants
@@ -576,7 +577,7 @@ and when symbol has an `@plt` suffix it expands to:
 
 ==== PC-Relative Jumps and Branches
 
-Unconditional jump (U+J-Type) instructions have a `R_RISCV_JAL` relocation
+Unconditional jump (J-Type) instructions have a `R_RISCV_JAL` relocation
 that can represent an even signed 21-bit offset (-1MiB to +1MiB-2).
 
 Branch (SB-Type) instructions have a `R_RISCV_BRANCH` relocation that


luismarques:
Thanks for the patch. This makes sense to me.

> As to `GOT_HI20`, `G` means the offset inside the .got in the other ELF specs. So the correct expression is not `G+A-P` but `G+GOT+A-P` where `GOT` means the address of the .got.

Table 4 (Variables used in relocation calculation) defines the terms. That table should also be updated.

MaskRay:
> The indirect jump instruction JALR (jump and link register) uses the I-type encoding.

LG

jrtc27:
> As to GOT_HI20, G means the offset inside the .got in the other ELF specs.
So the correct expression is not G+A-P but G+GOT+A-P where GOT means
the address of the .got.

This isn't true. In the AArch64 ELF ABI G (well, G(sym) because they're overly pedantic) is defined as the address of the GOT entry. Given we only ever want the address of the GOT entry, not its index, we should just fix the definition of G in the table to be as intended.

riscv-elf.adoc
```diff
@@ -576,7 +576,7 @@ and when symbol has an `@plt` suffix it expands to:
 
 ==== PC-Relative Jumps and Branches
 
-Unconditional jump (U+J-Type) instructions have a `R_RISCV_JAL` relocation
+Unconditional jump (U+I-Type) instructions have a `R_RISCV_JAL` relocation
```
jrtc27, line 579: ... except this should be J-type

riscv-elf.adoc
```diff
@@ -299,12 +299,12 @@ Description:: Additional information about the relocation
                                             <| S + A - P
 .2+| 17      .2+| JAL           .2+| Static  | _J-Type_          .2+| 20-bit PC-relative jump offset
                                             <| S + A - P
-.2+| 18      .2+| CALL          .2+| Static  | _U+J-Type_        .2+| 32-bit PC-relative function call, macros `call`, `tail`
+.2+| 18      .2+| CALL          .2+| Static  | _U+I-Type_        .2+| 32-bit PC-relative function call, macros `call`, `tail`
```
jrtc27, line 302: I agree with these changes

jrtc27 (commented):
Somehow we all managed to miss the incorrect changes in #168, ouch, every single change was wrong... too focused on naming I guess that we didn't stop and think about whether the changes were correct

riscv-elf.adoc
jrtc27: (since it's for a JAL instruction, not an AUIPC+JALR pair)

rui314:
> > As to GOT_HI20, G means the offset inside the .got in the other ELF specs. So the correct expression is not G+A-P but G+GOT+A-P where GOT means the address of the .got.
>
> This isn't true. In the AArch64 ELF ABI, G (well, G(sym) because they're overly pedantic) is defined as the address of the GOT entry. Given we only ever want the address of the GOT entry, not its index, we should just fix the definition of G in the table to be as intended.

x86-64, i386,  MIPS and SPARC psABI documents all use `G` to indicate "the offset into the global offset table at which the address of the relocation entry’s symbol will reside during execution". AArch64 is indeed an outlier, but it's as you pointed out not `G` but `G(expr)` (the psABI says "`G(expr)` is the address of the GOT entry for the expression expr"). To my knowledge, just `G` is exclusively used to indicate the offset in the .got. Therefore, I think it  is better to follow that convention in this document for the sake of consistency.

riscv-elf.adoc
rui314: Fixed.

rui314:
I added `GOT` to Table 4. Please take a look at this again.

jrtc27:
> > > As to GOT_HI20, G means the offset inside the .got in the other ELF specs. So the correct expression is not G+A-P but G+GOT+A-P where GOT means the address of the .got.
> > 
> > 
> > This isn't true. In the AArch64 ELF ABI, G (well, G(sym) because they're overly pedantic) is defined as the address of the GOT entry. Given we only ever want the address of the GOT entry, not its index, we should just fix the definition of G in the table to be as intended.
> 
> x86-64, i386, MIPS and SPARC psABI documents all use `G` to indicate "the offset into the global offset table at which the address of the relocation entry’s symbol will reside during execution". AArch64 is indeed an outlier, but it's as you pointed out not `G` but `G(expr)` (the psABI says "`G(expr)` is the address of the GOT entry for the expression expr"). To my knowledge, just `G` is exclusively used to indicate the offset in the .got. Therefore, I think it is better to follow that convention in this document for the sake of consistency.

MIPS and SPARC don't use PC-relative GOT relocations, their relocations are for GOT indices, so defining G to be the GOT index makes sense. The same is true for i386. For x86_64 it makes sense for it to inherit its 32-bit ancestor's definitions. It does not make sense for us to mess about defining `G` and `GOT` for everything to then be written as `GOT + G`, it's unnecessarily verbose, we should just define `G` as what we actually want it to be, which is how it's currently used. If AArch64 can define `G(expr)` to be what makes sense for them we can define `G` to be what makes sense for us; I disagree that the additional argument makes a difference.

rui314:
> MIPS and SPARC don't use PC-relative GOT relocations, their relocations are for GOT indices, so defining G to be the GOT index makes sense. The same is true for i386. For x86_64 it makes sense for it to inherit its 32-bit ancestor's definitions. It does not make sense for us to mess about defining G and GOT for everything to then be written as GOT + G, it's unnecessarily verbose, we should just define G as what we actually want it to be, which is how it's currently used. If AArch64 can define G(expr) to be what makes sense for them we can define G to be what makes sense for us; I disagree that the additional argument makes a difference.

You can freely define `G` as the address in the GOT in your psABI spec, but that's simply too confusing. AArch64 redefines `G` as a function rather than a variable, so there's no chance to confuse `G(expr)` with `G`. But if you redefine `G` to have a different meaning without changing the notation at all, I'm sure that that'll confuse lots of readers of this spec.

This very document is a good example of the confusion of the notation conflict. You guys defines `G` as the offset in the GOT in Table 4 but uses `G` as the address of the GOT entry in Table 3. Even the authors of the psABI can make such mistake by redefining `G`, I think it is unavoidable for readers to make the same mistake.

The easiest way to avoid that confusion is to write it as `G + GOT`. Just adding `+ GOT` in one place can make it clear what that means to those who are familiar with other psABIs. Or, you can introduce a new symbol other than `G` to indicate the address in the GOT. But please do not redefine existing, established notation for a similar but different purpose.

kito-cheng:
My first impression is fix the `G` notation become `Address of the GOT (Global Offset Table) plus the offset of the symbol into the GOT`, then the new symbol is not needed anymore but I realized @rui314's point is `G` already widely used in another target, the exception is AArch64, but AArch64 redefine that as function rather than a symbol, so that could prevent confusion when implement the linker.

The major reader/user of this part would be the system software developers like @rui314, they well know lots of system software knowledge amount different architecture, but they might not RISC-V expert, so redefine those symbol might misleading those developer.

So I think current we have following options:
- Introduce `GOT` symbol (@rui314's proposal)
- Re-define the `G` to a function rather than symbol like AArch64.
- Define another symbol other than `G` to prevent confusion, maybe `GO` or ?

I don't have strong opinion for those options, but I would prefer made a change here rather than just update the definition of `G`.

kito-cheng:
@rui314 We've discuss this issue in last psABI call, and we agree redefine that is too ambiguous, and prefer to define `G` as function as `AArch64`, did you mind update with that? Otherwise this PR is LGTM :)

[1] Meeting minutes: https://github.com/riscv-admin/psabi/blob/master/MINUTES/meeting-20220207.adoc

rui314:
OK, great! But what does the `G` take?

kito-cheng:
I would prefer using same term as `AArch64`, take an expression (`expr`), thanks :)

rui314:
Do you want me to introduce another symbol `GDAT` just like AArch64? Here is the AArch64 psABI. https://developer.arm.com/documentation/ihi0056/g/

rui314:
For RISC-V, `G`'s argument will always be `S+A`, so we'll always write `G(S+A)` to indicate the address of the GOT entry. But I'm not sure if it is any better than `G + GOT`.

Is redefining `G` as a function and always using it as `G(S+A)` better than always writing `G + GOT` (which is in line with other psABIs)? I honestly think that we are creating a bigger problem than the one we are solving with that...

kito-cheng:
Oh, I got your point, but I am concern about the addend (`A`) is kind of weird used in current expression, although we are always emit zero addend for now, but that is broken if addend is non-zero, e.g. `A` = 8 in RV64 means point to next GOT entry of `S`?

The problem is not introduce by your PR and existing here for long time, but I think using `G(S+A) - P` could fix that issue, but `G + GOT + A - P` isn't, that's why I prefer introduce `G(expr)`, or we can fix that issue by changing `G + GOT - P` that to fix that issue, did you have suggestion on this issue?

rui314:
As an expected reader of the document, I'd appreciate if it would just be `G + GOT - P`. I'm familiar with these symbols and their definitions already, and they are easy to understand. I actually use these symbols as macros as you can see here: https://github.com/rui314/mold/blob/main/elf/arch-x86-64.cc#L284. It wanted to keep the code in sync with the spec, so a simpler notation would be appreciated.

kito-cheng (approved):
LGTM for `G + GOT + A - P`, I think the problem is out of scope for this PR, we could having separated issue / PR for that, @jrtc27 let me know what your thought, are you OK with this version?

jrtc27 (approved):
Let's just go with this and move on

PR #254, opened 2022-02-04
Define the large code model

ebahapo:
**This is a working draft.**

### DIFF
--- a/riscv-large.md
+++ b/riscv-large.md
@@ -1,0 +1,473 @@
+# RISC-V Large Code Model
+
+#### Evandro Menezes
+#### evandro.menezes@sifive.com
+
+
+## Introduction
+
+Code models are based on the portion of the total address space that may be reached.  The native range of instructions usually determines the code models.  The range of instructions such as branches and loads may determine different code models to reach code and data in the address space.
+
+When the code or data of a program is within the native range of instructions, it is deemed "small". When the code or data of a program is without the native range of instructions, it is deemed "large". Code models are sometimes defined as:
+
+| Code models    | Small code | Large code
+| ---            | ---        | ---
+| **Small data** | Small      | Medium
+| **Large data** | Compact    | Large
+
+
+## RISC-V Data Models
+
+| Data models | Address space | Code addressing range | Data addressing range
+| ---         | ---           | ---                   | ---
+| **ILP32**   | 4GiB          | 4GiB                  | 4GiB
+| **LP64**    | 16EiB         | 2GiB                  | 4GiB
+
+For ILP32, the whole address space is within the native range for both code and data. Therefore, the small code model is sufficient.
+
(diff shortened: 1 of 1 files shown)


riscv-large.md
```diff
@@ -0,0 +1,473 @@
...
+
+### Function Prologue
+
+Functions in executable objects expect that the `gp` register was initialized to the global pointer at program start up and the `gp` may be used as the base register to access objects in the global data area.
+
+Functions in shared objects that refer to the global pointer must setup an arbitrary register in its prologue instead which must be used as the base register to access objects in the global data area.  For example:
+
+```assembly
+  auipc	s0, %pcrel_hi(__global_pointer__)
+  addi	s0, s0, %pcrel_lo(__global_pointer__)
```
jrtc27, line 106: This doesn't work if you have more then 2G-4G of code in the object

riscv-large.md
```diff
@@ -0,0 +1,473 @@
...
+```
+
+The expression `__global_pointer$ - .` results in the relocation type `R_RISCV_64_PCREL`. This relocation is position independent, provided that the information available to the linker matches the run time environment, therefore it can be used in both executable and shared objects.
+
+Note that the code example above places the literal for the global data area in a `comdat` section that is shared by all references to this literal in the object.
+
+
+### Program Linkage Table
+
+Each PLT entry follows the pattern below:
```
jrtc27, line 126: How does the linker know which you want?

riscv-large.md
```diff
@@ -0,0 +1,473 @@
...
+  auipc	t0, %hi_pcrel(2f)	# address of 2f
+  addi	t0, %lo_pcrel(1b)
+  ld	t2, (t0)		# difference between .got.plt - 2f
+  add	t0, t0, t2		# address of .got.plt
+  add	t0, t0, t3		# address of <function>@.got.plt
+  ld	t3, (t0)		# address of <function>
+  jr	t3
+  nop
+2:
+  .quad	.got.plt - ., 0
```
jrtc27, line 148: This either forbids lazy binding or requires RWX PLT entries; the latter is something that is strongly discouraged and has been deliberately moved away from on most architectures

riscv-large.md
```diff
@@ -0,0 +1,473 @@
...
+The stub above may be repeated as needed in order to bring it within range of a referring PLT entry.
+
+
+### Data Objects
+
+The code examples below assume that the `gp` register, only used in executable objects, points to the global data area.
+
+| Source			| Assembly				| Relocations
+| --				| --					| --
+| `extern int src;`		| `.extern src`				| `R_RISCV_64`
```
jrtc27, line 160: No they're not, declarations generate no code

rui314:
From a linker author's perspective, we want to turn a PLT-indirect reference to a direct reference if a referenced symbol is within reach. Does this spec allow such relaxation?

kito-cheng:
Prefer #388, so close this PR

PR #256, opened 2022-03-18
riscv-cc: Describe that vxrm/vxsat is not preserved across calls

arcbbb:
There are discussions in https://github.com/riscv/riscv-v-spec/issues/739.
The reason that ISA makes the vx round mode only available in CSR is for encoding space.
I think we can make it non-preserved across calls to use it efficiently.

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -64,6 +64,8 @@ duration in accordance with C11 section 7.6 "Floating-point environment
 | v0-v31  |              | Temporary registers          | No
 | vl      |              | Vector length                | No
 | vtype   |              | Vector data type register    | No
+| vxrm    |              | Vector fixed-point rounding mode register    | No
+| vxsat   |              | Vector fixed-point saturation flag register  | No
 |===
 
 
@@ -71,7 +73,8 @@ Vector registers are not used for passing arguments or return values; we
 intend to define a new calling convention variant to allow that as a future
 software optimization.
 
-The `vxrm` and `vxsat` fields of `vcsr` have thread storage duration.
+The `vxrm` and `vxsat` fields of `vcsr` are not preserved across calls, which means
+their values are unknown on entry to/return from a call.
 
 Procedures may assume that `vstart` is zero upon entry. Procedures may assume
 that `vstart` is zero upon return from a procedure call.


kito-cheng:
Current description already implied call-clobber/caller-save, did you have practical example to show the advantage or intention behind this change?

aswaterman:
I believe these should actually be treated the same as `frm`/`fflags`: that is, they have thread storage duration.  (In other words, they are effectively global variables within a thread context.)

cc @kasanovic

aswaterman (changes requested):
I believe these should actually be treated the same as frm/fflags: that is, they should have thread storage duration.

cc @kasanovic

arcbbb:
Since it is already call-clobbered, I think I have no problem now. Thanks.

arcbbb:
I have a case to show the reason not to preserve vcsr/vxrm/vxsat across calls like VL/VType does.

Current ABI allows VXRM to be set in one function (e.g. foo()) and to be used in another function (e.g. bar()).
```C
vint8m1_t test0(vint8m1_t dest, vint8m1_t op1, vint8m1_t op2, size_t vl)
{
    vint8m1_t t2 = vaadd_rdn_vv_i8m1(op1, op2, vl);
    foo();  // set VXRM to specific mode
    bar();  // user of the specific mode set by foo()
    ...
}
```
This imposes a strong order on intrinsics which use VXRM in function `test0`
The expected codegen of intrinsic `vaadd_rdn_vv_i8m1` [1] is
```asm
  csrwi vxrm, 2 ; round-down
  vaadd.vv v8, v9, v10
```
Once the intrinsic `vaadd_rdn_vv_i8m1` is moved to the between of the foo() and bar() by the compiler,
```diff
vint8m1_t test0(vint8m1_t dest, vint8m1_t op1, vint8m1_t op2, size_t vl)
{
-   vint8m1_t t2 = vaadd_rdn_vv_i8m1(op1, op2, vl);
    foo();  // set VXRM to specific mode
+   vint8m1_t t2 = vaadd_rdn_vv_i8m1(op1, op2, vl);
    bar();  // user of the specific mode set by foo()
    ...
}
```
the codegen would become the asssembly below:
```asm
   csrrwr a0, vxrm, 2   // Previous mode is unknown, then save it before setting it.
   vaadd.vv v8, v9, v10
   csrw vxrm, a0         // Restore the previous mode for bar();
```
If VXRM is not preserved across calls, then the compiler doesn't need to generate the instructions to save and restore the mode.


[1] `vaadd_rdn_vv_i8m1` is in the proposal https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/144

aswaterman:
I am still trying to understand why we don't treat `vxrm` and `vxsat` the same as `frm` and `fflags` (i.e. neither call-preserved nor call-clobbered, but thread-local).

arcbbb:
> I am still trying to understand why we don't treat `vxrm` and `vxsat` the same as `frm` and `fflags` (i.e. neither call-preserved nor call-clobbered, but thread-local).

I think the corresponding registers on X86 of the `frm` and `fflags` are the control bits and status bits of 
 `MXCSR`.
As [x86-64 psABI](https://github.com/hjl-tools/x86-psABI/wiki/x86-64-psABI-1.0.pdf) says
```
The control bits of the MXCSR register are callee-saved (preserved across
calls), while the status bits are caller-saved (not preserved).
```
my understanding is that 
when pragma FENV ACCESS is OFF, FP round mode is callee-saved. floating point operation can be optimized assuming they are all round-to-nearest-even in the caller function.
And when pragma FENV ACCESS is ON, callee's change on FP round mode can be observed by the caller.
Floating point operations need to be kept in strong order in the caller function and then the compiler would disable some floating-point optimizations.

I think it follows the same reason why we make VL/VTYPE non-preserved across calls and adopt explicit vl model on rvv intrinsics.

Based the [discussion](https://github.com/riscv/riscv-v-spec/issues/739), flipping the fixed-point rounding mode could happen very frequently. And current compiler implementation doesn't perform CSE on fixed-point [rvv intrinsics](https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/144).

I think calling convention of X86 Direction Flag is one possible option we can learn.

X86 ABI says
```
The direction flag DF in the %rFLAGS register must be clear (set to “forward”
direction) on function entry and return.
```
[Calling convention guide](https://www.agner.org/optimize/calling_conventions.pdf) by Agner Fog
```
Direction flag
The rules for the direction flag is the same in all systems. The direction flag is cleared by
default. If the direction flag is set, then it must be cleared again before any call or return. 
```

The default value on the function entry/return would introduce some issue for interrupt handler.
I see llvm clears the direction flag explicitly in the function prolog when the calling convention is X86_INTR.

aswaterman:
The x86 psABI stuff is a distraction, IMO.  The C language's floating-point environment, which includes both the flags and the rounding mode, is defined to have thread storage duration [1], and that's true regardless of ABI.

[1] http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf page 224

kito-cheng:
IIUC, you are trying to making `vxrm` changing implicitly without restore, and that's intrinsic design issue than ABI issue IMO, let me elaborate that by talking about the fenv first.

And why I using ﬂoating-point environment in C as reference is because I think we should align the behavior between `vxrm`/`vxsat` and `frm`/`fflags`.
 
ISO/IEC 9899:2011: 7.6:
```
1) The header <fenv.h> deﬁnes several macros, and declares types and functions that
provide access to the ﬂoating-point environment.  The ﬂoating-point environment refers
collectively to any ﬂoating-point status ﬂags and control modes supported by the
implementation.204) A ﬂoating-point status ﬂag is a system variable whose value is set
(but never cleared) when a ﬂoating-point exception is raised, which occurs as a side effect
of exceptional ﬂoating-point arithmetic to provide auxiliary information.205) A ﬂoating-
point control mode is a system variable whose value may be set by the user to affect the
subsequent behavior of ﬂoating-point arithmetic.
```

and more text from ISO/IEC 9899:2011: 7.6:
```
2) The ﬂoating-point environment has thread storage duration. The initial state for a
thread’s ﬂoating-point environment is the current state of the ﬂoating-point environment
of the thread that creates it at the time of creation.
```

That give us the reason why define `frm`/`fflags` as thread local storage rather than callee-save or caller-save like Andrew said in other comment, it's more like a global variable in the execution environment, and you won't say a global variable is callee-save or caller-save :p

---

And why I think that's intrinsic design issue than ABI issue? ISO/IEC 9899:2011: 7.6.3: give some description on that:

```
Certain programming conventions support the intended model of use for the ﬂoating-
point environment:206)
— a function call does not alter its caller’s ﬂoating-point control modes, clear its caller’s
ﬂoating-point status ﬂags, nor depend on the state of its caller’s ﬂoating-point status
ﬂags unless the function is so documented;
— a function call is assumed to require default ﬂoating-point control modes, unless its
documentation promises otherwise;
— a function call is assumed to have the potential for raising ﬂoating-point exceptions,
unless its documentation promises otherwise.
```

So based on the conventions described in C11 spec, yes, you *CAN* design those fixed-point operation intrinsic functions with the changing `fenv`, but I think it's might be a bad idea since those intrinsic functions is trying to emulate the static rounding mode by setting dynamic rounding mode, and those intrinsic function come with side effect of changing default rounding isn't make too much sense, I would strongly prefer those flags should only changed on explicitly function call, like fesetround in fenv stuffs.

nick-knight (approved):
I understand the terms "call-preserved" and "call-clobbered" as referring to architectural registers, and the term "thread storage duration" as referring to C-language objects. The C floating-point environment, let's call it "`fenv`", is an (implementation-defined) C-language object with thread storage duration, accessible via the various `fe*` utility functions. I feel uncomfortable with how the ELF psABI refers to `fcsr` as having thread storage duration: I think it's really talking about `fcsr`'s `frm` and `fflags` subfields' C-language representations within `fenv`. It's up to the C compiler/runtime to keep `frm`/`fflags` consistent with `fenv` across thread-context switches (specified by the C language) and across function calls (specified by the ABI calling convention). And since the ELF psABI does not explicitly mention whether `frm`/`fflags` are call-preserved or call-clobbered, I'm assuming they default to the latter.

The situation is murkier for architectural state associated with the RISC-V vector extension, since there is currently no "`venv`" analogue in the C language. My opinion is that we should treat `vxrm`/`vxsat` by analogue with `frm`/`fflags`. I'm less clear what we should say about other vector state (`vstart`, `vtype`, `vl`, `vlenb`), since there is no clear analogue in C.

In summary, I support @arcbbb 's PR and I would go further by making the same change to the floating-point section: clarifying `frm`/`fflags` as call-clobbered and removing mention of thread storage duration. Separately, I would suggest defining `venv` elsewhere, perhaps in the C API doc.

A separate question is whether `fenv` or `venv` are useful constructs to programmers. Another question is how C-language extensions, like RISC-V vector intrinsics, interact with them. I don't think these details need to mentioned in the ABI.

This is my understanding, feel free to correct it.

arcbbb:
> In summary, I support @arcbbb 's PR and I would go further by making the same change to the floating-point section: clarifying `frm`/`fflags` as call-clobbered and removing mention of thread storage duration.

If the calling convention of `frm` & `fflags` changed, I am not sure how not to break a C conforming program. so I only address vxrm & vxsat here.

> A separate question is whether `fenv` or `venv` are useful constructs to programmers. Another question is how C-language extensions, like RISC-V vector intrinsics, interact with them. I don't think these details need to mentioned in the ABI.
> 
> This is my understanding, feel free to correct it.

From the language perspective, I attempt to make the RISC-V vector intrinsics being context-free
in proposal [[1]](https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/144) & [[2]](https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/147).
I think we can move to rvv-intrinsic-doc for the follow-up discussion.

aswaterman (changes requested):
@nick-knight Your point is entirely valid that the C language requirements for `fenv` need have nothing to do with the ABI discipline for `fcsr`.  But I'd argue that coupling them is both simple and performant.

I think we agree on the semantics of the hypothetical C concept of `venv`, but I still don't see how this PR helps.  As Kito mentioned above in https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/256#issuecomment-1076197273, the semantics don't match the concept of callee-saved or caller-saved.  They're just a different thing.

arcbbb:
Assume we have a kernel function like this
```C
void f(uint16_t c[256], uint16_t a[256], uint16_t b[256])
{
  for (int i = 0; i < 256; i++) {
    c[i] = (a[i] + b[i] + 1) / 2;
  }
}
```
and the compiler automatically vectorizes it into this
(It is a made-up example. correct me if there are mistakes)
```asm
f:
    li a6, 256
    vsetvli a4, a6, e16, m1, ta, mu
loop:
    vle8.v  v8, (a1)
    vle8.v  v9, (a2)
    csrwi vxrm, 0    ; round up mode
    vaadd.vv    v8, v8, v9
    vse8.v  v8, (a0)
    sub a6, a6, a4
    add a1, a1, a4
    add a2, a2, a4
    add a0, a0, a4
    vsetvli a4, a6, e16, m1, ta, mu
    bnez    a4, loop
    ret 
```
It is done if `vxrm` is not preserved across calls.
But given the programming convention of FENV, user is not aware that the function could change the `frm` state. The compiler has to save and restore the state like this
```diff
f:
    li a6, 256
    vsetvli a4, a6, e16, m1, ta, mu
loop:
    vle8.v  v8, (a1)
    vle8.v  v9, (a2)
-   csrwi vxrm, 0    ; round up mode
+ csrrwi a5, vcsr, 0 ; save vxrm & vxsat. set round-up mode.
    vaadd.vv    v8, v8, v9
+  csrw vcsr, a5  ; restore vxrm & vxsat
    vse8.v  v8, (a0)
    sub a6, a6, a4
    add a1, a1, a4
    add a2, a2, a4
    add a0, a0, a4
    vsetvli a4, a6, e16, m1, ta, mu
    bnez    a4, loop
    ret
```
This PR is to save the compiler's effort to save/restore the vcsr state.

kito-cheng (changes requested):
I think that should be intrinsic spec issue and irreverent to psABI, you can define those fixed-point operation intrinsic functions with the changing `vxrm`/`vxsat` field in `vcsr`.

And I am try to prevent list `vxrm` and `vxsat` in the `Vector register convention` since that might give user wrong impression that is treat as caller-save rather than `thread local storage` semantics (or global variable like semantics).

arcbbb:
> I think that should be intrinsic spec issue and irreverent to psABI, you can define those fixed-point operation intrinsic functions with the changing `vxrm`/`vxsat` field in `vcsr`.
> 

My last example is a pure C code without using built-in intrinsic functions,
so it is beyond the scope of what intrinsic-doc can define.

Besides, the programming language can be C or other domain specific language, for example, Halide.
I think it is the reason to we specify calling convention in psABI.

kito-cheng:
Ideally compiler could hoist set/restore rounding mode before/after loop, so I would say that's compiler optimization issue rather than ABI issue.

```
f:
    li a6, 256
    vsetvli a4, a6, e16, m1, ta, mu
+ csrrwi a5, vcsr, 0 ; save vxrm & vxsat. set round-up mode.
loop:
    vle8.v  v8, (a1)
    vle8.v  v9, (a2)
    vaadd.vv    v8, v8, v9
    vse8.v  v8, (a0)
    sub a6, a6, a4
    add a1, a1, a4
    add a2, a2, a4
    add a0, a0, a4
    vsetvli a4, a6, e16, m1, ta, mu
    bnez    a4, loop
+  csrw vcsr, a5  ; restore vxrm & vxsat
    ret
```

kito-cheng:
ABI change isn't means we grant compiler permission to change the CSR _silently_, in that case, compiler will change `vxrm` and `vxsat` only *IF* vectorization successful, that mean user can't predict where `vxrm` and `vxsat` might be changed.

arcbbb:
> Ideally compiler could hoist set/restore rounding mode before/after loop, so I would say that's compiler optimization issue rather than ABI issue.

Right, compiler could hoist the rounding mode setting out of the loop.
But once psABI defines it is non-preserved across calls, the extra registers and instructions to backup the incoming value and restore the mode on return can be removed.

> ABI change isn't means we grant compiler permission to change the CSR _silently_, in that case, compiler will change `vxrm` and `vxsat` only _IF_ vectorization successful, that mean user can't predict where `vxrm` and `vxsat` might be changed.

Right, that's why we prefer `non-preserved across calls`.
Upon returning from a call, user don't assume `vxrm` state and need to set it again, as exactly how we deal with `vl`/`vtype`.

aswaterman (approved):
I've thought about this some more and am now leaning in @arcbbb's direction.  (Sorry, @kito-cheng :))

We give fflags and frm thread storage duration to efficiently implement C fenv.  Since there is no C fenv analogue for the fixed-point state, we shouldn't blindly copy the floating-point ABI strategy.

Making vxrm/vxsat call-clobbered would seem to optimize for the most common case: leaf functions containing vector loops.  If we instead used the frm/fflags strategy, we would have extra saves and restores that don't have a useful effect.

kito-cheng:
Yeah, I guess we should consider that, but one issue is we gonna to run release process next Monday, so that's kind of last minutes change...

@jrtc27 I don't like last minutes too :(, but I guess that would be harder to change once we release, so I proposal remove the sentence "The `vxrm` and `vxsat` fields of `vcsr` have thread storage duration." from the spec for now, and then add few more words after release?

aswaterman:
It would be better to miss a release than to bake in a short-sighed ABI decision.

jrtc27:
Whatever gets decided it needs to be done with the input of the LLVM and GNU developers actively working on vector support

kito-cheng:
LLVM vector folks: @rofirrim @fraser 
GNU vector folks: @zhongjuzhe @rjiejie

I would suggest remove all words about `vxrm` and `vxsat` from psABI spec at this release, and then we could having more time to discuss more complete/detailed solution, are you guys OK with that?

A +1 emoji or explicitly LGTM would be appreciate, of cause more comment/inputs are welcome!

aswaterman:
SGTM.

rofirrim:
Fine for me

frasercrmck:
SGTM too, assuming I'm _that_ Fraser :)

kito-cheng:
Thanks everyone, I created another PR to pull out the sentence :)

And I will keep this PR open, and welcome for more discussion on that.

kito-cheng:
@arcbbb did you mind write up a RFC and having more complete proposal and having comparison between different scheme? maybe include how interactive with vector intrinsic stuffs?

arcbbb:
I'll just stop here if no one is against the idea.

kito-cheng:
Hmmm, I would strongly suggest you should prepare a proposal, then we can have further discussion on psABI and gather boarder feedback.

arcbbb:
That's all I have for now. Thanks.

kito-cheng:
I am not sure we are on the same page yet, I just summary what we have now:

@nick-knight's opinion is that we should treat `vxrm`/`vxsat` by analogue with `frm`/`fflags` and having something like `venv` with defining extra API to controlling that like this proposal: https://github.com/riscv-non-isa/rvv-intrinsic-doc/pull/46.

@aswaterman's thought is we don't have C `fenv` analogue for the fixed-point state, so we might not rush do decide that yet.

@arcbbb is trying to propose ABI spec change, can let compiler modify  `vxrm` and `vxsat`, implicitly without restoring.

Am I understand/interpret that right?

My thought is more close to @nick-knight's opinion, defining `venv` and API to access that,
And furthermore, I would like to following the `fenv`convention described in C lang spec.

But I think part of this issue is more than psABI issue like Nick mention that might be describe in [C API doc](https://github.com/riscv-non-isa/riscv-c-api-doc) rather than psABI doc.

And pull it out for now is to preventing we are too rush to decide anything without complete consideration.

aswaterman:
My current position arose from a lengthy conversation with @nick-knight, so I assume he's roughly in agreement with me.

kasanovic:
We can't leave this issue undecided if we want a ratified standard vector calling convention.
I believe (and agree with) the consensus from other discussions (https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/287) was that these would be temporary state, not preserved across function calls.

kito-cheng:
@arcbbb and me will propose more complete proposal in next few days.

kito-cheng:
Moving to #294

PR #257, opened 2022-03-28
Add a note for ILP32E

pcwang-thead:


riscv-cc.adoc
```diff
@@ -260,6 +260,10 @@ boundary.  Registers x16-x31 do not participate in the calling convention, so
 there are only six argument registers, a0-a5, only two callee-saved registers,
 s0-s1, and only three temporaries, t0-t2.
 
+NOTE: As a consequence, stack alignment is 32-bit when passing arguments by stack
```
jrtc27, line 263: This is more than just non-normative text, this is an actual specification change (to match what GCC apparently does) regarding how register pairs are handled

pcwang-thead: I'm agree with you, but I think we should describe it in more details. And since ILP32E(GCC) has been used by some customers, we should be careful of big changes.

There are some discussions three years ago (https://github.com/riscv-collab/riscv-gcc/issues/148), we may:
* align stack to 8 bytes.
* or pass arguments with size of 2*XLEN differently and write it in specification document.

I am looking forward to your advice.

kito-cheng: I would prefer to make a ABI _fix_ on GCC, and add one more rule for require 16 byte stack alignment, but I expect that will happen after psABI 1.0 release.

That would be a ABI incomparable change, but we already put a disclaimer there to notice that's unstable...:P

pcwang-thead: May I ask why we need 16 bytes stack alignment? would 8 bytes be better if we need to make it an exception?

kito-cheng: 16 byte since the spec allow RV32EQ as valid combination, but I guess I should spend more time to make sure is 2*XLEN is enough or we should require 16 byte to make sure RV32EQ is also valid.

pcwang-thead: Thanks! I will follow up your fantastic works.

kito-cheng: That should be 2 * XLEN is enough since more than that will passed in stack.

kito-cheng:
I guess we should revise ilp32e part after release to fix those corner issue and F/D support issue on ilp32e.

EABI might going with divination-based, which means EABI will based on the base ABI with some variant instead of a brand new ABI for that.

So I think we should fix ilp32e issue, and that might break the compatibility of old ilp32e stuffs, that means we also need to figure out how to detect and report error for linking incompatible  ilp32e version.

riscv-cc.adoc
pcwang-thead: That makes sense to me.

pcwang-thead:
> 
Based on my experience with LLVM implementation, there is only one issue: the alignment of types (C/C++ types or aggregates ) with size in (XLEN, 2*XLEN]. That results in issues on stack alignment and handling of variadic arguments. For your reference, I think there are two solutions:
* Force them to align to 4 bytes, which is counterintuitive. This may increase the usage of registers and save some stack spaces. This is what GCC is doing.
* Force them to align to 8 bytes, which is intuitive but against common stack alignment of 4 bytes. some registers may not be used during passing variadic arguments and stack will be larger. This is LLVM‘s default behavior if we don't change any things.

I don't know if there is any asm directive or attribute that shows the ABI we used, so it may be difficult to detect ilp32e version.

PR #259, opened 2022-03-31
Remove description of vxrm and vxsat from the spec

kito-cheng:
Pull out the `vxrm` and `vxsat` for this release and having a more complete discussion after 1.0 release.

Ref: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/256


jrtc27 (approved):

kito-cheng:
This give us more time to discuss for the vector ABI design.

PR #260, opened 2022-04-02
riscv-elf: Rework specifying when STO_RISCV_VARIANT_CC is needed

jrtc27:
This new text both captures using registers with a specific purpose in
the ABI (like SP) and also avoids creating incompatibilities with
glibc's LD_AUDIT by being more general, as discussed on libc-alpha.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -112,9 +112,8 @@ table (GOT) for non-local symbol addresses.
 
 == Dynamic Linking
 
-Run-time linkers that use lazy binding must preserve all argument registers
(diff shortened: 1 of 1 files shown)


jrtc27:
@palmer-dabbelt @fweimer-rh

kito-cheng:
LGTM, but let wait Palmer ack, I've ping him in another place :)

kito-cheng (approved):

kito-cheng:
We gonna to merge this for moving to public review, I think we can revise the word if any feedback during public review.

PR #262, opened 2022-04-14
Add state statement in preamble

kito-cheng:
I guess the RVI's state statement might well fit for psABI, especially
we are open to accept new changes after 1.0 ratified, but I think we
could revise the state statement for ratified later.

### DIFF
--- a/preamble.adoc
+++ b/preamble.adoc
@@ -1,3 +1,36 @@
+
+ifeval::["{revlifecycle}" == "dev"]
+[WARNING]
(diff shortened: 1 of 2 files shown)


kito-cheng:
Changes:
- Remove local definition of `development`, `stable`, `frozen` and `ratified`.

cmuellner:
I think this change is fine.
It actually does not really change anything, but streamlines definitions and links to the original source of the definitions.

Issue #273, opened 2022-04-21
Clean up for Table 13. TLS models


kito-cheng:
Review comment come from @anderslindgren-iar and Anders Berg(IAR) https://lists.riscv.org/g/tech-toolchain-runtime/message/344

| Mnemonic | Model | Compiler flags |
|----------|:-------------|:------|
| TLS LE | Local Exec | -ftls-model=local-exec |
| TLS IE | Initial Exec | -ftls-model=initial-exec |
| TLS LD | Local Dynamic | -ftls-model=local-dynamic |
| TLS GD | Global Dynamic | -ftls-model=global-dynamic |

Anders: The compiler flags are toolchain specific and may be different depending on which toolchain one is using.

nick-knight:
I would support moving the flags to https://github.com/riscv-non-isa/riscv-toolchain-conventions

MaskRay:
"general dynamic" is the more commonly used expansion of GD. "global dynamic" occurs very few times in Drepper's TLS ABI and the name is unfortunately used by GCC.

kito-cheng:
@MaskRay added a note, that should...make this few more better.

PR #275, opened 2022-04-25
Add note to mention `Global Dynamic` model is known as `General Dynamic` model

kito-cheng:
Address @MaskRay's comment: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/273#issuecomment-1104698498

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -711,6 +711,8 @@ thread local storage models:
 The program linker in the case of static TLS or the dynamic linker in the case
 of dynamic TLS allocate TLS offsets for storage of thread local variables.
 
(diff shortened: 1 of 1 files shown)


riscv-elf.adoc
```diff
@@ -711,6 +711,8 @@ thread local storage models:
 The program linker in the case of static TLS or the dynamic linker in the case
 of dynamic TLS allocate TLS offsets for storage of thread local variables.
 
+NOTE: `Global Dynamic` model is known as `General Dynamic` model.
+
```
luismarques, line 715: "is known as" -> "is also known as"

kito-cheng: Thanks :)

PR #277, opened 2022-04-25
Add forward reference to GOT in Medium position independent code model section.

### LINKED ISSUE #271: Define GOT before use

> global offset table (GOT)

Anders: Where is the specification for GOT (how can one find it)?

```
    # Load value from a local symbol 
0:  auipc a0, %pcrel_hi(symbol)
    lw    a0, %pcrel_lo(0b)(a0)
    # Store value to a local symbol 
1:  auipc a0, %pcrel_hi(symbol)
    sw    a1, %pcrel_lo(1b)(a0)
    # Calculate address of a local symbol 
2:  auipc a0, %pcrel_hi(symbol)
    addi  a0, a0, %pcrel_lo(2b)
    # Calculate address of non-local symbol 
3:  auipc  a0, %got_pcrel_hi(symbol)
    l[w|d] a0, a0, %pcrel_lo(3b)
```

Anders: The example is pointless if one cannot find the specification for GOT.
Should it not be an additional case of how to read a non-local symbol?

kito-cheng:
Fix #271

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -103,8 +103,8 @@ for that, please use the medium position independent code model below.
 
 === Medium position independent code model
 
-This model is similar to the medium code model, but uses the global offset
-table (GOT) for non-local symbol addresses.
+This model is similar to the medium code model, but uses the
+<<Global Offset Table,global offset table>> (GOT) for non-local symbol addresses.
 
 [,asm]
 ----


riscv-elf.adoc
```diff
@@ -103,8 +103,8 @@ for that, please use the medium position independent code model below.
 
 === Medium position independent code model
 
-This model is similar to the medium code model, but uses the global offset
-table (GOT) for non-local symbol addresses.
+This model is similar to the medium code model, but uses the
+global offset table(GOT, see also <<Global Offset Table>>) for non-local symbol addresses.
```
luismarques, line 107: It doesn't really matter but, personally, I would just add the link to the existing text. Something like:

"This model is similar to the medium code model, but uses the <<Global Offset Table,global offset table>> (GOT) for non-local symbol addresses."

kito-cheng: I tried that but PDF will render that as section number like "...but uses the  Section 8.5.5(GOT) for non-local..." :(

MaskRay: The convention is to ensure a space between `(` and the word before it.

kito-cheng:
Changes:
- Rebase
- Address both comments

PR #278, opened 2022-05-06
Do not using backward numeric labels reference

### LINKED ISSUE #270: Don't use GCC-specific syntax for example code on psABI

```
    # Load value from a symbol
0:  auipc a0, %pcrel_hi(symbol)
    lw    a0, %pcrel_lo(0b)(a0)
    # Store value to a symbol
1:  auipc a0, %pcrel_hi(symbol)
    sw    a1, %pcrel_lo(1b)(a0)
    # Calculate address
2:  auipc a0, %pcrel_hi(symbol)
    addi  a0, a0, %pcrel_lo(2b)
```

Anders: The example is GCC-specific. An example should preferably be tool chain neutral or at least it should be mentioned that it is GCC-specific and that the syntax may differ a bit depending on which toolchain one is using.

kito-cheng:
Fix #270.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -84,17 +84,17 @@ an address in the medany code model.
 
 [,asm]
 ----
-    # Load value from a symbol
-0:  auipc a0, %pcrel_hi(symbol)
-    lw    a0, %pcrel_lo(0b)(a0)
+         # Load value from a symbol
+.Ltmp0:  auipc a0, %pcrel_hi(symbol)
+         lw    a0, %pcrel_lo(.Ltmp0)(a0)
 
-    # Store value to a symbol
-1:  auipc a0, %pcrel_hi(symbol)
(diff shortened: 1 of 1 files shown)


aswaterman (commented):
FTR, I do not support this change.  I don't agree that #270 is actually a problem that needs fixing.

(But I won't block this PR if others feel strongly.)

aswaterman:
Yeah, I guess that's also how I feel: this individual change is fine, but I don't want to set the precedent that we can't use GNU syntax based upon the dubious argument in #270.

kito-cheng:
Changes:
- Re-indent

kito-cheng:
Hopefully we can have some volunteers to improve the poor asm manual, and document that as part of spec, then we can claim that's RISC-V asm syntax rather than GNU sytax in future :(

kito-cheng:
Gonna to merge this, no strong objection for months.

PR #281, opened 2022-05-06
Explicitly say upper bit in GPR is undefined when passing floating po…

### LINKED ISSUE #267: More clarification for passing floating point value on GPR

> without extending the integer to XLEN bits,

Anders: This may need an additional clarification. If the higher bits are regarded as undefined, it should be clear it is so.

kito-cheng:
…int value in hard fp CC.

Fix #267.

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -234,10 +234,12 @@ floating-point reals.
 
 A struct containing one floating-point real and one integer (or bitfield), in
 either order, is passed in a floating-point register and an integer register,
-without extending the integer to XLEN bits, provided the floating-point real
-is no more than ABI_FLEN bits wide and the integer is no more than XLEN bits wide,
-and at least one floating-point argument register and at least one integer
-argument register is available. Otherwise, it is passed according to the
+provided the floating-point real is no more than ABI_FLEN bits wide and the
+integer is no more than XLEN bits wide, and at least one floating-point
+argument register and at least one integer argument register is available.
+If the struct is passed in this manner, and the integer is narrower than XLEN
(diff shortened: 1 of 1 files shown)


aswaterman (commented):
This sentence is already very long.  May I suggest breaking it up?  Here's my attempt:

A struct containing one floating-point real and one integer (or bitfield), in
either order, is passed in a floating-point register and an integer register,
provided the floating-point real is no more than FLEN bits wide and the
integer is no more than XLEN bits wide, and at least one floating-point
argument register and at least one integer argument register is available.
If the struct is passed in this manner, and the integer is narrower than XLEN
bits, the remaining bits are unspecified.
If the struct is not passed in this manner, then it is passed according to the
integer calling convention.

jrtc27:
> This sentence is already very long. May I suggest breaking it up? Here's my attempt:
> 
> A struct containing one floating-point real and one integer (or bitfield), in either order, is passed in a floating-point register and an integer register, provided the floating-point real is no more than FLEN bits wide and the integer is no more than XLEN bits wide, and at least one floating-point argument register and at least one integer argument register is available. If the struct is passed in this manner, and the integer is narrower than XLEN bits, the remaining bits are unspecified. If the struct is not passed in this manner, then it is passed according to the integer calling convention.

This minor reorganising works for me (and also addresses the grammatical issue)

asb:
Andrew's proposed rewording LGTM.

kito-cheng:
Changes:
- Apply @aswaterman's suggestion

kito-cheng:
Rebase and merge :)

PR #284, opened 2022-05-27
introduction: Add note about memory model synchronization primitives

cmuellner:
It was suggested to cover the memory model synchronization primitives
as part of the psABI documentation. Let's add a sentence to the
introduction that announces that for future revisions.

CC: @kito-cheng


kito-cheng (approved):
LGTM, we'll add that in future as we discussed in https://lists.riscv.org/g/tech-psabi/topic/90910220,

PR #286, opened 2022-05-31
Public review comments

palmer-dabbelt:
These are really just a bunch of independent comments, but I figured it'd be easier to have them all in one PR.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -933,7 +933,7 @@ value if the tag number is even.
 
 | Tag_RISCV_stack_align               |        4 | uleb128        | Indicates the stack alignment requirement in bytes.
 | Tag_RISCV_arch                      |        5 | NTBS           | Indicates the target architecture of this object.
-| Tag_RISCV_unaligned_access          |        6 | uleb128        | Indicates whether to impose unaligned memory accesses in code generation.
+| Tag_RISCV_unaligned_access          |        6 | uleb128        | Indicates nothing of interest.
 | Tag_RISCV_priv_spec                 |        8 | uleb128        | Indicates the major version of the privileged specification.
 | Tag_RISCV_priv_spec_minor           |       10 | uleb128        | Indicates the minor version of the privileged specification.
 | Tag_RISCV_priv_spec_revision        |       12 | uleb128        | Indicates the revision version of the privileged specification.


palmer-dabbelt:
> For `Don't require the GOT for maybe-undefined-weak symbols`, FWIW using GOT matches aarch64/powerpc64/x86-32/x86-64.

I haven't read the specs, but GCC will emit direct references for these -- unless I'm missing something about how GOT works on x86?

For example

````
$ cat udw.c
#include <stdint.h>

extern int udw __attribute__((weak));
intptr_t udw_ptr(void) { return (intptr_t)(&udw); }
$ gcc udw.c -c --no-pie
$ objdump -dr udw.o
...
0000000000000000 <udw_ptr>:
   0:	55                   	push   %rbp
   1:	48 89 e5             	mov    %rsp,%rbp
   4:	b8 00 00 00 00       	mov    $0x0,%eax
			5: R_X86_64_32	udw
   9:	5d                   	pop    %rbp
   a:	c3                   	ret    
...
````

That all links and executes, at least on my system.  It is the Gentoo system compiler, though, which tends to be a bit wacky...

**Comment by kito-cheng** [comment]
Discussion about GOT for medany weak undef before: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/201

**Comment by jrtc27** [comment]
GCC does that, and LLVM currently does as it was originally made to match GCC, but that requires linker relaxations to be able to implement at all. Using the GOT requires no relaxations for correctness, but the linker can still relax the sequence to what GCC wants to use if it so desires. This is why there is the following note:

> This is not yet a requirement as existing toolchains predating this part of the specification do not adhere to this, and without improvements to linker relaxation support doing so would regress performance and code size.

**Comment by palmer-dabbelt** [comment]
> GCC does that, and LLVM currently does as it was originally made to match GCC, but that requires linker relaxations to be able to implement at all. Using the GOT requires no relaxations for correctness, but the linker can still relax the sequence to what GCC wants to use if it so desires. This is why there is the following note:
> 
> > This is not yet a requirement as existing toolchains predating this part of the specification do not adhere to this, and without improvements to linker relaxation support doing so would regress performance and code size.

I guess that really depends on the definition of linker relaxation, as this doesn't require deleting bytes it just requires rewriting non-immediate bits in the input instruction (ie, to `addi R, x0, 0` or `lw R, 0(x0)`).  If you want to deprecate the direct accesses because they rewrite the source register that's totally fine with me, my argument is simply against saying it's GOT-indirect as the GOT comes with other constraints -- even the existence of a GOT can be a headache for some targets.

The reason I posed that GCC output was to demonstrate that, on x86, GCC will still emit non-GOT-indirect references to possibly-undefined weak symbols.  If there's some spec saying that x86 systems have GOT-indirect references to possibly-undefined weak symbols and GCC's just ignoring it then that's also fine with me, as that should be enough precedent for us to ignore it when a GOT is a headache (and maybe to relax away the indirection, though I'm way less worried about that).

File: riscv-elf.adoc [hunk 1]
```diff
@@ -18,14 +18,14 @@ However, some code models can't access the whole address space. The linker may
...
-The small code model, or `medlow`, allows the code to address the whole RV32
-address space or the lower 2 GiB and highest 2 GiB of the RV64 address space
-(`0xFFFFFFFF7FFFF800` ~ `0xFFFFFFFFFFFFFFFF` and `0x0` ~ `0x000000007FFFF7FF`).
-By using the instructions `lui` and `ld` or `st`, when referring to an object, or
-`addi`, when calculating an address literal, for example,
-a 32-bit address literal can be produced.
+The medium low code model, or `medlow`, allows the code to address the whole RV32
+address space or any 2 GiB region within the lower 2 GiB and highest 2 GiB of
+the RV64 address space (`0xFFFFFFFF7FFFF800` ~ `0xFFFFFFFFFFFFFFFF` and `0x0` ~
+`0x000000007FFFF7FF`).  By using the instructions `lui` and `ld` or `st`, when
```
nick-knight, line 26: I'm probably misunderstanding your intention here, but I don't think your change improves the text.

My understanding is we can express $2^{32}$ addresses. In RV32, this is the whole address space. In RV64, we can express the addresses 

* in the intervals $[0,2^{31}-2^{11}-1]$ and $[2^{64} - 2^{31} - 2^{11}, 2^{64}-1]$, interpreting addresses as unsigned integers, and
* in the interval $[-2^{31} - 2^{11}, 2^{31}-2^{11}-1]$, interpreting addresses as two's complement integers.

If there's a problem in this paragraph, then it's in the phase "lower 2 GiB and highest 2 GiB", so I would suggest just removing that:

```suggestion
address space or a subset of the RV64 address space, 
`0xFFFFFFFF7FFFF800` ~ `0xFFFFFFFFFFFFFFFF` and `0x0` ~ `0x000000007FFFF7FF`.
By using the instructions `lui` and `ld` or `st`, when
```

nick-knight (changes requested):

riscv-elf.adoc
palmer-dabbelt: There's two constraints on medlow addressing:

* The "data" relocations, ie load/store/add (R_RISCV_HI20 and R_RISCV_LO12_{I,S}) are lui-based, which need to be in the 4GiB region around absolute 0 (signed extended).  That's what the original text matches.
* Calls (ie, R_RISCV_CALL) are PC-relative and have essentially the same constraints as medany (it's exactly the same relocation), which is a 4GiB region around the current PC.

So to ensure that it's possible to link a program, addresses must all lie within a single 2GiB range (the same as medany), but rather than that range being anywhere in the address space (as medany has all relocations as PC-relative) that range must be in the lower 4GiB (sign extended).

That, or we do a similar auipc->lui conversion for calls that's just being deprecated for load/store/add -- I'd do the absolute conversions for both for all accesses (thus making medany a superset of medlow), but that's not the way the spec looks like it's going.

nick-knight: Thanks Palmer. This isn't the first time you've explained this to me, and it's probably not the last either.

MaskRay:
> > GCC does that, and LLVM currently does as it was originally made to match GCC, but that requires linker relaxations to be able to implement at all. Using the GOT requires no relaxations for correctness, but the linker can still relax the sequence to what GCC wants to use if it so desires. This is why there is the following note:
> > > This is not yet a requirement as existing toolchains predating this part of the specification do not adhere to this, and without improvements to linker relaxation support doing so would regress performance and code size.
> 
> I guess that really depends on the definition of linker relaxation, as this doesn't require deleting bytes it just requires rewriting non-immediate bits in the input instruction (ie, to `addi R, x0, 0` or `lw R, 0(x0)`). If you want to deprecate the direct accesses because they rewrite the source register that's totally fine with me, my argument is simply against saying it's GOT-indirect as the GOT comes with other constraints -- even the existence of a GOT can be a headache for some targets.

Using GOT requires that the distance between the code and the GOT entry be materializable.
Using a direct access requires that the code can be converted to load a zero, in case the undefined weak symbol is unresolved.

The latter is usually a stricter requirement.

> The reason I posed that GCC output was to demonstrate that, on x86, GCC will still emit non-GOT-indirect references to possibly-undefined weak symbols. If there's some spec saying that x86 systems have GOT-indirect references to possibly-undefined weak symbols and GCC's just ignoring it then that's also fine with me, as that should be enough precedent for us to ignore it when a GOT is a headache (and maybe to relax away the indirection, though I'm way less worried about that).

The current wording mentions "for position-independent code". GCC uses GOT-indirect addressing for position-independent code.

riscv-cc.adoc
```diff
@@ -59,7 +59,7 @@ duration in accordance with C11 section 7.6 "Floating-point environment
 .Vector register convention
 [%autowidth]
 |===
-| Name    | ABI Mnemonic | Meaning                      | Preserved across calls?
+| Name    | ABI Mnemonic | Meaning                      | Preserved into calls?
```
kito-cheng, line 62: I would suggest keep using `Preserved across calls?` here, I guess the intention you want change that into `into` is because that reflect the dynamic linking/`STO_RISCV_VARIANT_CC` stuffs, however just like we are using `t0`/`t1` in PLT stubs and all caller-save register other than argument register might clobber at PLT resolver, but we still using `Preserved across calls?` to describe that, so I think we don't need to change that.

So we could add NOTE to mentions that: `all register not marked as Argument registers might not preserved into calls`, that also describe the thing right for GPR and FPR situation, but the word might not 100% correct for this situation: hard float with soft float ABI like `-march=rv64gc` with `-mabi=lp64` - anyway I think we can find better wording about that later.

riscv-elf.adoc
```diff
@@ -18,14 +18,14 @@ However, some code models can't access the whole address space. The linker may
 raise an error if it cannot adjust the instructions to access the target address
 in the current code model.
 
-=== Small code model
+=== Medium low code model
 
-The small code model, or `medlow`, allows the code to address the whole RV32
```
kito-cheng, line 23: Small code model are used in the clang/LLVM, but I think it's ok to re-title that to `Medium low` and tweak that little bit.

kito-cheng:
Cherry-pick https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/286/commits/f7fe921c188a5b2c65b479862344dc682f516311 to https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/262

riscv-cc.adoc
kito-cheng: Address by https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/290

kito-cheng:
Close due to this has break into few other PR and (mostly) address separately.

PR #288, opened 2022-06-04
Clarify implications for fcsr

nick-knight:
My understanding is that the psABI intends for `fcsr` to be the ABI manifestation of `fenv`, when the latter is relevant. With this understanding, I find the current wording to be misleading:

* `fcsr` is an object in the ISA, not the C language, so it is nonsensical to say it has "thread storage duration".
* It can be interpreted as imposing constraints on `fcsr` for all programs, even those that don't access/modify `fenv`.
* The relevant properties of `fenv` have been present since C99, so the specific reference to C11 is not meaningful.

This PR attempts to clarify the intent while addressing these three points. Your feedback is most welcome.

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -50,9 +50,10 @@ calls if they are no larger than the width of a floating-point register in the
 targeted ABI. Therefore, these registers can always be considered temporaries
 if targeting the base integer calling convention.
 
-The Floating-Point Control and Status Register (fcsr) must have thread storage
-duration in accordance with C11 section 7.6 "Floating-point environment
-<fenv.h>".
+In contexts where accesses or modifications to the C floating-point environment 
+(`fenv`) are meaningful, the state of `fenv` is coherent with that of the RISC-V 
+floating-point control and status register (`fcsr`). In these contexts, `fcsr` can
+be thought of as having thread storage duration.
 
 === Vector Register Convention
 
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -51,9 +51,10 @@ targeted ABI. Therefore, these registers can always be considered temporaries
 if targeting the base integer calling convention.
 
 In contexts where accesses or modifications to the C floating-point environment 
-(`fenv`) are meaningful, the state of `fenv` is coherent with that of the RISC-V 
-floating-point control and status register (`fcsr`). In these contexts, `fcsr` can
-be thought of as having thread storage duration.
+(`fenv`) are meaningful, like when the `FENV_ACCESS` pragma is "on", the state of
+`fenv` is coherent with that of the RISC-V floating-point control and status 
+register (`fcsr`). In these contexts, `fcsr` can be thought of as having thread 
+storage duration.
 
 === Vector Register Convention
 


riscv-cc.adoc
```diff
@@ -50,9 +50,10 @@ calls if they are no larger than the width of a floating-point register in the
 targeted ABI. Therefore, these registers can always be considered temporaries
 if targeting the base integer calling convention.
 
-The Floating-Point Control and Status Register (fcsr) must have thread storage
-duration in accordance with C11 section 7.6 "Floating-point environment
-<fenv.h>".
+In contexts where accesses or modifications to the C floating-point environment 
+(`fenv`) are meaningful, the state of `fenv` is coherent with that of the RISC-V
```
aswaterman, line 54: "meaningful with respect to the setting of FENV_ACCESS" perhaps?  I don't think this changes the normative meaning, but it might make the intent clearer.

nick-knight: Agreed. I was hedging because it seems that some implementations allow other means of meaningful access besides `FENV_ACCESS`. But I think it helps to mention this (standard) means.

kito-cheng:
> fcsr is an object in the ISA, not the C language, so it is nonsensical to say it has "thread storage duration".

I am struggling with `thread storage duration` since that also give promise about the CSR is thread-safe and execution environment must guarantee that, but I think that different point from your PR, however that would be a wider issue that just `fcsr`, there are same issue for `vcsr`, `vl` and `vtype` and all URW unprivileged CSR.

> It can be interpreted as imposing constraints on fcsr for all programs, even those that don't access/modify fenv.

Just make sure you are trying to address this by following sentence ?
```
In contexts where accesses or modifications to the C floating-point environment 
(`fenv`) are meaningful, like when the `FENV_ACCESS` pragma is "on",
```

> The relevant properties of fenv have been present since C99, so the specific reference to C11 is not meaningful.

Agree with that, thanks!

kito-cheng:
Another attempt to clarify fcsr: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/289

nick-knight:
> the CSR is thread-safe and execution environment must guarantee that

My understanding is that the ISA guarantees that each hardware thread (hart) will have its own `fcsr`, so `fcsr` is "thread-safe" in this sense, and there is nothing further the ABI needs to say about it. Perhaps you mean C-language threads, not RISC-V harts? In this case, we need to first clearly define how `fcsr` is reflected by an object in the C language. My PR attempts to address this.

> Just make sure you are trying to address this by following sentence ?

Correct.

nick-knight:
As far as I can tell, neither Clang nor GCC fully supports `fenv` for RISC-V targets, even though it's been ~23 years. The `FENV_ACCESS` pragma is ignored, and the various command-line flags don't fully cover the required functionality.

My attitude is that my use-case would be better addressed by F-extension intrinsics. I acknowledge that this is a RISC-V-specific solution; however, in my experience, the people asking for this functionality generally aren't writing code that is meant to be broadly portable across architectures.

Another one of my objectives with this PR is to leave the psABI open to eventual implementation of `fenv`, as well as alternatives like F-extension intrinsics.

kito-cheng:
> As far as I can tell, neither Clang nor GCC fully supports fenv for RISC-V targets, even though it's been ~23 years. The FENV_ACCESS pragma is ignored, and the various command-line flags don't fully cover the required functionality.

Yeah, that's one reason I didn't mention FENV_ACCESS in https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/289 since I am not sure should we mention a thing without well implemented in mainstream open source compilers...

> My attitude is that my use-case would be better addressed by F-extension intrinsics. I acknowledge that this is a RISC-V-specific solution; however, in my experience, the people asking for this functionality generally aren't writing code that is meant to be broadly portable across architectures.

Might be little bit digress from the topic, but I am curious what the shape is for those intrinsic in your mind? providing static rounding mode argument? or guarantee the order among different F-ext. intrinsic?

kito-cheng:
Let close since @nick-knight is not working on RISC-V longer :(

nick-knight:
It’s fine to close it, Kito. IIRC, the toolchains that I care about seemed to do the right thing already; this PR was mainly an attempt to clarify what this means.

PR #289, opened 2022-06-07
Rewrite description of FCSR.

kito-cheng:
Try to rewrite description of FCSR, remove "thread storage duration", and reference of specific language standard, and also try to describe properties for different fields.

Description of `frm` are borrowed some word from the C lang spec[1].
And `fflags` might be changed by floating point operations or any function call, and no any promise of the value on the function entry.
Description of other bits are reference unpriv spec. 

[1] ISO/IEC 9899:2011: 7.6.3
```
— a function call does not alter its caller’s ﬂoating-point control modes, clear its caller’s
ﬂoating-point status ﬂags, nor depend on the state of its caller’s ﬂoating-point status
ﬂags unless the function is so documented;
```

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -50,9 +50,15 @@ calls if they are no larger than the width of a floating-point register in the
 targeted ABI. Therefore, these registers can always be considered temporaries
 if targeting the base integer calling convention.
 
-The Floating-Point Control and Status Register (fcsr) must have thread storage
-duration in accordance with C11 section 7.6 "Floating-point environment
-<fenv.h>".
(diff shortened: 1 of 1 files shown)


kito-cheng:
@nick-knight I don't know why I can add @aswaterman to reviewer but can't add you, so tag you manually :P

nick-knight:
@kito-cheng: I don't have privilege to review PRs in `riscv-non-isa` repos like this one. An admin will have to extend my privileges. (I would argue that this privilege should be a default, but I don't fully understand the ramifications.)

PR #290, opened 2022-06-14
Describe content of non-argument/non-return register are unspecified …

kito-cheng:
…upon entry/exit.

Address @palmer-dabbelt's comment for vector register convention: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/286/commits/eefd8b0e5a963391e63a15b65e41ee2945c80534

Explicitly say those content is unspecified upon entry/exit, we only say those value will preserved across call or not.

In practice, those content of those register might be corrupt by PLT resolver (lazy binding), glibc's auditing feature or something else.

I tried describe that carfully to make sure that is still correct when we define another calling convention variant in future.

[1] https://man7.org/linux/man-pages/man7/rtld-audit.7.html

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -85,6 +85,18 @@ Any procedure that does explicitly write `vstart` to a nonzero value must zero
 
 == Procedure Calling Convention
 
+This chapter defines standard calling conventions, and describes how to pass
+parameters and return values.
+
+Functions must follow the register convention defined in calling convention: the
+contents of any register without specifying it as an argument register
+in the calling convention are unspecified upon entry, and the content of any
+register without specifying it as a return value register or callee-saved in
+the calling convention are unspecified upon exit, the contents of all
+callee-saved registers must be restored to what was set on entry, and the
+contents of any fixed registers like `gp` and `tp` never change,
+
+
 NOTE: Calling convention for big-endian is *NOT* included in this specification
 yet, we intend to define that in future version of this specification.
 


riscv-cc.adoc
```diff
@@ -80,6 +80,15 @@ Any procedure that does explicitly write `vstart` to a nonzero value must zero
 
 == Procedure Calling Convention
 
+This chapter define standard calling conventions, and describe how to pass
+parameters and return values.
+
+Function should follow the register convention defined in calling
```
palmer-dabbelt, line 86: "should" is kind of meaningless

riscv-cc.adoc
```diff
@@ -80,6 +80,15 @@ Any procedure that does explicitly write `vstart` to a nonzero value must zero
...
+convention, content of any register without specifying as argument register
+in the calling convention are unspecified upon entry, and content of any
+register without specifying as return value register in the calling convention
+are unspecified upon exit.
```
palmer-dabbelt, line 90: This still has the bug: variant CC functions can use calling conventions that define any V register as an argument register, thus their values must be preserved during dynamic resolution.

palmer-dabbelt (commented):
Hey, sorry, I missed this one.

riscv-cc.adoc
palmer-dabbelt: oh, sorry, that's nonsense.  Hang on, I'm super out of it...

palmer-dabbelt (commented):
Sorry about that, I think this is good, though the fixed registers should be in there too.  Clearly I'm super out of it right now, though, so maybe I'm wrong...

kito-cheng:
Changes:
- Apply @palmer-dabbelt's suggestion.
- Add description about fixed register like `gp` and `tp`.

kito-cheng:
@davidlt thanks, fixed :P

riscv-cc.adoc
```diff
@@ -80,6 +80,15 @@ Any procedure that does explicitly write `vstart` to a nonzero value must zero
 
 == Procedure Calling Convention
 
+This chapter define standard calling conventions, and describe how to pass
+parameters and return values.
+
+Function should follow the register convention defined in calling
+convention, content of any register without specifying as argument register
```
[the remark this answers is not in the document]
kito-cheng: Good catch, according to this description fixed register is unspecified upon entry and exit :P

kito-cheng:
Changes:
- Minor tweak fixed register part.

asb:
Maybe I'm just failing to parse this correctly, but I worry that the description of any non-return registers as "undefined" upon exit is confusing (as obviously for callee-saved registers, they should be set to whatever they were set to upon entry).

aswaterman:
Agreed with @asb that the text as written implies that callee-saved registers are undefined on return, which is not the case.

kito-cheng:
Changes:
- Rebase
- Add description to callee-save must be register must restored to what was set on entry.

riscv-cc.adoc
```diff
@@ -85,6 +85,18 @@ Any procedure that does explicitly write `vstart` to a nonzero value must zero
 
 == Procedure Calling Convention
 
+This chapter define standard calling conventions, and describe how to pass
```
aswaterman, line 88: define => defines, describe => describes

riscv-cc.adoc
```diff
@@ -85,6 +85,18 @@ Any procedure that does explicitly write `vstart` to a nonzero value must zero
...
+parameters and return values.
+
+Function must follow the register convention defined in calling convention, the
```
aswaterman, line 91: Function => Functions

“, the” => “: the”

aswaterman (approved):
WFM. I made some minor grammatical suggestions.

kito-cheng:
Changes:
- Apply @aswaterman's suggestion.

kito-cheng:
@aswaterman thanks!!

kito-cheng:
Wait few more day to merge :)

PR #291, opened 2022-06-14
Correct name for medany and medlow

kito-cheng:
Cherry-pick from #286, but add one more note to mention `medlow` also called `small` code model in LLVM.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -18,9 +18,9 @@ However, some code models can't access the whole address space. The linker may
 raise an error if it cannot adjust the instructions to access the target address
 in the current code model.
 
-=== Small code model
+=== Medium low code model
 
-The small code model, or `medlow`, allows the code to address the whole RV32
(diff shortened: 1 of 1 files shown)


palmer-dabbelt:
LGTM

riscv-elf.adoc
```diff
@@ -59,9 +59,12 @@ lui a0, 0x80000 # a0 = 0xffffffff80000000
 addi a0, a0, -0x800 # a0 = a0 + -2048 = 0xFFFFFFFF7FFFF800
 ----
 
-=== Medium code model
+NOTE: `medlow` also called small code model in certain toolchain implementation
```
jrtc27, line 62: Not sure this belongs here... Clang accepts -mcmodel=medlow/medany and then remaps to small/medium internally. It's up to LLVM to document how it implements the psABI, not for us to document toolchain quirks in a standards document. Also if even after that we are going to document this then we should be consistent and also document that medany is called medium.

Issue #287, opened 2022-06-02
vxrm / vxsat vector state missing from vector register calling convention


kasanovic:
I assume these will be temporary state.

aswaterman:
Suggested edit is adding `vcsr` to Table 3, indicating it's not preserved across calls.

kito-cheng:
`vxrm` and `vxsat` was listed in the spec before but we decide to pull that out for now, see https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/256

Currently we have two option for `vxrm` and `vxsat`
1. Explicitly say that is caller-save.
2. Define that as a global state like `fcsr` and let user to control that like a global variable.

And the staged conclusion is don't rush to decide that yet, and all vector stuffs will be high priority task after psABI 1.0 spec released.

kasanovic:
I don’t think it makes sense to have only a part of the vector state described in the spec - either all or nothing.
But I can’t see how compilers can use the vector unit without having the calling convention established, so it seems pretty important to define this.

Reading through the thread, it seems the discussion was leaning towards 1).
Given that there is fixed-point equivalent to  fenv, then 1) makes the most sense as being the most efficient.

Looking ahead, I seriously doubt that a global rounding mode state makes sense for fixed-point code.
A global saturation flag could be desired, but alternatively an explicit additional saturation flag argument to selected routines might be more modern approach to providing this functionality.

I’d vote for going ahead with 1) just so we don’t have delay in compiling for vector code.

Krste

aswaterman:
I agree that having a partial vector calling convention is probably worse
than having all or nothing. And the right technical answer does seem to be
making it caller-saved. I suggest adopting this right now (and if you
reject that proposal, please delete the entire vector calling convention).

aswaterman:
I'll state more explicitly that I think this is a blocker for ratification.  A partial calling convention is the worst of both worlds: it creates constraints that can't be changed in a backwards-compatible manner, without providing sufficient guarantees to write software.

kasanovic:
I agree with @aswaterman that this needs to be resolved for ratification - we can't have a half-finished calling convention.  It seems there is consensus on treating these as temporary state.

kito-cheng:
Fixed via #294

PR #294, opened 2022-06-20
Define vxrm/vxsat is not preserved across calls and unspecified upon entry field upon entry

kito-cheng:
After discussion in #287, we realized that keeping the convention of `vxsat` and `vxrm` to be unspecified made the vector extension become unusable on the software side, because that might become an ABI breakage once we define that later, so we would like to define that within psABI 1.0 release.

Here is two potential proposals for `vxrm` and `vxsat` in #256:
- Same definition as `fcsr`, and followed the same convention as fenv.
- Define as `"not preserved across calls and unspecified upon entry"`.

Based on the discussion in #256 and few off list discussions, we have better understanding of the usage for `vxrm` and `vxsat` registers now, so we propose take the second proposal: “not preserved across calls and the value is unspecified upon entry” as the ABI for the `vxrm` and `vxsat` according following reasons:

- Using same policy as floating point environment might cause unnecessary overhead for maintain the convention described in C language specification : `"a function call does not alter its caller's floating-point control modes, clear its caller’s ﬂoating-point status ﬂags, nor depend on the state of its caller’s floating-point status ﬂags unless the function is so documented"`[1]; all existing libraries are *NOT* documented as might change fixed point rounding mode, that means we must backup and restore the content of `vxrm` and `vxsat` at those library functions when we implement those functions with vector fixed-point operations to ensure we didn't violate the convention.

- Another key difference between fixed-point rounding mode and floating point environment is fixed-point didn't have it own default rounding mode, and define one isn't meaningful because fixed-point rounding modes are likely to be setting to specific mode for most fixed-point algorithm, and might be changed frequently within a single algorithm implementation[2] - unlike floating point rounding modes are unlikely to change in most situation.

Based on the above reason, we believe that is the best choice for `vxrm` and `vxsat`.

[1] ISO/IEC 9899:2011: 7.6.3
[2] https://github.com/riscv/riscv-v-spec/issues/739

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -64,12 +64,17 @@ duration in accordance with C11 section 7.6 "Floating-point environment
 | v0-v31  |              | Temporary registers          | No
 | vl      |              | Vector length                | No
 | vtype   |              | Vector data type register    | No
+| vxrm    |              | Vector fixed-point rounding mode register    | No
+| vxsat   |              | Vector fixed-point saturation flag register  | No
 |===
 
 
 Vector registers are not used for passing arguments or return values; we
 intend to define a new calling convention variant to allow that as a future
 software optimization.
+
+The `vxrm` and `vxsat` fields of `vcsr` are not preserved across calls and the
+value is unspecified upon entry.
 
 Procedures may assume that `vstart` is zero upon entry. Procedures may assume
 that `vstart` is zero upon return from a procedure call.


kito-cheng:
LLVM vector folks: @rofirrim @frasercrmck @topperc @asb @luismarques 
GNU vector folks: @zhongjuzhe @rjiejie
General vector folks: @nick-knight @aswaterman @kasanovic @rdolbeau

Sorry for the last minutes changes for the vector ABI, I need you guys feedback or LGTM, let me know if the definition isn't clear to you or any concern about this design.

Thanks!

riscv-cc.adoc
```diff
@@ -64,13 +64,18 @@ duration in accordance with C11 section 7.6 "Floating-point environment
...
+| vxsat   |              | Vector fixed-point saturation flag register  | No
 |===
 
 
 Vector registers are not used for passing arguments or return values; we
 intend to define a new calling convention variant to allow that as a future
 software optimization.
 
+The `vxrm` and `vxsat` fields of `vcsr` are not preserved across calls and the
+value is unspecified upon entry
```
aswaterman, line 77: Missing period at end of sentence.

aswaterman (approved):
As discussed elsewhere, I strongly endorse this addition to the spec.

riscv-cc.adoc
kito-cheng: Thanks!

kito-cheng:
Changes:

- Add missing period.

kito-cheng:
@palmer-dabbelt @preames

Do you mind take a review for that? thanks :)

palmer-dabbelt:
My only worry is the "unspecified on entry" bit.  Not that it's wrong here, but that it's something glibc may rely on for v0-v31 and thus calling it out only here is a bit confusing.

kito-cheng:
> My only worry is the "unspecified on entry" bit. Not that it's wrong here, but that it's something glibc may rely on for v0-v31 and thus calling it out only here is a bit confusing.

@palmer-dabbelt this PR isn't change anything for v0-v31, it's only focus on `vxrm` and `vxsat`* which is used for fixed-point vector operation only, general vector register is affect by https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/290, however psABI isn't limit user using different calling convention so that can be used for passing parameter.

* `vxsat` also used in P extension.

palmer-dabbelt:
> > My only worry is the "unspecified on entry" bit. Not that it's wrong here, but that it's something glibc may rely on for v0-v31 and thus calling it out only here is a bit confusing.
> 
> @palmer-dabbelt this PR isn't change anything for v0-v31, it's only focus on `vxrm` and `vxsat`*1 which is used for fixed-point vector operation only, general vector register is affect by #290, however psABI isn't limit user using different calling convention so that can be used for passing parameter.
> 
> *1: `vxsat` also used in P extension.

Ah, I hadn't seen that other one.  I think this is good, but there's a lot in flux right now and these are pretty new wordings so it'll probably be easier to read for real when things settle down.

zhongjuzhe:
LGTM

rdolbeau:
@kito-cheng wrote:
> all existing libraries are NOT documented as might change fixed point rounding mode, that means we must backup and restore the content of vxrm and vxsat at those library functions

Checking my copy of ISO/IEC 9899:1999 (don't have 2011), I tend to disagree; the next point in 7.6 is:

>  a function call is assumed to require default floating-point control modes, unless its documentation promises otherwise;

Meaning if you handle the vector environment like the FP environment, then e.g. vectorized libm functions don't need to touch `vxrm`, they can assume it's in the proper state on entry. It's the job of the caller to ensure it.

My expectations as a code writer would be that I don't have to worry about it any more than I have to worry about `fenv`, and that I will be in the default state unless I change it explicitly. 'unspecified on entry' mean I have do worry about it in every function, even if it's been called by one of my own function (and odds are it will be left alone by the call, but the specification says I have to check...).

So I would suggest following fenv. Rounding specification for the FPU should be handled same whether the FPU is scalar or vector, IMHO.

As for fixed-point, I have no meaningful experience with it so won't express an opinion.

frasercrmck:
Still LGTM.

kito-cheng:
Hi @rdolbeau thanks for your feedback!

Inline reply but with few reorder for better context:

> So I would suggest following fenv. Rounding specification for the FPU should be handled same whether the FPU is scalar or vector, IMHO.
>
> As for fixed-point, I have no meaningful experience with it so won't express an opinion.

Floating point vector operation are following the same rule as scalar FPU, it focus on fixed-point here, and I must admit that my intuitive is also feel fixed-point rounding mode should following same rule as floating point rounding mode, since they are all `rounding mode`, but after several round discussion with different people, I realized that's different situations between fixed point and floating point: fixed-point rounding mode is likely to change but floating point rounding mode are unlikely to change - most program are use default floating point rounding mode only.

>> a function call is assumed to require default floating-point control modes, unless its documentation promises otherwise;
>
> Meaning if you handle the vector environment like the FP environment, then e.g. vectorized libm functions don't need to touch vxrm, they can assume it's in the proper state on entry. It's the job of the caller to ensure it.

Yeah, that what I thought for fixed-point rounding mode too :P so I can't well understand the situation describe in https://github.com/riscv/riscv-v-spec/issues/739 before,

Default rounding mode would be a big challenge, it's hard to pick certain fixed-rounding mode as default mode, and it also might incompatible with existing run-time environment.



Really appreciate that you shared your opinion even you don't have too much experience with that!

kito-cheng:
@zhongjuzhe @frasercrmck thanks for your ack! :)

rofirrim:
@kito-cheng this comment from @rdolbeau worried me a bit

> 'unspecified on entry' mean I have do worry about it in every function,

I understand the interpretation of `value is unspecified upon entry` does not mean that the value is unpredictable, right?

In other words, if the caller sets the rounding mode right before the call, the callee can still rely on that value being there. The _unspecified_ bit refers to the fact that we cannot rely on it having some particular value.

Is this the intended interpretation?

kito-cheng:
> I understand the interpretation of value is unspecified upon entry does not mean that the value is unpredictable, right?
>
> In other words, if the caller sets the rounding mode right before the call, the callee can still rely on that value being there. The unspecified bit refers to the fact that we cannot rely on always having some particular value.
>
> Is this the intended interpretation?

Yes, your understanding is right, the value is unpredictable at entry and means callee can't rely what caller set, although generally PLT resolver and instrumentation stuffs (e.g. gprof/mcount) won't change the value of vxrm, but there is NO guarantees from ABI side, since we never know which function call in those stuffs might changed the vxrm.

rofirrim:
> > I understand the interpretation of value is unspecified upon entry does not mean that the value is unpredictable, right?
> > In other words, if the caller sets the rounding mode right before the call, the callee can still rely on that value being there. The unspecified bit refers to the fact that we cannot rely on always having some particular value.
> > Is this the intended interpretation?
> 
> Yes, your understanding is right, the value is unpredictable at entry and means callee can't rely what caller set, although generally PLT resolver and instrumentation stuffs (e.g. gprof/mcount) won't change the value of vxrm, but there is NO guarantees from ABI side, since we never know which function call in those stuffs might changed the vxrm.

Understood, thanks for the clarification (my interpretation was actually wrong then but you clarified the intent!).

No objection from my side.

rjiejie:
LGTM

kito-cheng:
We've decide go this way in the psABI, thanks everyone! this can't be made without your review!

riscv-cc.adoc
```diff
@@ -64,13 +64,18 @@ duration in accordance with C11 section 7.6 "Floating-point environment
...
+| vxsat   |              | Vector fixed-point saturation flag register  | No
 |===
 
 
 Vector registers are not used for passing arguments or return values; we
 intend to define a new calling convention variant to allow that as a future
 software optimization.
 
+The `vxrm` and `vxsat` fields of `vcsr` are not preserved across calls and the
+value is unspecified upon entry.
```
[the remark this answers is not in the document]
kito-cheng: For the fields only, PR created https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/302 :)

PR #295, opened 2022-06-23
Require that lazily bound function follow the standard calling conven…

kito-cheng:
…tion

glibc's implementation of lazy symbol resolution relies on the function
being bound following some details of the calling convention that are
not included in the register convention, like the direction of stack
growth.  One could imaging building systems that can lazily resolve
functions while only relying on the register convention, but we've got a
decade or so of binaries that don't behave that way and we can't change
those.

Spited from #286


kito-cheng:
Basically LGTM, this more fit the name of `STO_RISCV_VARIANT_CC`

PR #296, opened 2022-06-23
Tweak the wording for the note of STO_RISCV_VARIANT_CC.

kito-cheng:
The whole point of this VARIANT_CC bit is that we want to be able to use
V registers in _dl_resolve without saving them.  That requires that the
interface between the caller and the callee does not rely on the values
in the V registers.  That's a bit different than how we've histrorically
defined these temporary register clobbers, which is about the the
caller-side visibility into register changes.


kito-cheng:
Spited from #286, https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/286/commits/eefd8b0e5a963391e63a15b65e41ee2945c80534 but slightly modification which is address by #290, @palmer-dabbelt let me know the change is look good to you :)

kito-cheng:
Changes:
- Keep most of current wording, and re-write last sentence

PR #297, opened 2022-07-01
Add note to mention medany also called medium in certain toolchain im…

kito-cheng:
…plementation


kito-cheng:
Yeah, I think so, I guess I should never working on Friday night, but seems like it's Friday night again to me :P

kito-cheng:
Got feedback from two major RISC-V LLVM maintainer, so reopen this as alternative of #301, and clang/LLVM is one of RISC-V major compiler, so this should worth to mention I think.

kito-cheng:
We decide take #301. so close this.

PR #299, opened 2022-07-13
Add the LP64E ABI, to support RV64E

palmer-dabbelt:
This is just a quick copy of ILP32E, not sure if I missed something.

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -246,17 +246,24 @@ provided they hold values no more than FLEN bits wide.
 
 === ILP32E Calling Convention
 
-IMPORTANT: RV32E is not a ratified base ISA and so we cannot guarantee the
-stability of ILP32E, in contrast with the rest of this document. This documents
-the current implementation in GCC as of the time of writing, but may be subject
-to change.
+IMPORTANT: RV32E and RV64E are not ratified base ISAs and so we cannot
+guarantee the stability of ILP32E or LP64E, in contrast with the rest of this
+document. This documents the current ILP32E implementation in GCC as of the
+time of writing, but may be subject to change.
 
 The ILP32E calling convention is designed to be usable with the RV32E ISA. This
-calling convention is the same as the integer calling convention, except for the
-following differences.  The stack pointer need only be aligned to a 32-bit
-boundary.  Registers x16-x31 do not participate in the calling convention, so
-there are only six argument registers, a0-a5, only two callee-saved registers,
-s0-s1, and only three temporaries, t0-t2.
+calling convention is the same as the integer calling convention, except for
+the following differences.  The stack pointer need only be aligned to a 32-bit
+boundary.
+
+The LP64E calling convention is designed to be usable with the RV32E ISA. This
+calling convention is the same as the integer calling convention, except for
+the following differences.  The stack pointer need only be aligned to a 64-bit
+boundary.
+
+In both ILP32E and LP64E: Registers x16-x31 do not participate in the calling
+convention, so there are only six argument registers, a0-a5, only two
+callee-saved registers, s0-s1, and only three temporaries, t0-t2.
 
 If used with an ISA that has any of the registers x16-x31 and f0-f31, then
 these registers are considered temporaries.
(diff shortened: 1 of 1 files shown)


riscv-cc.adoc
```diff
@@ -246,17 +246,24 @@ provided they hold values no more than FLEN bits wide.
...
 === ILP32E Calling Convention
 
-IMPORTANT: RV32E is not a ratified base ISA and so we cannot guarantee the
-stability of ILP32E, in contrast with the rest of this document. This documents
-the current implementation in GCC as of the time of writing, but may be subject
-to change.
+IMPORTANT: RV32E and RV64E are not ratified base ISAs and so we cannot
+guarantee the stability of ILP32E or LP64E, in contrast with the rest of this
+document. This documents the current ILP32E implementation in GCC as of the
+time of writing, but may be subject to change.
```
luismarques, line 252:
```suggestion
IMPORTANT: RV32E and RV64E are not ratified base ISAs and so we cannot
guarantee their stability, in contrast with the rest of this
document. This documents the current ILP32E implementation in GCC as of the
time of writing, but may be subject to change.
```

riscv-cc.adoc
```diff
@@ -246,17 +246,24 @@ provided they hold values no more than FLEN bits wide.
...
-calling convention is the same as the integer calling convention, except for the
-following differences.  The stack pointer need only be aligned to a 32-bit
-boundary.  Registers x16-x31 do not participate in the calling convention, so
-there are only six argument registers, a0-a5, only two callee-saved registers,
-s0-s1, and only three temporaries, t0-t2.
+calling convention is the same as the integer calling convention, except for
+the following differences.  The stack pointer need only be aligned to a 32-bit
+boundary.
+
+The LP64E calling convention is designed to be usable with the RV32E ISA. This
```
luismarques, line 259:
```suggestion
The LP64E calling convention is designed to be usable with the RV64E ISA. This
```

kito-cheng:
Generally LGTM, but this will pending until 1.0 release :)

riscv-cc.adoc
```diff
@@ -246,17 +246,24 @@ provided they hold values no more than FLEN bits wide.
 
 === ILP32E Calling Convention
```
nick-knight, line 247:
```suggestion
=== ILP32E and LP64E Calling Conventions
```

asb:
Quick drive-by comment: The sentence "Tag_RISCV_stack_align records the N-byte stack alignment for this object. The default value is 16 for RV32I or RV64I, and 4 for RV32E." in riscv-elf.adoc should be expanded to cover RV64E.

palmer-dabbelt:
> Quick drive-by comment: The sentence "Tag_RISCV_stack_align records the N-byte stack alignment for this object. The default value is 16 for RV32I or RV64I, and 4 for RV32E." in riscv-elf.adoc should be expanded to cover RV64E.

Thanks.  I guess it'd be 8-byte aligned for RV64E, as that's the same tradeoff we made for RV32E (ie, penalize Q in favor of everything else).

Did the RV64E stuff actually happen?  The original announcement was kind of weird, I'm not sure if it was on purpose.

asb:
> Did the RV64E stuff actually happen? The original announcement was kind of weird, I'm not sure if it was on purpose.

Per the [recently ratified extension list](https://wiki.riscv.org/display/HOME/Recently+Ratified+Extensions) it was ratified in January, so it looks like it's official, opening the question of what's next for the RV32E and RV64E ABIs!

palmer-dabbelt:
> > Did the RV64E stuff actually happen? The original announcement was kind of weird, I'm not sure if it was on purpose.
> 
> Per the [recently ratified extension list](https://wiki.riscv.org/display/HOME/Recently+Ratified+Extensions) it was ratified in January, so it looks like it's official, opening the question of what's next for the RV32E and RV64E ABIs!

I guess we should probably get this finished, then.  Have you guys had any interest in RV64E on the LLVM side of things?

asb:
> Have you guys had any interest in RV64E on the LLVM side of things?

I've been giving some guidance to a colleague who saw it was a gap vs the ratified extensions and thought it might be interesting to sort it (initially just RV64E at the MC/assembler layer). But no external demand.

a4lg:
Since RV32E and RV64E are ratified and psABI 1.0 is already released, I think it's a good time to merge this.

Also, we should move forward to ratify ILP32E and LP64E ABIs.

riscv-cc.adoc
a4lg, line 252:
```suggestion
```

Since RV32E and RV64E are ratified, this section is no longer necessary.

asb: RV32E and RV64E are ratified, but we shouldn't instantly mark ILP32E and LP64E as being finalised. For instance, it looks like we don't properly document the alignment on ILP32E for 2xlen types - where I believe _only on the stack_ these types are 4-byte aligned in GCC? (The note about the stack being 4-byte aligned and the D extension not being compatible is relevant - but you could have a 4 byte stack and require it to be realigned for stack objects that need greater alignment).

nazar-pc:
Any blockers preventing this from moving forward?
https://github.com/paritytech/polkavm is using ILP32E/LP64E with Rust/LLVM, so it would be nice to be able to rely on this long-term.

jrtc27:
Having it as an experimental ABI is fine but we need people invested in RV64E to drive this to completion, which likely includes ensuring what's written down is a complete description of GCC's implementation, and that that implementation is coherent (hopefully it is? but LP64E is probably not widely used enough to find all the corner cases).

nazar-pc:
Looks like GCC deprecated LP64E with intention to remove it: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=116152
@koute, this might be interesting to you and the rest of the team.

kito-cheng:
I incline to add that as an experimental ABI too, I don't think ilp64e has enough user to test and verify that yet...but this is is needed as experimental ABI for moving that forward.

PR #301, opened 2022-07-22
Revert "Add note to mention medlow also called small in certain toolc…

kito-cheng:
…hain implementation"

This reverts commit 621b07468ecbb35c9a0803fe2ee18494ad3e0c52.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -59,9 +59,6 @@ lui a0, 0x80000 # a0 = 0xffffffff80000000
 addi a0, a0, -0x800 # a0 = a0 + -2048 = 0xFFFFFFFF7FFFF800
 ----
 
-NOTE: `medlow` also called small code model in certain toolchain implementation
-like LLVM.
-
 === Medium any code model
 
 The medium any code model, or `medany`, allows the code to address the range


luismarques:
LGTM. (I don't have permission to officially review the PR...)

Personally, I would have preferred to have erred more towards being informative and convenient, leaving such information in, rather than being an "unbiased" specification.

asb:
I have a slight preference for retaining the current explanation. It doesn't look like many of these mappings to toolchain-specific namings are required right now, so it's not like it's a big issue for maintenance.

But it's not a huge issue either way.

kito-cheng:
Counter PR for this https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/297

jrtc27:
It’s just a matter of principle to me, this is a specification not a description of implementation quirks. Especially when you don’t need to know the LLVM mappings to use Clang as the standard names do work; the underlying ones just also happen to.

asb:
Personally I'd be in favour of compromising on that suggest principle slightly if there's a major benefit in terms of readability / clarity to the end user. But you make a good point that `-mcmodel=medlow` and `-mcmodel=medany` are accepted by clang (and documented in the `--help` output). So on reflection, this particular case isn't a big deal and I'm happier than I was in going ahead with this revert.

kito-cheng:
Sound like we got consensus here, thanks everyone!

PR #304, opened 2022-07-27
Clarify non-sequential relationship with *_HI20 relocs

### LINKED ISSUE #303: Clarification on the relative order of HI20 and LO12 relocations
https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#absolute-addresses

> The R_RISCV_HI20 refers to an LUI instruction containing the high 20-bits to be relocated to an absolute symbol address. The LUI instruction is followed by an I-Type instruction (add immediate or load) with an R_RISCV_LO12_I relocation or an S-Type instruction (store) and an R_RISCV_LO12_S relocation.

It's not very clear to me what exactly "The LUI instruction is followed by an I-Type instruction" means. I read it as "the LUI instruction is guaranteed to appear before its corresponding I-type instruction in a section". But it can be read as "the LUI instruction is executed before its corresponding I-type instruction at runtime, but there's no guarantee on their physical locations."

Which is the intended meaning?

I'm asking this because the mold linker currently assumes that HI20 relocations always appear before LO12 (https://github.com/rui314/mold/blob/main/elf/arch-riscv64.cc#L193-L393). If this is a wrong assumption, I need to change the linkers behavior so that arbitrary order is allowed on HI20 and LO12 relocations.

aswaterman:
R_RISCV_HI20 and R_RISCV_PCREL_HI20 relocs might not precede in program order the relocs they're paired with.

Fixes #303.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -506,10 +506,10 @@ of instructions which have an associated pair of relocations:
 
 The `R_RISCV_HI20` refers to an `LUI` instruction containing the high
 20-bits to be relocated to an absolute symbol address. The `LUI` instruction
-is used in conjunction with an I-Type instruction
-(add immediate or load) with an
-`R_RISCV_LO12_I` relocation or an S-Type instruction (store) and an
-`R_RISCV_LO12_S` relocation. The addresses for pair of relocations are
+is used in conjunction with one or more I-Type instructions (add immediate or
+load) with `R_RISCV_LO12_I` relocations or S-Type instructions (store) with
+`R_RISCV_LO12_S` relocations.
+The addresses for pair of relocations are
 calculated like this:
 
 [horizontal]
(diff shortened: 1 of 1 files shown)


aswaterman:
@jrtc27 please merge if you approve.

jrtc27:
Thanks, I was thinking up alternative wordings yesterday and struggling to find quite the right term, but this is good. Will wait for Kito.

jrtc27 (approved):

luismarques:
This wording (using the singular) suggests that the high and low relocations/instructions always appear in pairs. But I assume that's not actually strictly required? E.g.:

```
lui  a0, %hi(symbol)
sw   a1, %lo(symbol)(a0)
sw   a1, %lo(symbol)(a0)
```

That could appear in a kind of volatile pointer store scenario. As long as we're tweaking this, perhaps it might be worth clarifying such details...?

jrtc27:
I don't think the current wording forbids that

kito-cheng:
@luismarques After re-read the paragraph, I also feel ***an*** instruction in the description give me some impression that is in pair.

How about this?

    The `R_RISCV_HI20` refers to an `LUI` instruction containing the high 20-bits to be relocated to an absolute symbol address. 
    The `LUI` instruction is used in conjunction with one or more I-Type instructions (add immediate or load) with 
    `R_RISCV_LO12_I` relocation or S-Type instructions (store) with `R_RISCV_LO12_S` relocations.

aswaterman:
IMO, that's unrelated to the topic of this PR, but I'll take a stab anyway.

luismarques:
That does seem to address the issue I raised, thanks.
I'm not sure why the "The addresses for pair of relocations are" was moved to a new line, but ignoring that it LGTM.
Maybe update the PR topic to reflect the broader scope before committing?
@jrtc27 Does this still look good to you?

aswaterman:
I've been trying to start sentences on their own lines so it's more obvious when a diff affects only one sentence.

kito-cheng (approved):
LGTM

kito-cheng:
But I would wait one more day to make sure @jrtc27 is OK with that before merge :)

aswaterman:
👍 please merge when ready rather than waiting for me.

PR #310, opened 2022-08-01
Tweak wording in packed bitfield description

asb:
All other references to parts of the structure use "bits xx-yy". So
update the reference to bits 15-10 to do the same.


asb:
Not sure if this is an improvement or not. It stood out to me upon reading through, but I can see the argument that the "bits" in "bits 31-28" also applies to 15-10.

kito-cheng (approved):
My first impression when I read that again is `bits` also applied to `15-10`, but adding `bits` could make the description is more context free which led confusion :P

PR #311, opened 2022-08-01
Escape __attribute__ and __packed__ using asciidoc inline passthrough

asb:
Unlike Markdown `` in Asciidoc simply applies monospace formatting
rather than literal text. This means `__attribute__` and `__packed__`
were being formatted incorrectly. This patch uses the inline passthrough
syntax <https://docs.asciidoctor.org/asciidoc/latest/pass/pass-macro/>
to avoid this.


kito-cheng:
Thanks! Seem like `C++` confused the Asciidoc parsing, let me handled that in another PR.

kito-cheng:
https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/316 fix the pdf formatting issue, got following pdf render result if no #316.

Selection_110

PR #316, opened 2022-08-01
Replace C++ with {Cpp}, ++ might confuse Asciidoc parsing in somecase.

kito-cheng:
#311 use single quote and a plus sign to escape the asciidoc formating,
but seems like it the C++ confuse the asciidoc parser, and then
formating is become strange, so replace all C++ with {Cpp} to prevent
any accidents...


kito-cheng:
Without this fix:
Selection_110

With this fix:
Selection_111

PR #321, opened 2022-08-03
Clarify ELF Reloc Type

### LINKED ISSUE #319: Runtime-only relocations
https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#relocations

> Type
> Whether the relocation is a static or runtime relocation:
> - Static relocations are always resolved by the static linker
> - Runtime relocations can be resolved by both static and dynamic linkers

`RELATIVE`, `COPY`, `JUMP_SLOT`, `TLS_DTPMOD32`, `TLS_DTPMOD64`, `TLS_DTPREL32`, `TLS_DTPREL64`, `TLS_TPREL32`, `TLS_TPREL64` and `IRELATIVE` are runtime-only relocations. They cannot appear as an input to a static linker. That is contrary to the statement that "Runtime relocations can be resolved by both static and dynamic linkers"

MaskRay:
* "runtime relocation" is not a commonly used term. Use "dynamic relocation" instead.
* "Runtime relocations can be resolved by both static and dynamic linkers" is inappropriate.
  Rewrite the description.
* RELATIVE can appear in an executable (usually PIE). Reword the description.
* IRELATIVE can appear in an executable. Just remove "shared object" as
  the previous paragraph has mentioned that "Dynamic" can be used in an executable.
  Change "local ifunc symbol" to the more appropriate "non-preemptible ifunc symbol".
 
Fix #319

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -295,10 +295,11 @@ Enum:: The number of the relocation, encoded in the r_info field
 
 ELF Reloc Type:: The name of the relocation, omitting the prefix of `R_RISCV_`.
 
-Type:: Whether the relocation is a static or runtime relocation:
+Type:: Whether the relocation is a static or dynamic relocation:
 +
-- Static relocations are always resolved by the static linker
-- Runtime relocations can be resolved by both static and dynamic linkers
+- A static relocation relocates a location in a relocatable file, processed by a static linker.
+- A dynamic relocation relocates a location in an executable or shared object, processed by a run-time linker.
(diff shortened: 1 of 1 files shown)


rui314:
LGTM, though I'm not a right person to sign-off.

kito-cheng:
@rui314 Thanks for your review!

@MaskRay Thanks for tweaking those wording, tried to ask Google "dynamic relocation linker" vs "runtime relocation linker" which is 1,060,000 vs 298,000 :P

kito-cheng (approved):

PR #323, opened 2022-08-05
Remove R_RISCV_GNU_VTINHERIT and R_RISCV_GNU_VTENTRY

### LINKED ISSUE #320: What are GNU_VTINHERIT and GNU_VTENTRY?
The psABI spec doesn't say anything other than "GNU C++ vtable hierarchy" and "GNU C++ vtable member usage". Are they deprecated relocation types? If static linkers don't have to support them, we should say so.

MaskRay:
They appeared to be copied from other ports while there is no assembly syntax
for them. Just remove them like we remove R_RISCV_GPREL_[IS].

Fix #320

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -386,9 +386,7 @@ Description:: Additional information about the relocation
                                             <| V - S - A
 .2+| 40      .2+| SUB64         .2+| Static  | _word64_          .2+| 64-bit label subtraction
                                             <| V - S - A
-.2+| 41      .2+| GNU_VTINHERIT .2+| Static  |                   .2+| GNU {Cpp} vtable hierarchy
-                                            <|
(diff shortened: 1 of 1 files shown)


rui314:
All relocation types from 0 to 255 are either allocated or reserved in this psABI. I think we want to be consistent with it.

rui314:
I think it's OK to return them back to "reserved" as these relocations have never been actually used. We need to actively remove declarations of these two relocations from GNU and LLVM toolchains, though. With that, LGTM, but please wait for other reviewers' responses.

kito-cheng (approved):
Although this is not internal use only relocation like R_RISCV_GPREL_[IS], but we never use those two relocation anywhere, and no way to use that since lack of assembly syntax, so just remove that is LGTM.

PR #325, opened 2022-08-09
Document how merge elf file headers and ELF attribute

### LINKED ISSUE #318: Consider making a recommendation on how mismatched ELF attributes are handled by a linker/loader
Currently the spec says "Attributes are used to record information about an object file/binary that a linker or runtime loader
needs to check compatibility.". You might reasonably wonder what compability checks you should be performing. Therefore, I suggest the specification either:
* Makes it explicit that the behaviour of compatibility checks are out of scope of this document (i.e. it takes no position on under what circumstances a warning should be emitted vs a failure to link)
* Make a recommendation. e.g. recommending a warning but continuing to link by default.

I think @compnerd mentioned some issue related to this about linking in ELF files with no .text section (please correct me here / help link to the relevant issue).

kito-cheng:
Also added a NOTE to mention the behavior of merging `e_flags` when the inputs didn't have any executable section.
 
Fix #318


asb:
CC @compnerd too.

Just keeping comments to the suggested logic at this point rather than the phrasing.

At a first look, I think the rules seem sensible. The only bit that I'm concerned about is how to handle versions:
* It's not specified how to handle differing versions for ISA extensions when merging object files. It's not very obvious what to do here other than perhaps emit a warning. For standard extensions, we'd hope that there are no backwards incompatible changes for version numbers > 1.0, but we won't know the versioning policy for any non-standard extensions.
* Producing an error for linking objects with a different privspec version seems like it could be overly harsh (I know in this case you're just rephrasing the existing documented behaviour...). Imagine a compiler raises the privspec version it uses by default. All of a sudden, you might fail to link objects on your system - particularly annoying if e.g. running RISC-V linux natively and you can't link your distro-provided libc any more.

kito-cheng:
> It's not specified how to handle differing versions for ISA extensions when merging object files. It's not very obvious what to do here other than perhaps emit a warning. For standard extensions, we'd hope that there are no backwards incompatible changes for version numbers > 1.0, but we won't know the versioning policy for any non-standard extensions.

IIRC @Nelson1225 and me has discuss this issue long times ago, and try to put the rule as:
- Silently merge any version with bigger version number IF version is large or equal to `1.0`, e.g. Merge `1.0` and `1.1` into `1.1` without warning/error message.
- Raise warning if the version number is  version is less than `1.0`.

And having ad-hoc rule if there any incompatible change after version `1.0`.

> Producing an error for linking objects with a different privspec version seems like it could be overly harsh (I know in this case you're just rephrasing the existing documented behaviour...). Imagine a compiler raises the privspec version it uses by default. All of a sudden, you might fail to link objects on your system - particularly annoying if e.g. running RISC-V linux natively and you can't link your distro-provided libc any more.

Agree with that, and actually I am consider about just deprecate this...since RISC-V using different version scheme for priv spec - everything in ISA string.

kito-cheng:
Changes:
- Specify how to merge different extension version.
- Relax priv spec version check

preames:
Process comment only.

This is a fairly significant proposed change being made very late during the ratification process.  Unless we have strong and compelling reason, I believe this should not be merged into the current document going for ratification and should be deferred to the next psABI version.

kito-cheng:
Spited #336 and #337, only keep uncontroversial part in this PR

kito-cheng:
Gonna merge this, this PR only contain the de facto behaviors.

PR #326, opened 2022-08-16
Improve linker relaxation section

### LINKED ISSUE #315: RELAX relocation description should forward-reference the "code relaxation" chapter
Creating as an issue as I'm not sure how to handle cross-document forward references within the asciidoc toolchain used here. Arguably some other points should forward reference that chapter too.

Other related thoughts:
* Perhaps relaxation should be defined in the "Terms and Abbreviations" table
* "Linker relaxation" seems to be used more commonly than "code relaxation". Perhaps the name of the chapter should be adjusted to reflect this.

kito-cheng:
- Add `linker relaxation` to the "Terms and Abbreviations" table
- Always use `linker relaxation` rather than `relaxation`
- Rename `Code relaxation` to `Linker Relaxation`

Fix #315

### DIFF
--- a/introduction.adoc
+++ b/introduction.adoc
@@ -18,18 +18,19 @@ This specification uses the following terms and abbreviations:
 
 [width=80%]
 |===
-| Term  | Meaning
+| Term              | Meaning
 
-| ABI   | Application Binary Interface
-| gABI  | Generic System V Application Binary Interface
-| ELF   | Executable and Linking Format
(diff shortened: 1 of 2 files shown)


asb:
LGTM (thanks!), though someone who's more of a linker person should double-check too.

This doesn't quite fully close #315, as that also suggested an explicit forward reference (to make it easier to follow for people reading the PDF really).

rui314:
LGTM

Issue #197, opened 2021-07-08
Document position on zero-page optimization for linker in psABI


compnerd:
The binutils linker currently performs a zero-page optimization (near zero address relocations using `R_RISCV_PCREL_HI20`, `R_RIRSCV_PCREL_LO2{I,S}` are rewritten to a `lui` + `addi` form.  lld does not perform this rewriting and will reject input which expects this.

compnerd:
CC: @luismarques

jrtc27:
My view is this is split into two parts:

1. If the symbol is weak undef, then my opinion in #126 applies that this is a CodeGen bug. Possibly-undef weak symbols in medany code should be referenced using GOT_PCREL_HI20+PCREL_LO12 so that if they are undef they can load a constant 0 from memory. Linkers can optionally choose to optimise this sequence a la X86 GOTPCRELX to not be indirect, but that's not required for functional correctness.
2. If the symbol is anything else, the generated code was wrong. Linkers may choose to support this, but that is a toolchain-specific extension and should not be relied upon, since such symbols violate the premise of the medany code model.

jim-wilson:
There is a philosophical divide here.  On the GNU toolchain side, linker relaxation is considered an important feature of the toolchain for hitting code size and performance goals.  Thus this handling of weak undef is just another relaxation.  On the LLVM toolchain side, linker relaxation is not implemented, and it is considered wrong for the linker to ever change an opcode emitted by the compiler.  Thus this handling of weak undef is considered a compiler bug.

Another concern here is that weak symbols were originally invented for shared libraries, but for ~30 years the GNU toolchain has had an extension allowing weak symbols to be used in statically linked code, and this extension is frequently used by embedded programmers.  There is a weak symbol in the newlib crt0.S for instance.  And the SiFive freedom-e-sdk heavily uses weak symbols, though I'm hoping that this someday gets fixed in a rewrite.  So any solution that increases code size and/or reduces performance for the handling of weak under symbols would be a major problem for embedded users and hardware companies that live and die by their benchmark numbers.

On the linux side we have the problem that most distros are being compiled by gcc, so if the C library is gcc compiled, and contains undef weak, then programs can not be linked using lld.  This problem needs to be fixed, and is probably why this issue was raised.

[...]

PR #328, opened 2022-08-16
Document linker relaxation type.

### LINKED ISSUE #327: Document all possible linker relaxation types
RISC-V has several linker relaxation types like convert PC-relative access to GP-relative access, or `AUIPC+JALR` to `JAL`, but we don't have full list of this, we need to document that in the psABI spec.

kito-cheng:
List all linker relaxation type, and intentionally to not document zero-page relaxation for function call and pc-relative addressing instruction since we haven't reached a consensus yet (see #197 for more detail).

Fix #327

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1148,8 +1148,8 @@ Relaxation result:
 
   Condition:: Offset between global-pointer and symbol is within +-2KiB,
   `R_RISCV_PCREL_LO12_I` and `R_RISCV_PCREL_LO12_S` resolved as indirect
-  relocation pointer, and it will always point to another `R_RISCV_PCREL_HI20`
-  relocation, it will the symbol which pointed by `R_RISCV_PCREL_HI20` in
+  relocation pointer. It will always point to another `R_RISCV_PCREL_HI20`
+  relocation, the symbol pointed by `R_RISCV_PCREL_HI20` will be used in
   the offset calculation.
 
   Relaxation::
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1008,6 +1008,253 @@ are commonly resolved at compile-time, such as intra-function jumps), code
 generators must in general ensure that relocations are always emitted when
 relaxation is enabled.
 
+=== Linker Relaxation Types
+
+The purpose of this section is to describe all types of linker relaxation,
+the linker may implement a part of linker relaxation type, and can be skipped
+the relaxation type is unsupported.
+
+Each candidate relocation might fit more than one relaxation type, the linker
+should only apply one relaxation type.
+
+==== Function Call Relaxation
+
+  Target Relocation::: R_RISCV_CALL, R_RISCV_CALL_PLT.
+
+  Description:: This relaxation type relax can relax `AUIPC+JALR` into `JAL`.
+
+  Condition:: The offset between the location of relocation and target symbol
+  is within +-1MiB.
+
+  Relaxation::
+  - Instruction sequence associated with `R_RISCV_CALL` or `R_RISCV_CALL_PLT.`
+  can be rewrite to a single JAL instruction with the offset between the
+  location of relocation and target symbol.
+
+  Example::
++
+--
+Relaxation candidate:
+[,asm]
+----
+    auipc ra, 0           # R_RISCV_CALL_PLT (symbol), R_RISCV_RELAX
+    jalr  ra, ra, 0
+----
+
+Relaxation result:
+[,asm]
+----
+    jal  ra, 0            # R_RISCV_JAL (symbol)
+----
+--
+
+==== Compressed Function Call Relaxation
+
+  Target Relocation::: R_RISCV_CALL, R_RISCV_CALL_PLT.
+
+  Description:: This relaxation type relax can relax `AUIPC+JALR` into `C.JAL`
+  instruction sequence.
+
+  Condition:: The offset between the location of relocation and target symbol
+  is within +-2MiB and rd operand of second instruction in the instruction
+  sequence is `X1`/`RA`.
+
+  Relaxation::
+  - Instruction sequence associated with `R_RISCV_CALL` or `R_RISCV_CALL_PLT.`
+  can be rewrite to a single `C.JAL` instruction with the offset between the
+  location of relocation and target symbol.
+
+  Example::
++
+--
+Relaxation candidate:
+[,asm]
+----
+    auipc ra, 0           # R_RISCV_CALL_PLT (symbol), R_RISCV_RELAX
+    jalr  ra, ra, 0
+----
+
+Relaxation result:
+[,asm]
+----
+    c.jal  ra, <offset-between-pc-and-symbol>
+----
+--
+
+==== Compressed Tail Call Relaxation
+
+  Target Relocation::: R_RISCV_CALL, R_RISCV_CALL_PLT.
+
+  Description:: This relaxation type relax can relax `AUIPC+JALR` into `C.J`
+  instruction sequence.
+
+  Condition:: The offset between the location of relocation and target symbol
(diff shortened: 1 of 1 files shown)


riscv-elf.adoc
```diff
@@ -1001,6 +1001,253 @@ are commonly resolved at compile-time, such as intra-function jumps), code
...
+should only apply one relaxation type.
+
+==== Function Call Relaxation
+
+  Target Relocation::: R_RISCV_CALL, R_RISCV_CALL_PLT.
+
+  Description:: This relaxation type relax can relax `AUIPC+JALR` into `JAL`.
+
+  Condition:: The offset between the location of relocation and target symbol
+  is within +-1MiB.
```
luismarques, line 1020: I don't know if it's necessary to discuss that the target may not be the symbol itself, but instead the PLT entry for that symbol?

riscv-elf.adoc
```diff
@@ -1001,6 +1001,253 @@ are commonly resolved at compile-time, such as intra-function jumps), code
...
+
+  Target Relocation:: R_RISCV_HI20, R_RISCV_LO12_I, R_RISCV_LO12_S,
+  R_RISCV_PCREL_HI20, R_RISCV_PCREL_LO12_I, R_RISCV_PCREL_LO12_S
+
+  Description:: This relaxation type relax can relax a sequence of load address
+  of a symbol or load, store with a symbol reference into
+  global-pointer-relative instruction.
+
+  Condition:: Offset between global-pointer and symbol is within +-2KiB, and
+  `__global_pointer$` symbol is defined.
```
luismarques, line 1120: Arguably if `__global_pointer$` isn't defined then there is no global pointer, so the condition "Offset between global-pointer and symbol is within +-2KiB" would fail, so requiring the symbol to be defined is redundant. I don't know if we want to document such toolchain-specific details here? (IMO we should still mention that that's the usual way to set the global pointer)

riscv-elf.adoc
```diff
@@ -1001,6 +1001,253 @@ are commonly resolved at compile-time, such as intra-function jumps), code
...
+[,asm]
+----
+    lw t1, <gp-offset-for-symbol>(gp)
+----
+--
+
+NOTE: This relaxation require program initialized the `gp` register with the
+address of `__global_pointer$` symbol before access any symbol address,
+strongly recommended initialize `gp` at the beginning of the program entry
+function like `_start`.
```
luismarques, line 1149: I like these informative notes. I'd also add that such initialization must be done with relaxations disabled for that code portion.

riscv-elf.adoc
```diff
@@ -1001,6 +1001,253 @@ are commonly resolved at compile-time, such as intra-function jumps), code
...
+
+  Target Relocation:: R_RISCV_TPREL_HI20, R_RISCV_TPREL_ADD,
+  R_RISCV_TPREL_LO12_I, R_RISCV_TPREL_LO12_S.
+
+  Description:: This relaxation type relax can relax a sequence of load address of a symbol or
+  load, store with a thread-local symbol reference into a thread-pointer-relative
+  instruction.
+
+  Condition:: Offset between thread-pointer and thread-local symbol is within
+   +-2KiB
```
luismarques, line 1225: Detail the exact requirements in terms of the instructions being sequential or not, all of the three relocations having an associated relaxation relocation, etc.

riscv-elf.adoc
kito-cheng: Addressed in https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/328/commits/3f1a3965e1e5da5794bb53f83120cc7c3141faa9

riscv-elf.adoc
kito-cheng: Addressed by https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/328/commits/96d6a968878fad747b165178035012a750d3f1ae

riscv-elf.adoc
kito-cheng: Try to address that in https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/328/commits/4a47dbdcf53f7daa345ef0df918100d6229e9dae and also add description for `R_RISCV_PCREL_LO12_I` and `R_RISCV_PCREL_LO12_S`.

kito-cheng:
Prebuilt PDF: [riscv-abi.pdf](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/files/9402979/riscv-abi.pdf)


Changes:
- Rebase and grammar fix
- Introduce  concept called relocation group to describe the we can't apply relaxation on only part of relocations which binded with same symbol.
- Describe the `R_RISCV_CALL` or `R_RISCV_CALL_PLT` might also resolved as PLT stub of target symbol.
- Describe `R_RISCV_PCREL_LO12_[I|S]` correctly in the relaxation condition, it will take symbol from the `R_RISCV_PCREL_HI20`

kito-cheng:
Ping, public review is closing soon, and we are trying to close all related PR within this week, let me know if you have comment or concern about this PR, we gonna to merge this PR is no further comment in next few days :)

riscv-elf.adoc
```diff
@@ -1008,6 +1008,298 @@ are commonly resolved at compile-time, such as intra-function jumps), code
...
+  location of relocation and target symbol.
+
+  Example::
++
+--
+Relaxation candidate:
+[,asm]
+----
+    auipc ra, 0           # R_RISCV_CALL_PLT (symbol), R_RISCV_RELAX
+    jalr  ra, ra, 0
```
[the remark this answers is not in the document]
kito-cheng: `ra` is alias of `x1`, see https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc#integer-register-convention

riscv-elf.adoc
```diff
@@ -1008,6 +1008,298 @@ are commonly resolved at compile-time, such as intra-function jumps), code
...
+==== Compressed Function Call Relaxation
+
+  Target Relocation::: R_RISCV_CALL, R_RISCV_CALL_PLT.
+
+  Description:: This relaxation type relax can relax `AUIPC+JALR` into `C.JAL`
+  instruction sequence.
+
+  Condition:: The offset between the location of relocation and target symbol or
+  the PLT stub of the target symbol is within +-2KiB and rd operand of second
+  instruction in the instruction sequence is `X1`/`RA`.
```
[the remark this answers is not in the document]
kito-cheng: Thank, I missed this condition.

kito-cheng:
Changes:
- Grammar tweaks.
- Address (most) @rui314's comments.

riscv-elf.adoc
```diff
@@ -1008,6 +1008,298 @@ are commonly resolved at compile-time, such as intra-function jumps), code
...
+Relaxation candidate:
+[,asm]
+----
+    lui t0, 0       # R_RISCV_HI20 (symbol), R_RISCV_RELAX
+    lw t1, 0(t0)    # R_RISCV_LO12_I (symbol), R_RISCV_RELAX
+----
+Relaxation result:
+[,asm]
+----
+    c.lui t0, 0     # RVC_LUI (symbol), R_RISCV_RELAX
```
rui314, line 1262: `C.LUI` cannot be used if the immediate is 0. If that's the case, I think we need to use `C.ADDI t0, x0, 0` instead.

kito-cheng: `C.LUI = 0` case should apply zero-page relaxation in practice, added explicitly condition for this relaxation type and update asm example.

rui314:
LGTM

FYI, we have implemented AUIPC → JAL/C.J/C.JAL and TP relaxation to mold, and it looks like they are working as expected.

kito-cheng:
Thanks everyone! finally we have more complete document for linker relaxation now :)

PR #330, opened 2022-08-17
Drop unclear description for e_machine

kito-cheng:
I don't even know what's RISC-V v2 family means, that seems existing there at least 5 years ago.


aswaterman (approved):
This sentence is a convoluted way of saying that we only support ratified base ISAs, not old versions that predate ratification.  I agree that is implicit and so it's OK to delete the sentence.

PR #331, opened 2022-08-17
Add Status of ABI

### LINKED ISSUE #312: Document should be clear about which ABIs are ratified / proposed for ratification
I get the impression that ILP32E _isn't_ going to be ratified as part of the [current process](https://discourse.llvm.org/t/risc-v-public-review-for-non-isa-specification-psabi/63822). What about LP64Q?

In both cases (and for any other part of the spec), it would be helpful to be clear - perhaps a status table like the one in the ISA manual?

kito-cheng:
Preview PDF: [riscv-abi.pdf](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/files/9363412/riscv-abi.pdf)

Fix #312

### DIFF
--- a/introduction.adoc
+++ b/introduction.adoc
@@ -34,3 +34,21 @@ This specification uses the following terms and abbreviations:
 | FLEN              | The width of a floating-point register in bits
 | Linker relaxation | A mechanism for optimizing programs at link-time, see <<Linker Relaxation>> for more detail.
 |===
+
+= Status of ABI
+
+[width=80%]
+|===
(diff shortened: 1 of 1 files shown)


asb:
Thanks Kito - I think this should mention LP64Q as well?

kito-cheng:
Added, thanks! https://github.com/riscv-non-isa/riscv-elf-psabi-doc/commit/08c1998681bb7a9a5c74714b0286e6b2c85eb644

PR #332, opened 2022-08-17
Explicitly disallow larger-than-type bitfields

kito-cheng:
C language standard is not allow that and both GCC and LLVM will
emit error when the width of bitfield exceeds width of its type.

C99 6.7.2.1-3:
```
The expression that specifies the width of a bit-field shall be an integer constant
expression that has nonnegative value that shall not exceed the number of bits in an object
of the type that is specifies
...
```

C++11 9.6-1:

```
The bit-ﬁeld attribute is not part of the type of the class member. The constant-expression shall be an integral constant 
expression with a value greater than or equal to zero. The value of the integral constant expression may be larger than the 
number of bits in the object representation (3.9) of the bit-ﬁeld’s type; in such cases the extra bits are used as padding bits
and do not participate in the value representation (3.9) of the bit-ﬁeld.
...
```

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -152,6 +152,9 @@ undefined.  By contrast, `struct { short x : 10; short y : 12; }` is a 32-bit
 type with `x` in bits 9-0, `y` in bits 27-16, and bits 31-28 and bits 15-10
 undefined.
 
+Bitfields wider than its integer type is not allowed. For example,
+`struct { short x : 17; }` is not allowed.
+
 Arguments passed by reference may be modified by the callee.
 
 Floating-point reals are passed the same way as aggregates of the same size;


nick-knight:
C++ allows wider-than-type bitfields:

> If the specified size of the bit-field is greater than the size of its type, the value is limited by the type: a [std::uint8_t](http://en.cppreference.com/w/cpp/types/integer) b : 1000; would still hold values between 0 and 255. the extra bits are [padding bits](https://en.cppreference.com/w/cpp/language/object#Object_representation_and_value_representation).
> [...]
> In the C programming language, the width of a bit-field cannot exceed the width of the underlying type, and whether `int` bit-fields that are not explicitly `signed` or `unsigned` are signed or unsigned is implementation-defined. For example, `int b:3`; may have the range of values `0..7` or `-4..3` in C, but only the latter choice is allowed in C++. 

https://en.cppreference.com/w/cpp/language/bit_field

kito-cheng:
I only tested with C mode, and re-read C++ spec and tested with C++ again, yes, C++ did support.

kito-cheng:
Gonna create another PR since this isn't right description for current implementation on both GCC and clang.

PR #334, opened 2022-08-18
Document the behavior of larger-than-type bitfields

kito-cheng:
C language does not allow that, but C++ language allows that, and has specify
the behavior, so we try to document what GCC and LLVM do for that, fortunately,
they have consistent behavior for larger-than-type bitfields.

C99 6.7.2.1-3:

```
The expression that specifies the width of a bit-field shall be an integer constant
expression that has nonnegative value that shall not exceed the number of bits in an object
of the type that is specifies
...
```

C++11 9.6-1:

```
The bit-ﬁeld attribute is not part of the type of the class member. The constant-expression shall be an integral constant
expression with a value greater than or equal to zero. The value of the integral constant expression may be larger than the
number of bits in the object representation (3.9) of the bit-ﬁeld’s type; in such cases the extra bits are used as padding bits
and do not participate in the value representation (3.9) of the bit-ﬁeld.
...
```

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -152,6 +152,13 @@ undefined.  By contrast, `struct { short x : 10; short y : 12; }` is a 32-bit
 type with `x` in bits 9-0, `y` in bits 27-16, and bits 31-28 and bits 15-10
 undefined.
 
+Bitfields may larger than its integer type, bits excess than its integer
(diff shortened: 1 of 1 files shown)


nick-knight:
This addresses the concern I raised in #332 . (I'm not permitted to review PRs in this repo; this is my informal approval.)

PR #336, opened 2022-08-19
Update merge rule for Tag_RISCV_priv_spec*

kito-cheng:
The merge rule is too restrict for now, relax the merge rule to
accept different version, but linker require to report error if
incompatible.


kito-cheng:
We decide deprecate that (#339) instead of update the merge rule, which decide in last [psABI call](https://github.com/riscv-admin/psabi/blob/master/MINUTES/meeting-20220822.adoc#325-document-how-merge-elf-file-headers-and-elf-attribute).

PR #337, opened 2022-08-19
Clarify on how to merging the same extension with different versions

kito-cheng:
We didn't specify how linker should merge the extension with different
versions before, and here is proposal for how to manage that.

NOTE: binutils just silently merge that for now.


kito-cheng:
We would like to more non-normative text to describe how to handle that like https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/338

PR #338, opened 2022-08-22
Add recommendation for how to merging the same extension with differe…

kito-cheng:
…nt versions

We didn't specify how linker should merge the extension with different
versions before, and here is recommendation for how to manage that,
rather mandatory rule about that.

NOTE: binutils just silently merge that for now.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1088,6 +1088,19 @@ Merge Policy:::
 The linker should merge the different architectures into a superset when
 object files are merged, and should report errors if the merge result contains
 conflict extensions.
++
+This specification does not mandate rules on how to merge ISA strings that
+refer to different versions of the same ISA extension.
+The suggested merge rules are as follows:
++
+* Merge versions into the latest version of all input versions that are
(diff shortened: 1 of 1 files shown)


kito-cheng:
Changes:
- Refer `ratified version` rather than refer a specific version number like `1.0`

riscv-elf.adoc
```diff
@@ -1088,6 +1088,18 @@ Merge Policy:::
 The linker should merge the different architectures into a superset when
 object files are merged, and should report errors if the merge result contains
 conflict extensions.
++
+This specification does not mandatory rules on how to merge different versions.
```
asb, line 1092: Perhaps: "This specification does not mandate rules on how to merge ISA strings that refer to different versions of the same ISA extension."

asb:
Left a suggested reword comment, but I think the guidance is sensible.

riscv-elf.adoc
kito-cheng: Thanks for the suggestion!

PR #340, opened 2022-08-22
Deprecated R_RISCV_CALL, prefer R_RISCV_CALL_PLT

### LINKED ISSUE #98: R_RISCV_CALL vs R_RISCV_CALL_PLT
```c
int foo();
int g() { return foo()+1; }
```

This compiles to `R_RISCV_CALL` (`call foo)` in `-fno-pic` mode and `R_RISCV_CALL_PLT` (`call foo@plt`) in -fpie/-fpic mode.

IMHO the distinction is not really useful. We can avoid `R_RISCV_CALL` and use `R_RISCV_CALL_PLT` everywhere (also in `-fno-pic` mode). If the target symbol is non-preemptable (local/hidden/not -shared/etc), the linker can omit PLT creation for `R_RISCV_CALL_PLT`. AFAICT, the only differences are these lines:

```c
// binutils-gdb/bfd/elfnn-riscv.c
	case R_RISCV_CALL_PLT:
	  /* This symbol requires a procedure linkage table entry.  We
	     actually build the entry in adjust_dynamic_symbol,
	     because this might be a case of linking PIC code without
	     linking in any dynamic objects, in which case we don't
	     need to generate a procedure linkage table after all.  */

	  if (h != NULL)
	    {
	      h->needs_plt = 1;
	      h->plt.refcount += 1;
	    }
	  break;

	case R_RISCV_CALL:
	  /* Handle a call to an undefined weak function.  This won't be
	     relaxed, so we have to handle it here.  */
	  if (h != NULL && h->root.type == bfd_link_hash_undefweak
	      && h->plt.offset == MINUS_ONE)
	    {
	      /* We can use x0 as the base register.  */
	      bfd_vma insn = bfd_get_32 (input_bfd,
					 contents + rel->r_offset + 4);
	      insn &= ~(OP_MASK_RS1 << OP_SH_RS1);
	      bfd_put_32 (input_bfd, insn, contents + rel->r_offset + 4);
	      /* Set the relocation value so that we get 0 after the pc
		 relative adjustment.  */
	      relocation = sec_addr (input_section) + rel->r_offset;
	    }
```

I don't know much about BFD internals, but it appears the existence of a PLT in -fpie/-fpic code does not make differences. R_RISCV_CALL_PLT is not necessary to force the creation of a PLT if the symbol is non-preemptable.

In -fno-pic code, there may be a so-called "canonical PLT" (created due to an absolute/pc relative relocation to a function). However, its handling has nothing to do with the dichotomy of `R_RISCV_CALL{,_PLT}`.

I don't know what the weak-undef code does.

On x86_64, newer gas/llvm-mc (after r358652) produces `R_X86_64_PLT32` for `call foo` and `call foo@plt`. `R_X86_64_PLT32` can be optimized to `R_X86_64_PC32` if the symbol is non-preemptable.

powerpc32 ELFv1 specifies `R_PPC_LOCAL24PC` `R_PPC_REL24` and `R_PPC_PLTREL24`. powerpc64 ELFv2 just uses `R_PPC64_REL24`.

### LINKED ISSUE #307: Deprecating one of CALL and CALL_PLT?
The psABI says that R_RISCV_CALL and R_RISCV_CALL_PLT had slightly different semantics but are now interchangeable. Then, why don't you deprecate one of the two and recommend always using another? I think it's better than keeping explaining newcomers that the distinction is a historical artifact and they are actually the same thing.

kito-cheng:
Fix #98 and #307.


MaskRay:
This should reference #98 . #307 is a duplicate.

MaskRay (approved):

PR #344, opened 2022-09-02
Fix typos and formatting in relaxation section

ksco:


ksco:
In addition, there is one thing that is not quite sure that needs to be confirmed. In **Compressed LUI Relaxation**, if the Imm is 0, I think we can safely replace it with a `c.li` instruction.

riscv-elf.adoc
```diff
@@ -1393,7 +1393,7 @@ Relaxation result:
   instruction sequence if possible.
 
   Condition:: The symbol address can be presented by a `C.LUI` plus an `ADDI`
-  instructions.
+   or `ld` or `st` instruction.
```
kito-cheng, line 1396: "`ADDI`, load or store instruction.", I saw other place also use `ld` or `st`, but I would prefer use load / store instead of `ld` / `st` here and fit that in other place.

kito-cheng:
> In addition, there is one thing that is not quite sure that needs to be confirmed. In Compressed LUI Relaxation, if the imm is 0, I think we can safely replace it with a c.li instruction.

In theory, yes, it could be replace with `c.li`, but zero-page would be better for such case :)

riscv-elf.adoc
ksco: Done.

kito-cheng:
Thanks!

PR #349, opened 2022-10-25
Document relaxation and speical section in zc extension

linsinan1995:
# 1. Brief Intro

Zcmt extension introduces cm.jt and cm.jalt instructions to compress a jump or function call instruction or instruction sequence to a single table jump instruction (More details on [ZC Spec](https://github.com/riscv/riscv-code-size-reduction)).

During linking, the linker will scan all R_RISCV_CALL, R_RISCV_CALL_PLT, and R_RISCV_JAL relocations and calculate the additional gain brought from table jump relaxation, where the expected size saving from existing relaxations and the size of jump table will be taken into account. If the table jump instruction can bring extra size saving, then the table jump relaxation is enabled to replace the function call or jump instructions. Otherwise, the table jump relaxation will be ignored.

# 2. Linker Workflow

1. Section initialization
Create a table jump section and reserve 256-entry XLEN wide spaces for the target addresses. The first 32 slots are used by cm.jt (jump via table) and the next 192 are used by cm.jalt (jump and link via table).

2. Scanning: calculate additional size gain for each address
In the scan phase, we first create two hash tables corresponding to the cm.jt and cm.jalt instructions. The hash tables will record the size gain for each target address if the table jump relaxation happens. The existing expected size gain from existing relaxation is considered in the calculation. That is, if an R_RISCV_CALL is optimized from auipc+jalr to jal by Function Call Relaxation, the gain is 2 bytes by relaxing the function call into a table jump instruction, and similarly, if auipc+jalr can be compressed to c.j, the gain by replacing it with the table jump instruction is 0.

The calculation process will target R_RISCV_JAL, R_RISCV_CALL, and R_RISCV_CALL_PLT relocations. If there is an additional size gain, it will be accumulated to the gain value of the corresponding address in the hash table. The gains tracked by hash maps of cm.jalt and cm.jt are calculated independently, i.e. the gains of table jump on 32-bit jal ra,symbol1 and jal x0,symbol1 will not be accumulated.

2. Profiling: decide whether the table jump relaxation will be used.
When the traversal of the relocation table is complete, the linker can find the 32 target addresses with the highest gain in the hash table corresponding to cm.jt and the 224 target addresses with the highest gain corresponding to cm.jalt in the hash tables. The total additional size saving can then be calculated after determining the final elements to be filled into the table jump section. If the overall gain is greater than the size of the table jump section, the 32+224 most profitable target addresses are selected to fill into the table jump section, and if there are not enough elements hash table, then 0 is filled in. Unused slots at the end of the section can be trimmed after table jump relaxation.

3. Relax: replace jump or call instructions to table jump instructions
Iterate through all R_RISCV_JAL, R_RISCV_CALL, and R_RISCV_CALL_PLT relocations, if the target address has been filled in and the rd operand matches, then the instruction or instruction sequence can be replaced by `cm.jt index` or `cm.jalt index`.

# 3. Relocation
Zcmt does not need to add an additional relocation type, the target address stored in the table jump section can be relocated using R_RISCV_32 or R_RISCV_64.

# 4. Special Section
Zcmt introduces a linker-created section named .text.riscv.tbljalvec to hold addresses.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -949,15 +949,19 @@ The defined processor-specific section types are listed in <<rv-section-type>>.
 
 [[rv-section]]
 .RISC-V-specific sections
-[cols="3,3,1"]
+[cols="3,3,3"]
 [width=80%]
 |===
 | Name                       | Type                 | Attributes
 
 | .riscv.attributes          | SHT_RISCV_ATTRIBUTES | none
+| .riscv.jvt                 | SHT_PROGBITS         | SHF_ALLOC + SHF_EXECINSTR
 |===
 
 +++.riscv.attributes+++ names a section that contains RISC-V ELF attributes.
+
++++.riscv.jvt+++ is a linker-created section to store table jump
+target addresses. The minimum alignment of this section is 64 bytes.
 
 === Program Header Table
 
@@ -1215,6 +1219,7 @@ NOTE: Using address of PLT stubs of the target symbol or address target symbol
 directly will resolve by linker according to the visibility of the target
 symbol.
 
+[[compress-func-call-relax]]
 ==== Compressed Function Call Relaxation
 
   Target Relocation::: R_RISCV_CALL, R_RISCV_CALL_PLT.
@@ -1248,6 +1253,7 @@ Relaxation result:
 ----
 --
 
+[[compress-tailcall-relax]]
 ==== Compressed Tail Call Relaxation
 
   Target Relocation::: R_RISCV_CALL, R_RISCV_CALL_PLT.
@@ -1330,6 +1336,7 @@ strongly recommended initialize `gp` at the beginning of the program entry
 function like `_start`, and code fragments of initialization must disable
 linker relaxation to prevent initialization instruction relaxed into a NOP-like
 instruction (e.g. `mv gp, gp`).
+[[gp-relax-asm]]
 [,asm]
 ----
     # Recommended way to initialize the gp register.
@@ -1450,6 +1457,81 @@ Relaxation result:
 ----
 --
 
+==== Table Jump Relaxation
+
+  Target Relocation::: R_RISCV_CALL, R_RISCV_CALL_PLT, R_RISCV_JAL.
+
+  Description:: This relaxation type can relax a function call or jump
+  instruction into a single table jump instruction with the index of the target
+  address in table jump section (<<rv-section>>).
+  Before relaxation, the linker scans all relocations and calculates whether
+  additional gains can be obtained by using table jump instructions, where
+  expected size saving from function-call-related relaxations and the size of jump
+  table will be taken into account. If there is no additional gain, then table
+  jump relaxation is ignored. Otherwise, this relaxation is switched on.
+  <<compress-tailcall-relax, Compressed Tail Call Relaxation>> and
+  <<compress-func-call-relax, Compressed Function Call Relaxation>> are
+  always prefered during relaxation, since table jump relaxation has no
+  extra size saving over these two relaxations and might bring a performance
+  overhead.
+
+  Condition:: The `zcmt` extension is required, the linker output is not
+  position-independent and the rd operand of a function call or jump instruction
+  is `X0` or `RA`.
+
+  Relaxation::
+  - Instruction sequence associated with `R_RISCV_CALL` or `R_RISCV_CALL_PLT`
+  can be rewritten to a table jump instruction.
+  - Instruction associated with `R_RISCV_JAL` can be rewritten to a table
+  jump instruction.
+  Example::
++
+--
+Relaxation candidate:
+[,asm]
+----
+    auipc ra, 0           # R_RISCV_CALL (symbol), R_RISCV_RELAX (symbol)
+    jalr  ra, ra, 0
+
+    auipc ra, 0           # R_RISCV_CALL_PLT (symbol), R_RISCV_RELAX (symbol)
+    jalr  x0, ra, 0
+
+    jal ra, 0             # R_RISCV_JAL (symbol), R_RISCV_RELAX (symbol)
+
+    jal x0, 0             # R_RISCV_JAL (symbol), R_RISCV_RELAX (symbol)
+----
+
+Relaxation result:
+[,asm]
+----
+    cm.jalt  <index-for-symbol>
+
+    cm.jt    <index-for-symbol>
+
+    cm.jalt  <index-for-symbol>
+----
+--
+
+NOTE: The `zcmt` extension cannot be used in position-independent binaries.
+
+NOTE: Jump or call instructions with the rd operand `RA` will be relaxed into
+`cm.jalt` and instructions with the rd operand `X0` will be relaxed into
+`cm.jt`. The table jump section holds target addresses for these two
+instructions separately. More details are available in the _ZC* extension
+specification_ <<riscv-zc-extension-group>>.
+
+NOTE: This relaxation requires programs to initialize the `jvt` CSR with the
+address of the `__jvt_base$` symbol before executing table jump
+instructions. It is recommended to initialize `jvt` CSR immediately after
+<<gp-relax-asm,global pointer initialization>>.
+[,asm]
+----
+    # Recommended way to initialize the jvt CSR.
+1:  auipc a0, %pcrel_hi(__jvt_base$)
+    addi  a0, a0, %pcrel_lo(1b)
+    csrw  jvt, a0
+----
+
 [bibliography]
 == References
 
@@ -1468,3 +1550,6 @@ https://www.akkadia.org/drepper/tls.pdf, Ulrich Drepper
 * [[[riscv-unpriv]]] "The RISC-V Instruction Set Manual, Volume I: User-Level
 ISA, Document", Editors Andrew Waterman and Krste Asanovi´c,
 RISC-V International.
+
+* [[[riscv-zc-extension-group]]] "ZC* extension specification"
+https://github.com/riscv/riscv-code-size-reduction


rui314:
From a linker implementer's view, it shouldn't be too hard to implement this relaxation. One thing I'd like to note is that the spec doesn't have to say about how to actually implement it. "Hash tables" are implementation details; there are more efficient find-the-most-frequent-items algorithms. So we can instead just say that the linker finds 32 and 192 most frequent target addresses of R_RISCV_JAL and R_RISCV_CALL/R_RISCV_CALL_PLT, respectively, and construct the jump table if applying this optimization overall reduces the binary size.

MaskRay:
Does this table jump scheme have a demonstrable benefit? It adds significant complexity to the linker. Is this optimization only applicable to the main executable or also shared objects?

aswaterman:
@MaskRay Yes, the code-size advantage justifies the instructions.  This extension is intended for embedded systems, where shared libraries are normally not a consideration.  (It is incompatible with applications-class processors.)

rui314:
I'm curious how much you can actually save. For small programs, auipc+jalr is relaxed at least to jal (and to c.jal if you are lucky), so it looks to me that it can save only 2 bytes for each function call for most frequent 192 functions.

aswaterman:
https://docs.google.com/spreadsheets/d/1bFMyGkuuulBXuIaMsjBINoCWoLwObr1l9h5TAWN8s7k/edit#gid=1837831327

rui314:
Thanks. I was aware of the table but wasn't sure which column to see. It looks like `Zcmt` is the extension that contains only `cm.jt` and `cm.jalt`. So `cm.it` and `.cm.jalt` reduces text size by ~4% on average. Interesting!

linsinan1995:
> From a linker implementer's view, it shouldn't be too hard to implement this relaxation. One thing I'd like to note is that the spec doesn't have to say about how to actually implement it. "Hash tables" are implementation details; there are more efficient find-the-most-frequent-items algorithms. So we can instead just say that the linker finds 32 and 192 most frequent target addresses of R_RISCV_JAL and R_RISCV_CALL/R_RISCV_CALL_PLT, respectively, and construct the jump table if applying this optimization overall reduces the binary size.

Hi @rui314 , thank you for the suggestion. The workflow here indicates the current implementation approach in a [downstream binutils](https://github.com/openhwgroup/corev-binutils-gdb), and I put it here to offer more information about why zcmt requires an additional relaxation process, a new section, symbol, and so forth. The 'linker workflow' part is not added to the abi doc.

riscv-elf.adoc
```diff
@@ -959,10 +959,14 @@ The defined processor-specific section types are listed in <<rv-section-type>>.
 | Name                       | Type                 | Attributes
 
 | .riscv.attributes          | SHT_RISCV_ATTRIBUTES | none
+| .text.riscv.tbljalvec      | SHT_PROGBITS         | none
```
kito-cheng, line 962: Attribute should be `alloc and execute`, which same as text section.

linsinan1995:
Changes:

- update according to @kito-cheng 's feedback.

rui314:
But why it's `SHF_EXECINSTR`? I thought that the table would be placed in `.rodata` since it doesn't contain any code and intended to be read-only.

jrtc27:
Especially since (-static)-pie will need relocations; we don't want text relocations... though it doesn't belong _in_ .rodata just like .got isn't _in_ .rodata, but it should be in the relro segment.

kito-cheng:
> But why it's SHF_EXECINSTR? I thought that the table would be placed in .rodata since it doesn't contain any code and intended to be read-only.

@rui314:
Although I think that could be read-only too, but spec require table must have execute permission: https://github.com/riscv/riscv-code-size-reduction/blob/main/Zc-specification/tablejump.adoc#table-jump-fault-handling

> Especially since (-static)-pie will need relocations; we don't want text relocations... though it doesn't belong in .rodata just like .got isn't in .rodata, but it should be in the relro segment.

@jrtc27:
Maybe a separated section instead of append on `.text` section e.g.`.riscv.tbljalvec`? 

@linsinan1995 IIRC the section wasn't prefixed with `.text` in both lld and binutils implementation? curious why decide prefixed with `.text` now?

jrtc27:
> > But why it's SHF_EXECINSTR? I thought that the table would be placed in .rodata since it doesn't contain any code and intended to be read-only.
> 
> @rui314: Although I think that could be read-only too, but spec require table must have execute permission: https://github.com/riscv/riscv-code-size-reduction/blob/main/Zc-specification/tablejump.adoc#table-jump-fault-handling

Seriously? That's awful and makes no sense. This screams of being designed without consulting toolchain developers given the static PIE issue...

rui314:
Even though the extension is not compatible with shared libraries, it can be used for main executables (which can be relocatable if built as a PIE) at least in theory, so yeah, it's very strange if using this extension mean the output requires text relocations.

rui314:
If we really want to store the jump table into .text section, the values stored to the table should be relative addresses from the beginning of the table instead of absolute addresses. That might actually be desirable because with that design, 1) you don't need to relocate table entries at all as it's position-independent, and 2) storing 32-bit relative address to the table would be suffice even on RV64, which reduces the table size from 2 KiB to 1 KiB.

linsinan1995:
> @linsinan1995 IIRC the section wasn't prefixed with .text in both lld and binutils implementation? curious why decide prefixed with .text now?

Hi @kito-cheng , .text prefix is used in [Binutils implementation](https://github.com/openhwgroup/corev-binutils-gdb/blob/development/bfd/elfnn-riscv.c#L71), but there is no particular reason for this prefix. 

> Especially since (-static)-pie will need relocations; we don't want text relocations... though it doesn't belong in .rodata just like .got isn't in .rodata, but it should be in the relro segment.

Hi @jrtc27, thanks a lot for pointing this out. But IIRC, no text relocation indicates that executable area can not be written, so the problem still exists even if we remove the .text prefix?

> If we really want to store the jump table into .text section, the values stored to the table should be relative addresses from the beginning of the table instead of absolute addresses. That might actually be desirable because with that design, 1) you don't need to relocate table entries at all as it's position-independent, and 2) storing 32-bit relative address to the table would be sufficient even on RV64, which reduces the table size from 2 KiB to 1 KiB. Edit: If addresses in the table are relative, we can in total save 7 KiB; 1 KiB from the table, and 6 KiB from the relocation table. For PIE, we need to emit 256 base relocations for the table, and their total size is 24 bytes for each relocation entry * 256 = 6 KiB.

Hi @rui314, thanks a ton for the input! It sounds like a good solution and brings an extra size saving.

CC @tariqkurd-repo and @abukharmeh , and look for more opinions

abukharmeh:
Regarding execute permission, Tariq made it like this so 

> For table jump I wanted to express it so that it can all be executed in the instruction fetch, it doesn't need to use a load for jump table access, so that table jumps can be executed early in the pipe.

https://github.com/riscv/riscv-code-size-reduction/issues/134

Regarding relative addressing, that might make it easier to use Zcmt for CHERI maybe, but could increase path delay slightly ? I am not completely sure reducing pre-linked object files by 6KiB is something of a great concern to us ?

Tariq is currently away, but I imagine he would reply to this with more details soon !

jrtc27:
> Regarding execute permission, Tariq made it like this so
> 
> > For table jump I wanted to express it so that it can all be executed in the instruction fetch, it doesn't need to use a load for jump table access, so that table jumps can be executed early in the pipe.

There's nothing stopping you from treating it as a data load in instruction fetch. The use of fences may still need to be weird but at least permissions-wise it wouldn't require the ELF to have nonsense and problematic section/segment flags. Ditto for the capability permissions in a hypothetical CHERI+Zcmt world (do you really want your unsealed table capability to be executable? there's no need for it architecturally beyond claimed microarchitectural convenience in certain highly-constrained designs)

rui314:
> Regarding relative addressing, that might make it easier to use Zcmt for CHERI maybe, but could increase path delay slightly ? I am not completely sure reducing pre-linked object files by 6KiB is something of a great concern to us ?

I don't know the answer to that question, but the jump table should be extremely hot as it is accessed for every function call of the most popular 256 functions. So reducing its size by half should mean more than other data structures. It's in the CPU's I (not D?) cache.

abukharmeh:
Hi everyone,

Sorry for requesting update on this again, but I wonder if we currently have any blockers ? or when can this be merged ? 

Thanks,
Ibrahim

jrtc27:
The text relocation issue is the biggest

tariqkurd-repo:
Hi @jrtc27 can you clearly state the blocking issue here, and your proposal for resolving it, for the avoidance of doubt.

rui314:
I'm not convinced by the argument that this is fine because the main use case is RV32 because after all you guys are defining the extension for both RV32 and RV64. You may be using it for small RV32 processors only, but once it is standardized, other people may have different ideas. At least for RV64, an extension that requires text relocations seems pretty strange.

ptomsich:
> I'm not convinced by the argument that this is fine because the main use case is RV32, because after all you guys are defining the extension for both RV32 and RV64. You may be using it for small RV32 processors only, but once it is standardized, other people may have different ideas. At least for RV64, an extension that requires text relocations seems pretty strange.

I will go one step further than this: we can neither reduce the scope to RV32-only nor to "used without an operating system".  The ISA extension itself is generally applicable and the specifications and ecosystem surrounding it have to be generally applicable for the same reason.

Any solution that does not rise to meet the threshold of general applicability will not be sufficient.

jrtc27:
Both are acceptable to me from an ABI perspective, and I also lean towards option 1 personally.

kito-cheng:
Prefer option 1 since it's less disturbing option to the zc* spec in this moment, but both ok to me.

rui314:
For option 1, you can place the table into an RO section only for non-PIC.

For PIC, the section attribute containing the jump table needs to be `SHF_ALLOC | SHF_WRITE`. It has to be writable to allow the dynamic linker to apply dynamic relocations. The static linker should protect the section with RELRO so that the section is write-protected after dynamic relocations are applied.

tariqkurd-repo:
sorry for the slow response on this - I've filed this issue against the spec:

https://github.com/riscv/riscv-code-size-reduction/issues/195

please check my proposed change is accurate.

sorear:
> There's nothing stopping you from treating it as a data load in instruction fetch.

1. Instruction TLBs can currently ignore all permission bits other than X.  This change requires adding an R bit to all instruction TLBs.
2. At present, none of the defined mstatus flags affect instruction fetch. With the proposed change, writes to mstatus/sstatus that affect MXR, SUM, or MPRV potentially affect the fetch stage, so writes to mstatus/sstatus will in the general case have to restart the pipeline instead of just stalling load/store instructions.  If you want to avoid a several cycle penalty on every interrupt disable/enable or copy_from_user, that requires data dependent pipeline logic.
3. If fence.i instructions are no longer required, then handling table lookups as instruction fetches requires adding coherency mechanisms to the I-cache.

Personally I'd rather deal with one extra 32-bit adder (or just forbid Zcmt usage with PIEs) than the above control complications.

aswaterman:
> or just forbid Zcmt usage with PIEs

At risk of igniting a flame war, I'd like us to consider this option more seriously.  Zcmt is fundamentally incompatible with the RVA profiles, and while that doesn't technically mean that it's incompatible with _all_ systems that want PIE, it certainly is incompatible with the most common ones.  I worry that we are solving a non-problem here.

rui314:
As a linker's implementer, I don't have a strong opinion on it. If they are defined as non-PIC-only relocations, a linker can just report an error when such relocations are used for PIC.

tariqkurd-repo:
Hi everyone - have we reached a resolution on this yet?

riscv-elf.adoc
```diff
@@ -1454,6 +1461,85 @@ Relaxation result:
...
+  table jump relaxation is ignored. Otherwise, the relaxation is switched on.
+  <<compress-tailcall-relax, Compressed Tail Call Relaxation>> and
+  <<compress-func-call-relax, Compressed Function Call Relaxation>> are
+  always prefered during relaxation, since the table jump relaxation has no
+  extra size saving over these two relaxations and might bring a performance
+  overhead.
+
+  Condition:: Zcmt extension is required, rd operand of the function call or
+  jump instruction should be `X0` or `RA`, and table jump instructions can
+  bring extra size saving.
```
kito-cheng, line 1484: Add one more condition here: disable with position-independent mode.

kito-cheng (commented):
I am OK with incompatible with PIC, but that should be document in the spec, and updated binutils or LLD implementation are required.

riscv-elf.adoc
tariqkurd-repo: Hi @kito-cheng do I need to modify the Zc spec as well, or are all the details covered by this PR?

rui314: Rather than disabling, it is better to make the linker to report an error if Zcmt and position-independent output are requested at the same time.

kito-cheng: @tariqkurd-repo I think that should be part of psABI spec, have a note in Zc spec is plus :)

kito-cheng: @rui314: I would like use non-normative word here: `incompatible with position-independent mode.` and add a note `NOTE: Linker may report error when Zcmt extension is present and position-independent output are requested at the same time.`

MaskRay:
`.text.riscv.tbljalvec` takes from the space `.text.*` which isn't appropriate (a special case is needed in almost all linkers).

Modern distributions use PIE prevailingly.  I don't think making the extension completely incompatible with PIC justifies the extension in a normal ABI targeting Linux desktop/server, an embedded ABI would be fine. Better to keep supporting PIC (I still have some doubts for the usefulness of this making it into a normal ABI but will not object to it loudly). It's fine to give the section different flags depending on `-pie`/`-shared` vs `-no-pie`.

riscv-elf.adoc
```diff
@@ -953,16 +953,20 @@ The defined processor-specific section types are listed in <<rv-section-type>>.
...
 [[rv-section]]
 .RISC-V-specific sections
-[cols="3,3,1"]
+[cols="3,3,3"]
 [width=80%]
 |===
 | Name                       | Type                 | Attributes
 
 | .riscv.attributes          | SHT_RISCV_ATTRIBUTES | none
+| .text.riscv.tbljalvec      | SHT_PROGBITS         | SHF_ALLOC + SHF_EXECINSTR
```
kito-cheng, line 962: I also prefer that should not prefixed with `.text`

kito-cheng:
@MaskRay I think this extension won't be used in Linux or any *BSD platform in general, but we don't have separated EABI document (yet), so keep that in same document would be the best way for now.

linsinan1995:
updated:
- remove section prefix .text
- add one condition and a note for position-independent mode and zcmt
- grammar/readability change

riscv-elf.adoc
```diff
@@ -1450,6 +1457,87 @@ Relaxation result:
...
+----
+    cm.jalt  <index-for-symbol>
+
+    cm.jt    <index-for-symbol>
+
+    cm.jalt  <index-for-symbol>
+----
+--
+
+NOTE: The linker will report an error if the `zcmt` extension and
```
kito-cheng, line 1513: `will` is too strong for a non-normative test, I would suggest using `may` here.

linsinan1995: Changed. Thanks!

riscv-elf.adoc
```diff
@@ -1450,6 +1457,87 @@ Relaxation result:
...
+  expected size saving from function-call-related relaxations and the size of jump
+  table will be taken into account. If there is no additional gain, then table
+  jump relaxation is ignored. Otherwise, this relaxation is switched on.
+  <<compress-tailcall-relax, Compressed Tail Call Relaxation>> and
+  <<compress-func-call-relax, Compressed Function Call Relaxation>> are
+  always prefered during relaxation, since table jump relaxation has no
+  extra size saving over these two relaxations and might bring a performance
+  overhead.
+
+  Condition:: The `zcmt` extension is required, the linker output is not
```
jrtc27, line 1478:
```suggestion
  Condition:: The `Zcmt` extension is required, the linker output is not
```

riscv-elf.adoc
```diff
@@ -1450,6 +1457,87 @@ Relaxation result:
...
+  position-independent, the rd operand of a function call or jump instruction
+  is `X0` or `RA`, and table jump instructions can bring extra size saving.
```
jrtc27, line 1480:
```suggestion
  is `x0` or `ra`, and table jump instructions can bring extra size saving.
```
Only can? Surely will?

riscv-elf.adoc
```diff
@@ -1450,6 +1457,87 @@ Relaxation result:
...
+  Example::
++
+--
+Relaxation candidate:
+[,asm]
+----
+    auipc ra, 0           # R_RISCV_CALL (symbol), R_RISCV_RELAX (symbol)
+    jalr  ra, ra, 0
+
+    auipc ra, 0           # R_RISCV_CALL (symbol), R_RISCV_RELAX (symbol)
```
jrtc27, line 1496: No _PLT example?

riscv-elf.adoc
```diff
@@ -1450,6 +1457,87 @@ Relaxation result:
...
+----
+    cm.jalt  <index-for-symbol>
+
+    cm.jt    <index-for-symbol>
+
+    cm.jalt  <index-for-symbol>
+----
+--
+
+NOTE: The linker may report an error if the `zcmt` extension and
```
jrtc27, line 1513:
```suggestion
NOTE: The linker may report an error if the `Zcmt` extension and
```



If you're going to talk about linker errors, say "should" not "may", but I would argue we shouldn't be prescribing diagnostics and instead just stating incompatibilities in a spec.

riscv-elf.adoc
```diff
@@ -1450,6 +1457,87 @@ Relaxation result:
...
+position-independent output are requested at the same time.
+
+NOTE: Jump or call instructions with the rd operand `RA` will relaxed into
+`cm.jalt` and instructions with the rd operand `X0` will be relaxed into
+`cm.jt`. Table jump section holds target addresses for these two
+instructions separately. More details on _ZC extension specification_ <<riscv-zc>>.
```
jrtc27, line 1519:
```suggestion
NOTE: Jump or call instructions with the rd operand `ra` will relaxed into
`cm.jalt` and instructions with the rd operand `x0` will be relaxed into
`cm.jt`. The table jump section holds target addresses for these two
instructions separately. More details are available in the _Zc extension specification_ <<riscv-zc>>.
```

Also Zc isn't an extension, you have Zc* as a set of extensions.

riscv-elf.adoc
```diff
@@ -1450,6 +1457,87 @@ Relaxation result:
...
+
+NOTE: This relaxation requires programs to initialize the `jvt` csr with the
```
jrtc27, line 1521:
```suggestion
NOTE: This relaxation requires programs to initialize the `jvt` CSR with the
```

riscv-elf.adoc
```diff
@@ -1450,6 +1457,87 @@ Relaxation result:
...
+address of `__tbljalvec_base$` symbol before executing table jump instructions.
```
jrtc27, line 1522:
```suggestion
address of the `__tbljalvec_base$` symbol before executing table jump instructions.
```

riscv-elf.adoc
```diff
@@ -1450,6 +1457,87 @@ Relaxation result:
...
+Initialization should happen after <<gp-relax-asm,global pointer initialization>>.
```
jrtc27, line 1523: Why? If it's in .option norelax then it doesn't matter, and if it's in .option relax then it's subject to the same rules as any other lla.

riscv-elf.adoc
```diff
@@ -1450,6 +1457,87 @@ Relaxation result:
...
+
+[,asm]
+----
+    # Recommended way to initialize the jvt csr.
```
jrtc27, line 1527: Use a caption not a comment

riscv-elf.adoc
```diff
@@ -1450,6 +1457,87 @@ Relaxation result:
...
+
+    # Routine to initialize `gp` register.
+    .option push
+    .option norelax
+1:  auipc gp, %pcrel_hi(__global_pointer$)
+    addi  gp, gp, %pcrel_lo(1b)
+    .option pop
```
jrtc27, line 1534: This isn't needed here and duplicates existing code

riscv-elf.adoc
```diff
@@ -1450,6 +1457,87 @@ Relaxation result:
...
+
+2:  auipc a0, %pcrel_hi(__tbljalvec_base$)
+    addi  a0, a0, %pcrel_lo(2b)
+    csrw jvt, a0
```
jrtc27, line 1538:
```suggestion
    csrw  jvt, a0
```

riscv-elf.adoc
```diff
@@ -1468,3 +1556,6 @@ https://www.akkadia.org/drepper/tls.pdf, Ulrich Drepper
 * [[[riscv-unpriv]]] "The RISC-V Instruction Set Manual, Volume I: User-Level
 ISA, Document", Editors Andrew Waterman and Krste Asanovi´c,
 RISC-V International.
+
+* [[[riscv-zc]]] "ZC extension specification"
```
jrtc27, line 1560: Same as above

riscv-elf.adoc
```diff
@@ -949,16 +949,20 @@ The defined processor-specific section types are listed in <<rv-section-type>>.
...
 [[rv-section]]
 .RISC-V-specific sections
-[cols="3,3,1"]
+[cols="3,3,3"]
 [width=80%]
 |===
 | Name                       | Type                 | Attributes
 
 | .riscv.attributes          | SHT_RISCV_ATTRIBUTES | none
+| .riscv.tbljalvec           | SHT_PROGBITS         | SHF_ALLOC + SHF_EXECINSTR
```
jrtc27, line 958: I still don't like that this is tbljalvec, ie TJV, but the CSR is JVT, there should be consistency throughout in what this extension, CSR, section and relaxation is called.

tariqkurd-repo: maybe this is leftover from the original name of that CSR, before we renamed it to JVT. I agree that consistency is much better.

rui314: I guess that the table needs to be aligned to a natural boundary. Was that mentioned somewhere in the doc?

riscv-elf.adoc
kito-cheng: > Only can? Surely will?

My understanding is that means we should not apply this relaxation if it can't bring extra size saving.

riscv-elf.adoc
kito-cheng: Add a NOTE to mention that should at least align to 2 byte and suggest align to 4 byte?

@tariqkurd-repo any suggestion on alignment of the table?

riscv-elf.adoc
jrtc27: Yes, but this is under Condition::, and the condition is that it will

riscv-elf.adoc
linsinan1995: Hi @jrtc27 . I keep some formatting issues unchanged since the abi spec also uses them in the same way. e.g.  extensions are lowercases and the register names in relaxation section are all capitalized. Maybe @kito-cheng or someone can help fix them with another PR.

riscv-elf.adoc
linsinan1995: Thanks for the suggestion. The line about the linker behaviour was removed and only `incompatible with PIC` is mentioned.

riscv-elf.adoc
linsinan1995: Thanks. I removed this line from the condition since it is not really a condition. (at least we could force the linker to generate the table in corev toolchain.)

riscv-elf.adoc
linsinan1995: the alignment added. Thanks!

tariqkurd-repo: sorry for the slow repy @linsinan1995 , it needs to be at least 64-byte aligned as bits [5:0] of JVT aren't part of the address

tariqkurd-repo:
Hi everyone, are there any more outstanding issues? Any more review comments?

riscv-elf.adoc
linsinan1995: > sorry for the slow reply @linsinan1995 , it needs to be at least 64-byte aligned as bits [5:0] of JVT aren't part of the address, which agrees with what you wrote

Hi, @tariqkurd-repo . The alignment requirement was added in the latest commit.
```
+++.riscv.tbljalvec+++ is a linker-created section to store table jump
target addresses. The minimum alignment of this section is 64 bytes.
```

riscv-elf.adoc
```diff
@@ -1450,6 +1457,79 @@ Relaxation result:
...
+
+NOTE: This relaxation requires programs to initialize the `jvt` CSR with the
+address of the `__tbljalvec_base$` symbol before executing table jump
+instructions. It is recommended to initialize `jvt` CSR immediately after
+<<gp-relax-asm,global pointer initialization>>.
+[,asm]
+----
+    # Recommended way to initialize the jvt CSR.
+2:  auipc a0, %pcrel_hi(__tbljalvec_base$)
+    addi  a0, a0, %pcrel_lo(2b)
```
jrtc27, line 1529:
```suggestion
1:  auipc a0, %pcrel_hi(__tbljalvec_base$)
    addi  a0, a0, %pcrel_lo(1b)
```

riscv-elf.adoc
```diff
@@ -1450,6 +1457,79 @@ Relaxation result:
...
+----
+    cm.jalt  <index-for-symbol>
+
+    cm.jt    <index-for-symbol>
+
+    cm.jalt  <index-for-symbol>
+----
+--
+
+NOTE: The `zcmt` extension is incompatible with the position-independent mode.
```
jrtc27, line 1513:
```suggestion
NOTE: The `zcmt` extension cannot be used in position-independent binaries.
```

riscv-elf.adoc
jrtc27: So everything is consistently table jump now, except for the CSR...

tariqkurd-repo: is there any reason not to replace the string `tbljalvec` with `jvt`, as I did in the Zc spec many months ago? I vote for consistency.

tariqkurd-repo: Hi @linsinan1995 should we rename `tbljalvec` to `jvt`? Then can we merge this PR?

linsinan1995:
updated:
1. some changes according to @jrtc27 's suggestions.
2. rename `tbljalvec` to `jvt`. So the used names are `jvt` for the CSR, `.riscv.jvt` for the jump table entry section and `__jvt_base` for the symbol holding the address of the jump table entry section.

tariqkurd-repo:
Excellent, thanks @linsinan1995

tariqkurd-repo:
Are we still waiting for reviews from anyone, or can we merge now? Who actually does the merge?

riscv-elf.adoc
kito-cheng: I am OK to fix this with other stuffs later

kito-cheng (approved):
Generally LGTM, but I would like to gather one more LGTM from LLVM community per our policy.

@MaskRay @jrtc27 @rui314

riscv-elf.adoc
```diff
@@ -1450,6 +1457,79 @@ Relaxation result:
...
+Relaxation candidate:
+[,asm]
+----
+    auipc ra, 0           # R_RISCV_CALL (symbol), R_RISCV_RELAX (symbol)
+    jalr  ra, ra, 0
+
+    auipc ra, 0           # R_RISCV_CALL_PLT (symbol), R_RISCV_RELAX (symbol)
+    jalr  x0, ra, 0
+
+    jal ra, 0             # R_RISCV_JAL (symbol), R_RISCV_RELAX (symbol)
```
rui314, line 1499: Is it allowed or prohibited to relax

```
jal x0, 0      # R_RISCV_JAL (symbol), R_RISCV_RELAX (symbol)
```

to

```
cm.jt <index-for-symbol>
```

?

kito-cheng: Plz add example :)

linsinan1995: added it. Thanks.

rui314: You seems to forgot to add the following line

```
cm.jt <index-for-symbol>
```

linsinan1995: > You seem to forgot to add the following line
> 
> ```
> cm.jt <index-for-symbol>
> ```

Added. Thanks a lot!

rui314:
LGTM, but my LGTM probably shouldn't be considered as one from the LLVM community.

rui314 (approved):

jrtc27:
Please squash your 5 commits down into 1

linsinan1995:
> Please squash your 5 commits down into 1

pushed. Thanks!

tovine:
> For option 1, you can place the table into an RO section only for non-PIC.
> 
> For PIC, the section attribute containing the jump table needs to be `SHF_ALLOC | SHF_WRITE`. It has to be writable to allow the dynamic linker to apply dynamic relocations. The static linker should protect the section with RELRO so that the section is write-protected after dynamic relocations are applied.

I'm extremely late to the party, but just wanted to note that if it (in the future) becomes desirable to make this PIC-compatible (and perhaps more suitable to application class cores?), we could probably implement this JTVEC/JVT-relative addressing mode, since we have 6 configuration bits reserved for future modes in the [JVT CSR](https://github.com/riscv/riscv-code-size-reduction/blob/main/Zc-specification/jvt_csr.adoc).
This could be something to consider for a future revision of the spec.

tariqkurd-repo:
>I'm extremely late to the party, but just wanted to note that if it (in the future) becomes desirable to make this PIC-compatible (and perhaps more suitable to application class cores?), we could probably implement this JTVEC/JVT-relative addressing mode, since we have 6 configuration bits reserved for future modes in the [JVT CSR](https://github.com/riscv/riscv-code-size-reduction/blob/main/Zc-specification/jvt_csr.adoc).
This could be something to consider for a future revision of the spec.

Yes, that's certainly possible. It's left open for a wide range of future expansion.

@jrtc27 any further comments? Still waiting for your approval.

tariqkurd-repo:
@jrtc27 any reason not to merge this one now? Any other comments from anyone?

kito-cheng:
Will merge at this Friday if no further comment or objection.

tariqkurd-repo:
thank you!

PR #350, opened 2022-10-25
Remove wrong punctuation and duplicate Chapter Sections

linsinan1995:
Signed-off-by: Lin Sinan <sinan.lin@linux.alibaba.com>


kito-cheng (approved):
Thanks for catching this!

PR #357, opened 2023-01-09
Add details of .riscv.attributes section layout

vineetgarc:
Signed-off-by: Vineet Gupta <vineetg@rivosinc.com>

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1016,9 +1016,25 @@ linker or runtime loader needs to check compatibility.
 Attributes are encoded in a vendor-specific section of type SHT_RISCV_ATTRIBUTES
 and name .riscv.attributes. The value of an attribute can hold an integer
 encoded in the uleb128 format or a null-terminated byte string (NTBS).
+The tag number is also encoded as uleb128.
 
 RISC-V attributes have a string value if the tag number is odd and an integer
 value if the tag number is even.
+
+==== Layout of .riscv.attributes section
+
+The attributes section start with a format-version (uint8 = 'A') followed by
+vendor specific sub-section(s). A sub-section starts with sub-section length
+(uint32), vendor name (NTBS) and one or more sub-sub-section(s).
+
+A sub-sub-section consists of a tag (uleb128), sub-sub-section length (uint32)
+followed by actual attribute tag,value pair(s) as specified above.
+Sub-sub-section Tag Tag_file (value 1) specifies that contained attibutes
+relate to whole file.
+
+A sub-section with name "riscv\0" is mandatory. Vendor specific sub-sections
+are allowed in future. Vendor names starting with "[Aa]non" are reserved for
+non-standard ABI extensions.
 
 ==== List of attributes
 


riscv-elf.adoc
```diff
@@ -1016,10 +1016,28 @@ linker or runtime loader needs to check compatibility.
...
 and name .riscv.attributes. The value of an attribute can hold an integer
 encoded in the uleb128 format or a null-terminated byte string (NTBS).
+The tag number is also encoded as uleb128.
 
 RISC-V attributes have a string value if the tag number is odd and an integer
 value if the tag number is even.
 
+==== Layout of .riscv.attributes section
+
+The attributes section contains a header (different from the elf section header)
```
jrtc27, line 1026: This seems like an odd thing to say, especially when the format is then documented

riscv-elf.adoc
```diff
@@ -1016,10 +1016,28 @@ linker or runtime loader needs to check compatibility.
...
+and one or more sub-sections. As of now it only contains one sub-section.
+
+The header is encoded as:
+  uint8 : format-version: 'A'
```
jrtc27, line 1030: This is the only thing in the header

riscv-elf.adoc
```diff
@@ -1016,10 +1016,28 @@ linker or runtime loader needs to check compatibility.
...
+  uint32: section length (includes itself)
+  NTBS  : "riscv\0"
```
jrtc27, line 1032: These are the subsection header, where "riscv\0" is the vendor

riscv-elf.adoc
```diff
@@ -1016,10 +1016,28 @@ linker or runtime loader needs to check compatibility.
...
+
+Each sub-section has a header.
+  uint8 : Tag_Section (uleb128 encoded)
+  uint32: sub-section length (excludes itself)
```
jrtc27, line 1036: This is the "sub-subsection" (Arm's name) header (the format of a subsection is determined based on the vendor, but we use the same format as identified by Arm's "aeabi" vendor name), and the format you're documenting (the only one RISC-V is using) is for Tag_File (the others are deprecated for Arm), which you need to document as being 1.

riscv-elf.adoc
```diff
@@ -1016,10 +1016,28 @@ linker or runtime loader needs to check compatibility.
...
+The attributes section contains a header (different from the elf section header)
+and one or more sub-sections. As of now it only contains one sub-section.
+
+The header is encoded as:
+  uint8 : format-version: 'A'
+  uint32: section length (includes itself)
+  NTBS  : "riscv\0"
+
+Each sub-section has a header.
+  uint8 : Tag_Section (uleb128 encoded)
```
jrtc27, line 1035: This being uleb128 is rather hypothetical when the only valid values are 1-3... but uint8 disagrees with uleb128

riscv-elf.adoc
```diff
@@ -1016,10 +1016,28 @@ linker or runtime loader needs to check compatibility.
...
+  uint32: sub-section length (excludes itself)
+
+The actual attributes follow directly as tag:value pairs, terminated by a '\0'
```
jrtc27, line 1038: They're not terminated by anything (maybe you're getting confused by Tag_Section and Tag_Symbol having a 0 byte separator between the section/symbol numbers and attributes list)

riscv-elf.adoc
```diff
@@ -1016,10 +1016,28 @@ linker or runtime loader needs to check compatibility.
...
 encoded in the uleb128 format or a null-terminated byte string (NTBS).
+The tag number is also encoded as uleb128.
 
 RISC-V attributes have a string value if the tag number is odd and an integer
 value if the tag number is even.
 
+==== Layout of .riscv.attributes section
+
+The attributes section contains a header (different from the elf section header)
+and one or more sub-sections. As of now it only contains one sub-section.
```
jrtc27, line 1027: In theory vendors are free to add their own... that's why the vendor string exists (and "[Aa]non.*" is reserved for custom use by Arm)

jrtc27 (changes requested):

riscv-elf.adoc
vineetgarc: ok yeah makes sense now that read the ARM elf sepc - my initial interpretation was just reverse engineering from binutils code.

riscv-elf.adoc
vineetgarc: Right

riscv-elf.adoc
vineetgarc: there is a tag 0 at the end to signify end of sub-section which also doubles as string terminator in case the last tag/vaue is a NTBS.

jrtc27: According to what? The string terminator is always there and is part of the attribute, and if you have a ULEB128 then the final byte isn't 0.

jrtc27: What you've documented would have _two_ zero bytes at the end of the last string and a zero byte on the end of the last ULEB128.

riscv-elf.adoc
vineetgarc: FWIW unlike the ARM spec, the RV byte-size encoding for Tag_file doesn't include the lengths of tag or byte-size fields.

```
  [25] .riscv.attributes RISCV_ATTRIBUTE  0000000000000000  000b5e78
       0000000000000066  0000000000000000           0     0     1
                    ^^^ (length)
‘A’   sub-sec length    ‘r’  ‘i’  ‘s’  ‘c’ ‘v’ ‘\0’
41     65 0 0 0             72 69 73 63 76 0 

Tag-file   byte-size (of following contents)
1            5b 0 0 0
```

riscv-elf.adoc
vineetgarc: Right I understand and will fix that.

riscv-elf.adoc
vineetgarc: I was meaning to clarify that this is an "inline" header within the section as opposed to actual section table entry / header for this section as I got confused initially. But it was all muddled up due to section / sub-section confusion. But as you say the header part here is just 'A' so need not be called out as header per se.

riscv-elf.adoc
jrtc27: Yes it does. 0x65 == 0x5b + (4 + 6), the subsection header size

vineetgarc: there are 2 length fields here: the first one includes itself, the second one is data length and for ARM that includes the attr tag and itself, whereas it doesn't for RV.

Quoting ARM spec [1]
```
In each case, byte-size is a 4-byte unsigned integer in the byte order of the ELF file. Byte-size includes the initial tag byte, the size field itself, and the sub-subsection content. That is, it is the byte offset from the start of this sub-subsection to the start of the next sub-subsection.```

[1] https://github.com/ARM-software/abi-aa/blob/main/addenda32/addenda32.rst#representing-build-attributes-in-elf-files

Comment (jrtc27) [reply riscv-elf.adoc]: Yes it does for RISC-V. They are the same. I don't see how the readelf output you gave earlier demonstrates otherwise; I just showed how it _does_ add up.

Comment (vineetgarc) [reply riscv-elf.adoc]: You showed how 065 = 0x5b + 4 + 6  - we aagree there as 0x65 includes itself (not the intiail 'A' fwiw)

What I'm saying is that 0x5b doesn't include self length and tag length - because if it would then it would be 0x65 - 4 = 0x61 not 0x5b

File: riscv-elf.adoc [hunk 8]
```diff
@@ -1016,10 +1016,28 @@ linker or runtime loader needs to check compatibility.
...
 
+==== Layout of .riscv.attributes section
+
+The attributes section start with a format-version (uint8 = 'A') followed by
+vendor specific sub-section(s). Currently there is single "riscv" vendor
+sub-section encoded as follows:
+
+  uint32: sub-section length (includes itself)
+  NTBS  : Vendor Name "riscv\0"
+  uint8 : 1 (Tag_File)
```
jrtc27, line 1032: It's legal to have more than one file sub-subsection as far as I can tell, they just behave as if you concatenated them all together

riscv-elf.adoc
jrtc27: What is "self length" and "tag length"? 0x5b is 0x66 minus the "sub-sec length" (4 bytes) and "riscv\0" (6 bytes) as in your dump above, that is it is the size of the entire sub-subsection, _including_ the sub-subsection header.

vineetgarc: /me slaps myself - you are right of course - i was just overlooking the string chars themselves :-(

riscv-elf.adoc
```diff
@@ -1016,10 +1016,28 @@ linker or runtime loader needs to check compatibility.
...
+  uint32: sub-section length (includes itself)
+  NTBS  : Vendor Name "riscv\0"
+  uint8 : 1 (Tag_File: attributes apply to whole file)
+  uint32: byte size (includes tag byte, itself and attribtes following)
+
+The actual attributes follow directly as tag:value pairs
+  <tag1><value1><tag2><value2>
+
+In future, RISCV vendors can add vendor specific sub-sections as needed.
+The vendor name "[An]non" is reserved by ARM.
```
jrtc27, line 1039: 1. [Aa] not [An]
2. It's all vendors that start with that, which is why I said "[Aa]non.*"
3. Arm reserve it for their section, but this isn't Arm's section, it's ours, which is the same format as theirs but isn't theirs, so it's not really reserved by Arm here. Also, saying it's reserved by Arm isn't helpful, it's reserved by Arm _for unregistered private use_, which for us would be nonstandard ABI extensions (mirroring the terminology used earlier for relocations). So just say they're reserved for nonstandard ABI extensions.

riscv-elf.adoc
vineetgarc: ARM has deprecated Tag_section, Tag_symbol. Do we intend to introduce them to riscv abi or just have Tag_File.
Assuming it is latter is it still possible to have multiple of them - in RV ABI
I need to add Tag_file to the doc, should we add the rest as well ?

riscv-elf.adoc
vineetgarc: OK.

vineetgarc:
I've revised the contents per review and force pushed the PR.

Thx,

riscv-elf.adoc
```diff
@@ -1016,10 +1016,26 @@ linker or runtime loader needs to check compatibility.
...
 
+==== Layout of .riscv.attributes section
+
+The attributes section start with a format-version (uint8 = 'A') followed by
+vendor specific sub-section(s). A sub-section starts with sub-section length
+(uint32), vendor name (NTBS) and one or more sub-sub-section(s).
+
+A sub-sub-section consists of a tag (uleb128), sub-sub-section length (uint32)
+followed by actual attribute tag,value pair(s) as specified above.
+Sub-sub-section Tag Tag_RISCV_file (value 1) specifies that contained attibutes
```
kito-cheng, line 1032: Tag_RISCV_file -> Tag_File, the term is used in binutils and LLVM source tree.

kito-cheng (commented):
Just a nit, otherwise LGTM :)

riscv-elf.adoc
vineetgarc: Fixed and force pushed branch !

kito-cheng (approved):
LGTM, wait one week to make sure no further comment.

PR #360, opened 2023-01-30
Define mandatory/optional attribute tag

### LINKED ISSUE #352: Mandatory/optional merging of build attributes
I am concerned with the use of build attributes but I think it's useful noting down a GNU ld arm behavior regarding to `.ARM.attributes` merging.

For a tag number `tag`, if `tag%128 < 64`, the attribute is considered mandatory and GNU ld unaware of it reports an error (which will change the exit status to 1); otherwise, (`tag%128 >= 64`), the attribute is considered optional: there is a warning but the link will succeed (`--fatal-warnings` can upgrade the warning to an error).

```
% cat a.s
.eabi_attribute 88, 2
.eabi_attribute 128, 2
% arm-linux-gnueabi-gcc -nostdlib a.s a.s
/usr/lib/gcc-cross/arm-linux-gnueabi/12/../../../../arm-linux-gnueabi/bin/ld: warning: a.out: unknown EABI object attribute 88
/usr/lib/gcc-cross/arm-linux-gnueabi/12/../../../../arm-linux-gnueabi/bin/ld: a.out: unknown mandatory EABI object attribute 128
/usr/lib/gcc-cross/arm-linux-gnueabi/12/../../../../arm-linux-gnueabi/bin/ld: failed to merge target specific data of file /tmp/ccfJwSI8.o
/usr/lib/gcc-cross/arm-linux-gnueabi/12/../../../../arm-linux-gnueabi/bin/ld: warning: cannot find entry symbol _start; defaulting to 0000000000000178
collect2: error: ld returned 1 exit status
```

For RISC-V, it's perhaps useful specifying whether it's necessary to emit a warning/error when all input files with an attribute have the same value. Personally I think that's not useful. When GNU ld sees an unknown attribute only in one input, it doesn't give a warning/error.

kito-cheng:
It's mandatory when (tag number % 128) < 64 and optional when (tag number % 128) >= 64.

Fix #352

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1020,8 +1020,16 @@ and name .riscv.attributes. The value of an attribute can hold an integer
 encoded in the uleb128 format or a null-terminated byte string (NTBS).
 The tag number is also encoded as uleb128.
 
-RISC-V attributes have a string value if the tag number is odd and an integer
-value if the tag number is even.
+In order to improve the compatibility of the tool, the attribute follows below rules:
+
+- RISC-V attributes have a string value if the tag number is odd and an integer
+  value if the tag number is even.
+
+- The tag is mandatory;  If the tool does not recognize this attribute and the tag number
+  modulo 128 is less than 64 (`(N % 128) < 64`), errors should be reported.
+
+- The tag is optional; If the tool does not recognize this attribute and the tag number
+  modulo 128 is greater than or equal to 64 (`(N % 128) >= 64`), the tag can be ignored.
 
 ==== Layout of .riscv.attributes section
 


riscv-elf.adoc
```diff
@@ -1017,8 +1017,18 @@ Attributes are encoded in a vendor-specific section of type SHT_RISCV_ATTRIBUTES
...
 encoded in the uleb128 format or a null-terminated byte string (NTBS).
 
-RISC-V attributes have a string value if the tag number is odd and an integer
-value if the tag number is even.
+In order to improve the compatibility of the tool, the attribute follows below rules:
+
+- RISC-V attributes have a string value if the tag number is odd and an integer
+  value if the tag number is even.
+
+- The tag is mandatory; the tool should report errors if the attribute is NOT
```
MaskRay, line 1025: I feel that moving the `if` part to the start of the sentence is easier to read.

MaskRay (approved):

Nelson1225:
Seems like the obj_attrs_handle_unknown function is missing in the risc-v GNU ld when we ported.  The rule looks reasonable and good to me, thanks.

riscv-elf.adoc
kito-cheng: Updated, thanks, asked ChatGPT and he also give me same suggestion :P

kito-cheng:
Got LGTM from two major linker maintainer, gonna merged.

PR #361, opened 2023-01-30
Define relocations for ULEB128 value: R_RISCV_SET_ULEB128 and R_RISCV_SUB_ULEB128

kito-cheng:
ULEB128 has used in DWARF and exception handling table, it used to record value, or record the distance between two symbols.

The later one would be an issue for debug info and exception handling table, since the symbol address might updated during, then the value might incorrect after relaxation, some of those field has alternative encoding type, but some new field defined in DWARF 5 ins't provide alternative format other than ULEB128, e.g. DW_RLE_offset_pair, DW_RLE_startx_length, DW_RLE_startx_endx and DW_RLE_start_length.

This PR basically same as #162, but updated to the trunk, and also document ULEB128 should not shrink the size of the data, since we never know the data used in where and does the length has recorded in somewhere or not, so this would be most safe way.

binutils patch: https://sourceware.org/pipermail/binutils/2020-January/109672.html
LLVM patch: https://reviews.llvm.org/D142879
LLD patch: https://reviews.llvm.org/D142880

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -447,7 +447,11 @@ Description:: Additional information about the relocation
                                             <| `ifunc_resolver(B + A)`
 .2+| 59      .2+| PLT32         .2+| Static  | _word32_          .2+| 32-bit relative offset to a function or its PLT entry
                                             <| S + A - P
-.2+| 60-191  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
+.2+| 60      .2+| SET_ULEB128   .2+| Static  | _ULEB128_         .2+| Local label assignment <<uleb128-note,*note>>
+                                            <| S + A
+.2+| 61      .2+| SUB_ULEB128   .2+| Static  | _ULEB128_         .2+| Local label subtraction <<uleb128-note,*note>>
+                                            <| V - S + A
+.2+| 62-191  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
                                             <|
 .2+| 192-255 .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for nonstandard ABI extensions
                                             <|
@@ -459,6 +463,12 @@ purpose.  These relocations may conflict with other nonstandard extensions.
 This section and later ones contain fragments written in assembler. The precise
 assembler syntax, including that of the relocations, is described in the
 _RISC-V Assembly Programmer's Manual_ <<rv-asm>>.
+
+[[uleb128-note]]
+NOTE: The assembler must allocate enough space to accommodate ULEB128 data for
+`R_RISCV_SET_ULEB128` and `R_RISCV_SUB_ULEB128` relocations. This can be
+achieved either by padding with zeroes to fill up to the XLEN-bits boundary
+or by filling the precomputed result into the relocation location.
 
 ==== Calculation Symbols
 
@@ -501,6 +511,7 @@ of the `__global_pointer$` symbol into register `gp` (aka `x3`).
 | _word16_    | Specifies a 16-bit word
 | _word32_    | Specifies a 32-bit word
 | _word64_    | Specifies a 64-bit word
+| _ULEB128_   | Specifies a variable-length data encoded in ULEB128 format.
 | _wordclass_ | Specifies a _word32_ field for ILP32 or a _word64_ field for LP64
 | _B-Type_    | Specifies a field as the immediate field in a B-type instruction
 | _CB-Type_   | Specifies a field as the immediate field in a CB-type instruction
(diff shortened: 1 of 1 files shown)


kito-cheng:
Update the link of LLVM + LLD patch.

riscv-elf.adoc
```diff
@@ -1157,6 +1162,12 @@ are commonly resolved at compile-time, such as intra-function jumps), code
 generators must in general ensure that relocations are always emitted when
 relaxation is enabled.
 
+Linker should adjust relocations which refer to those symbol has updated the
+address.
+
+ULEB128 value with relocation must be padding to the same length even if the
```
rui314, line 1168: How do you know the original length?

kito-cheng: From my LLD implementation experience: read the value before apply the any relocation on that position, that's remind me I should add one more note, linker should report error if the new value is longer than the old value, otherwise linker will corrupt other data other than ULEB128 data.

rui314: Then we need to make it clear that the compiler/assembler writes a dummy value such as 0 to the relocated place in a possibly redundantly encoded form so that the linker can read it to know the expected length.

MaskRay:
> Deifne relocation for ULEB128 value 

Define. Suggest: just specify the full names of the two new relocation types.

riscv-elf.adoc
kito-cheng: Updated, thanks for suggestion :)

kito-cheng:
Update:
- Update title:
- Add note to mention assembler should preserve space for ULEB128 data
- Mention linker should report error if data size is extended

kito-cheng:
Changes:
- Rebased
- Change number to 60, 61 due to R_RISCV_PLT32 has added.

kito-cheng:
@rui314 @MaskRay do you mind review it again if it's look good to you? :)

riscv-cc.adoc
```diff
@@ -235,6 +235,12 @@ __attribute__((__packed__)) { int i; double d }+` are treated the same, as are
 `struct { float f; float g; }` and `+struct { float f; float g __attribute__
 ((aligned (8))); }+`.
 
+NOTE: One exceptional case for the flattening rule is the array with empty
+structs or unions; C language treats it as an empty field, but {Cpp} language
+treats it as a non-empty field since {Cpp} defines the size of an empty struct
+or union as 1. However, C defines it as 0, that result from the array with an
+empty struct or union having a different size between C and {Cpp}.
+
```
rui314, line 243: Isn't this an unrelated change?

rui314 (commented):
Overall looking good.

riscv-elf.adoc
```diff
@@ -460,6 +464,12 @@ This section and later ones contain fragments written in assembler. The precise
 assembler syntax, including that of the relocations, is described in the
 _RISC-V Assembly Programmer's Manual_ <<rv-asm>>.
 
+[[uleb128-note]]
+NOTE: Assembler must preserve enough space for ULEB128 data with
+R_RISCV_SET_ULEB128 and R_RISCV_SUB_ULEB128 relocation for the linker;
```
MaskRay, line 469: The assembler must allocate enough space to accommodate ULEB128 data for R_RISCV_SET_ULEB128 and R_RISCV_SUB_ULEB128 relocations. This can be achieved either by padding with zeroes to fill up to the XLEN-bits boundary or by filling the precomputed result into the relocation location.

riscv-elf.adoc
```diff
@@ -1221,6 +1232,14 @@ are commonly resolved at compile-time, such as intra-function jumps), code
 generators must in general ensure that relocations are always emitted when
 relaxation is enabled.
 
+Linker should adjust relocations which refer to those symbol has updated the
```
MaskRay, line 1235: Linkers should adjust relocations that refer to symbols whose addresses have been updated.

MaskRay (approved):

riscv-elf.adoc
kito-cheng: Applied, thank :)

riscv-elf.adoc
```diff
@@ -447,7 +447,11 @@ Description:: Additional information about the relocation
                                             <| `ifunc_resolver(B + A)`
 .2+| 59      .2+| PLT32         .2+| Static  | _word32_          .2+| 32-bit relative offset to a function or its PLT entry
                                             <| S + A - P
-.2+| 60-191  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
+.2+| 60      .2+| SET_ULEB128   .2+| Static  | _ULEB128_         .2+| Local label assignment <<uleb128-note,*note>>
+                                            <| S + A
+.2+| 61      .2+| SUB_ULEB128   .2+| Static  | _ULEB128_         .2+| Local label subtraction <<uleb128-note,*note>>
+                                            <| V - S + A
```
rui314, line 453: Actually I think this needs to be `V - S - A` for consistency with other SUB relocations.

kito-cheng: Good catch! it actually should be `V - (S + A)` but `V -S - A` would be consistency with others.

kito-cheng: hot-fix applied: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/commit/4f7de692724302a6d239ef7ad4d818d98ac34368

PR #362, opened 2023-01-30
Open the relocation number more than 255

kito-cheng:
ELF64 format has support up to 32 bits for the relocation type, but currenlty we only define lower 8 bits due to the compatibility with ELF32, so the idea is open those range to RV64, and add RV64 only relocation to those range.

Also reserved part of relocation number to non-standard use.

We open up to 4096 and then still reserve the upper 20-bits for future extension, which might use those bits other than identifying the relocation type in the future, like SPARC64.

`RV64 only` might not precise enough, but enough for now since we don't have RV64 with ilp32*, this should fix when adding that.

---

NOTE: Updated, open up to 4096 rather than 2^32.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -449,14 +449,23 @@ Description:: Additional information about the relocation
                                             <|
 .2+| 192-255 .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for nonstandard ABI extensions
                                             <|
+.2+| 256-2047 (0x100-0x7ff) .2+| *Reserved*           .2+| -       |                   .2+| Reserved for standard ABI extensions (RV64 only)
+                                            <|
+.2+| 2048-4095  (0x800-0xfff) .2+| *Reserved*         .2+| -       |                   .2+| Reserved for nonstandard ABI extensions (RV64 only)
+                                            <|
 |===
 
-Nonstandard extensions are free to use relocation numbers 192-255 for any
-purpose.  These relocations may conflict with other nonstandard extensions.
+Nonstandard extensions are free to use relocation numbers 192-255 and 2048-4095
+for any purpose.  These relocations may conflict with other nonstandard
+extensions.
 
 This section and later ones contain fragments written in assembler. The precise
 assembler syntax, including that of the relocations, is described in the
 _RISC-V Assembly Programmer's Manual_ <<rv-asm>>.
+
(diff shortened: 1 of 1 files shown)


rui314:
I don't think we realistically don't need that many types of relocations. Instead of allowing up to 2^31 different types of relocations, can't we accept only up to, say, 4096? That number should still be more than enough.

As far as I know, SPARC64 has an additional data member in the upper 24 bits of the "type" field (https://github.com/rui314/mold/blob/6caf0867b0fb63aee7b0347950493a486517e5b9/elf/elf.h#L1641). Using all the available bits now makes it impossible to do something similar to that in the future.

rui314:
Apologies, I accidentally closed this PR. Reopened it.

kito-cheng:
I agree the relocation allow up to 2^31 types is kind of unrealistically...or not so useful, and reserve upper bit for future extension other than interpret as relocation type sounds a interesting idea, 4096 should be a reasonable upper bound and split half to non-standard use still more than enough.

kito-cheng:
@rui314 updated the PR, thanks for the suggestion :) I don't know the SPARC64 has such strange usage of those bits before.

rui314:
Yup. One idea in my mind is to redefine an unused upper bit of the "type" field as the "relaxable" bit, so that we don't spend a whopping 24 bytes just for indicating that the next relocation is relaxable. The current scheme with the `R_RISCV_RELAX` type is very inefficient to encode just one bit of data.

kito-cheng:
It might not workable if we consider RV32, and that's definitely a backward-incompatible change, other than that I think it's good idea.

rui314:
I think declaring that the upper 20 bits of `r_type` is not really a part of `r_type` and reserved for future use is _not_ a breaking change until we actually define a new field in the reserved bits. Declaring as such makes it possible to add code to existing toolchains to ignore the upper 20 bits when reading `r_type` field, which will work as a preparation for a breaking change.

kito-cheng:
@MaskRay has proposal for RELLEB/CREL, I think this will be the right way to remove the limit of relocation type, although it may still need take few more time to cook, but we don't really urgent to extend the limit yet, so gonna close this issue.

https://maskray.me/blog/2024-03-09-a-compact-relocation-format-for-elf
https://groups.google.com/g/generic-abi/c/yb0rjw56ORw

PR #363, opened 2023-01-31
Define PLT-generating PC-Relative data relocation

PiJoules:
We would like to introduce a new relocation type that works similarly to x86’s R_X86_64_PLT32 relocation and Arm’s R_AARCH64_PLT32 relocation.

In the “Procedure Calls” section of riscv-elf.adoc, we would add a new type “R_RISCV_PLT32” that follows the existing wording for the “R_RISCV_CALL” and “R_RISCV_CALL_PLT” types, but rather than modifying an instruction, it fills a whole, naturally aligned, 32-bit data location with the signed byte offset from its own location to the PLT entry for the target symbol. If the relocation is within +/-2GB of the target symbol, the linker may relax this to a R_RISCV_32_PCREL.

The purpose of this relocation is to enable the use of PIC-friendly ABI on RISC-V. Example is the [relative vtables ABI for C++](https://youtu.be/9HGKlDiJy8E). This ABI saves memory by moving vtables into .rodata by replacing each of the relocations that would constitute the vtable (R_RISCV_64, R_AARCH64_ABS64, R_X86_64_64) with PC-relative offsets, requiring no dynamic relocations. Instead of placing a pointer to each virtual method into the vtable, we would instead take the offset between the vtable and the virtual function which can be statically computed if the compiler knows both symbols are DSO-local. In the event the symbol is defined elsewhere, the vtable slots will instead contain offsets to the PLT entries for these functions. On both x86_64 and aarch64, Clang lowers the IR down to the offset between the PLT entry and the vtable, generating the PLT32 reloc we want, but we do not see this on riscv64 since the only PLT-generating relocs are `R_RISCV_CALL[_PLT]` which can only be applied to specific instructions (AUIPC+JALR) and not to data locations.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -445,7 +445,9 @@ Description:: Additional information about the relocation
                                             <| S + A - P
 .2+| 58      .2+| IRELATIVE     .2+| Dynamic | _wordclass_       .2+| Relocation against a non-preemptible ifunc symbol
                                             <| `ifunc_resolver(B + A)`
-.2+| 59-191  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
+.2+| 59      .2+| PLT32         .2+| Static  | _word32_          .2+| 32-bit relative offset to a function or its PLT entry
+                                            <| S + A - P
+.2+| 60-191  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
                                             <|
 .2+| 192-255 .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for nonstandard ABI extensions
                                             <|


riscv-elf.adoc
```diff
@@ -445,7 +445,9 @@ Description:: Additional information about the relocation
                                             <| S + A - P
 .2+| 58      .2+| IRELATIVE     .2+| Dynamic | _wordclass_       .2+| Relocation against a non-preemptible ifunc symbol
                                             <| `ifunc_resolver(B + A)`
-.2+| 59-191  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
+.2+| 59      .2+| 32_PLT        .2+| Static  | _word32_          .2+| 32-bit PC relative to a function
```
rui314, line 448: This relocation name is somewhat irregular. As far as I know, PLT-generating 32-bit relocation is called `R_<ARCH>_PLT32` in psABIs for x86-64, i386, ARM64, ARM32, PPC32, SPARC, s390x, m68k and SH-4.

riscv-elf.adoc
```diff
@@ -1219,6 +1223,43 @@ NOTE: Using address of PLT stubs of the target symbol or address target symbol
...
+[,asm]
+----
+    .hidden func
+func:
+
+A:
+    .word  func - .   # R_RISCV_32_PCREL
+----
+--
+
```
rui314, line 1262: I don't think you need to define this as a relaxation because this is what `R_RISCV_CALL` and `R_RISCV_CALL_PLT` do. Despite their names, they don't generate a PLT entry if the symbol is resolved at link-time. I'd define `R_RISCV_PLT32` as "the same as `R_RISCV_CALL` except that it sets an entire 32-bit value."

rui314 (commented):
Generally it sounds like a reasonable addition to the psABI.

riscv-elf.adoc
PiJoules: Yeah. My intention was to stay consistent with `R_RISCV_32_PCREL` which had the 32 first, but I can change it to PLT32 here.

riscv-elf.adoc
PiJoules: SG. I'll remove this part. I wasn't sure if the idea of not generating a PLT entry was already implied in the standard which is why I added this.

riscv-elf.adoc
MaskRay: Suggest: `32-bit offset relative to a function or its PLT entry`

MaskRay:
The main description needs to say that the target symbol has type `STT_FUNC`.

MaskRay:
I feel that `R_RISCV_PLT32` may be a better name. The current name tries to follow `R_RISCV_32_PCREL` but the latter is an anomaly instead of common (see `SUB*`, `SET*`, `LO*`).

riscv-elf.adoc
PiJoules: Done

PiJoules:
> The main description needs to say that the target symbol has type STT_FUNC.
Updated.

jrtc27:
PLT32 is a bad name for this, to me that reads as being the _absolute_ address of the PLT entry, but that's what other architectures went with...

Re your description, R_RISCV_CALL_PLT isn't the only relocation that can generate a PLT. R_RISCV_CALL does too for LLD, and it sometimes does for GNU ld (but has really weird semantics that mean you should never use it).

riscv-elf.adoc
```diff
@@ -445,7 +445,9 @@ Description:: Additional information about the relocation
                                             <| S + A - P
 .2+| 58      .2+| IRELATIVE     .2+| Dynamic | _wordclass_       .2+| Relocation against a non-preemptible ifunc symbol
                                             <| `ifunc_resolver(B + A)`
-.2+| 59-191  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
+.2+| 59      .2+| PLT32        .2+| Static  | _word32_          .2+| 32-bit offset relative to a function or its PLT entry
```
jrtc27, line 448: 1. Columns don't line up with the rest above
2. "relative offset"

riscv-elf.adoc
```diff
@@ -473,6 +475,7 @@ calculation:
 | B         | Base address of a shared object loaded into memory
 | G         | Offset of the symbol into the GOT (Global Offset Table)
 | GOT       | Address of the GOT (Global Offset Table)
+| L         | Represents the place (section offset or address) of the Procedure Linkage Table entry for a symbol
```
jrtc27, line 478: Hm, CALL(_PLT) just use S, but can still end up at a PLT entry...

riscv-elf.adoc
```diff
@@ -613,7 +616,10 @@ pairs of instructions (`AUIPC+JALR`) generated by the `CALL` or `TAIL`
 pseudoinstructions.  Originally, these relocations had slightly different
 behavior, but that has turned out to be unnecessary, and they are now
 interchangeable,  `R_RISCV_CALL` is deprecated, suggest using `R_RISCV_CALL_PLT`
-instead.
+instead. The `R_RISCV_PLT32` relocation is a data relocation for calculating
```
jrtc27, line 619: Why is this here?

riscv-elf.adoc
PiJoules: Adding `L` was a suggestion in an offline discussion since x86 uses it to describe the plt. For consistency with `CALL_[PLT]` would you suggest either updating CALL to use `L` or just using `S` here?

PiJoules:
> Re your description, R_RISCV_CALL_PLT isn't the only relocation that can generate a PLT. R_RISCV_CALL does too for LLD, and it sometimes does for GNU ld (but has really weird semantics that mean you should never use it).

Sorry, poor wording on my end. I kinda group them interchangeably since they're very similar. Updated the description.

riscv-elf.adoc
PiJoules: Woops. Fixed

riscv-elf.adoc
rui314: We should not add `L` because it causes more confusion than without it. As I wrote before, PLT-generating relocations are not guaranteed to create a PLT entry, so "the place of the PLT entry for a symbol" may not exist.

riscv-elf.adoc
rui314: I think this description is not necessary. The expression (`S + A - P`) and the relocation size (`word32`) define what this relocation is.

MaskRay:
> PLT32 is a bad name for this, to me that reads as being the _absolute_ address of the PLT entry, but that's what other architectures went with...

I agree `PLT32` is a poor name (and x86 made the bad choice). When `R_AARCH64_PLT32` was added, `PLT32` was chosen because of the precedent of `R_{386,X86_64}_PLT32`. As `PLT32` fixes into RISC-V's relocation name scheme, it seems good to stick with the established convention now...

kito-cheng:
Few questions:
- Should we need 64 bit variant? something like R_RISCV_PLT64? I saw AArch64 and x86_64 only provide 32 version, so I guess we only need PLT32?
- Its require compiler, assembler, linker works, and also some library works:
  - LLVM and LLD seems already support that, so we just need to reference how AArch64 implement and implement that into RISC-V? https://reviews.llvm.org/D81446 and https://reviews.llvm.org/D81446
  - libcxxabi already support that, so we just need to rebuild that with clang + `-fexperimental-relative-c++-abi-vtables` flag?

And psABI changes require toolchain PoC, either LLVM toolchain or GNU toolchain, not sure you intend to provide that or need some help from community?

riscv-elf.adoc
PiJoules: Removed this description

riscv-elf.adoc
PiJoules: Removed `L`

PiJoules:
> Should we need 64 bit variant? something like R_RISCV_PLT64? I saw AArch64 and x86_64 only provide 32 version, so I guess we only need PLT32?

Yeah for our use case at least, only the 32-bit variant is needed. I suppose I could also propose other sized versions in this, but I'm not sure if they should be added now since I don't know if there's any immediate need for them now.

> LLVM and LLD seems already support that, so we just need to reference how AArch64 implement and implement that into RISC-V? https://reviews.llvm.org/D81446 and https://reviews.llvm.org/D81446

For the most part, yeah.

> libcxxabi already support that, so we just need to rebuild that with clang + -fexperimental-relative-c++-abi-vtables flag?

This technically doesn't require any library support, but it's when using relative vtables that this relocation will be needed. libcxx[abi] that uses relative vtables will definitely need this.

> And psABI changes require toolchain PoC, either LLVM toolchain or GNU toolchain, not sure you intend to provide that or need some help from community?

I was going to provide patches. I wasn't sure if WIP patches were required for submitting, but it looks like this answers that.

PiJoules:
https://reviews.llvm.org/D143115 has the LLVM + LLD changes which I'm able to use to build a fuchsia toolchain (that uses this reloc via relative vtables) for riscv.

riscv-elf.adoc
```diff
@@ -445,7 +445,9 @@ Description:: Additional information about the relocation
                                             <| S + A - P
 .2+| 58      .2+| IRELATIVE     .2+| Dynamic | _wordclass_       .2+| Relocation against a non-preemptible ifunc symbol
                                             <| `ifunc_resolver(B + A)`
-.2+| 59-191  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
+.2+| 59      .2+| PLT32        .2+| Static   | _word32_          .2+| 32-bit relative offset to a function or its PLT entry
```
rui314, line 448: nit: `.2+| Static` is one column off from other lines

rui314 (commented):
I don't have further comments other than this.

MaskRay:
> > The main description needs to say that the target symbol has type STT_FUNC.
> 
> Updated.

The latest diff appears to drop this paragraph?

rui314:
> > > The main description needs to say that the target symbol has type STT_FUNC.
> 
> > 
> 
> > Updated.
> 
> 
> 
> The latest diff appears to drop this paragraph?

We don't have that paragraph for CALL and CALL_PLT, too, so I think that should be handled as a separate issue.

riscv-elf.adoc
PiJoules: Fixed

PiJoules:
Any outstanding concerns blocking from merging this?

rui314:
LGTM

MaskRay (approved):

Nelson1225:
Since we don't have any data relocation to store the offset between pc (vtable) and the corresponding function or plt entry, this new data relocation looks reasonable to me.  Besides, x86 and aarch64 already have the similar support, so it seems like a right way to go.

kito-cheng (approved):
Seem like we gather LGTMs from all major linker experts and also implementation provided, just wait one more week to make sure no further comments.

kito-cheng:
Thanks everyone :)

PR #365, opened 2023-02-10
Add note for array with empty struct or union

### LINKED ISSUE #358: Clarification of rules for flattening structs containing arrays of empty records
The hard FP calling convention currently states "Fields containing empty structs or unions are ignored while flattening, even in C++, unless they have nontrivial copy constructors or destructors." It appears that g++ _doesn't_ ignore an array of empty records in a struct. If this is indeed the desired behaviour, it would be good to document it.

```
$ cat t.c
struct empty { struct { struct { } e; }; };
struct s { struct empty e[1]; float f; };

float test_s(struct s a) {
  return a.f+1.0;
}
```

In the above example, the float is passed in a0 with g++ but fa0 with gcc.

kito-cheng:
Fix #358

The key point of this issue is C and C++ having different type size for array of empty struct or union, size of empty struct or union is 0 for C, and 1 for C++, and the size of array is type size multiplied by the length of the array.

Also checked AArch64 code gen, they also generated different code between C and C++, and ABI isn't explicitly describe this but has describe the size of array is type * length.

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -270,13 +270,26 @@ hierarchy flattened, including any array fields.  That is, `struct { struct
 { float f[1]; } a[2]; }` and `struct { float f0; float f1; }` are
 treated the same.  Fields containing empty structs or unions are ignored while
 flattening, even in {Cpp}, unless they have nontrivial copy constructors or
-destructors.  Fields containing zero-length bit-fields are ignored while
-flattening.  Attributes such as `aligned` or `packed` do not interfere with a
-struct's eligibility for being passed in registers according to the rules
-below, i.e. `struct { int i; double d; }` and `+struct
+destructors.  Fields containing zero-length bit-fields or zero-length arrays are
+ignored while flattening.  Attributes such as `aligned` or `packed` do not
+interfere with a struct's eligibility for being passed in registers according
+to the rules below, i.e. `struct { int i; double d; }` and `+struct
 __attribute__((__packed__)) { int i; double d }+` are treated the same, as are
 `struct { float f; float g; }` and `+struct { float f; float g __attribute__
 ((aligned (8))); }+`.
+
+NOTE: One exceptional case for the flattening rule is an array of empty
+structs or unions; C treats it as an empty field, but {Cpp}
+treats it as a non-empty field since {Cpp} defines the size of an empty struct
+or union as 1. i.e. for `struct { struct {} e[1]; float f; }` as the first
+argument, C will treat it like `struct { float f; }` and pass `f` in `fa0` as
+described below, whereas {Cpp} will pass the pass the entire aggregate in `a0` 
+(XLEN = 64) or `a0` and `a1` (XLEN = 32), as described in the integer calling
+convention.
+Zero-length arrays of empty structs or union will be
+ignored for both C and {Cpp}. i.e. For `struct { struct {} e[0]; float f; };`,
+as the first argument, C and {Cpp} will treat it like `struct { float f; }`
+and pass `f` in `fa0` as described below.
 
 A real floating-point argument is passed in a floating-point argument
 register if it is no more than ABI_FLEN bits wide and at least one floating-point


asb:
Thanks for looking at this Kito. Quick high level comment: I think the case of a zero-length array of empty structs/unions should also be mentioned. Perhaps it's worth being more explicit about the case of a 1-sized array of an empty struct/union too?

kito-cheng:
Changes:
- Add example to show the case `struct { struct {} e[1]; float f; };`
- Add sentence to mention zero-length array will ignored during struct flatten.

riscv-cc.adoc
```diff
@@ -227,14 +227,22 @@ hierarchy flattened, including any array fields.  That is, `struct { struct
 { float f[1]; } g[2]; }` and `struct { float f; float g; }` are
 treated the same.  Fields containing empty structs or unions are ignored while
 flattening, even in {Cpp}, unless they have nontrivial copy constructors or
-destructors.  Fields containing zero-length bit-fields are ignored while
-flattening.  Attributes such as `aligned` or `packed` do not interfere with a
-struct's eligibility for being passed in registers according to the rules
-below, i.e. `struct { int i; double d; }` and `+struct
+destructors.  Fields containing zero-length bit-fields or zero-length array are
```
nick-knight, line 230: Grammatical suggestion:

```suggestion
destructors.  Fields containing zero-length bit-fields or zero-length arrays are
```

riscv-cc.adoc
```diff
@@ -227,14 +227,22 @@ hierarchy flattened, including any array fields.  That is, `struct { struct
...
+ignored while flattening.  Attributes such as `aligned` or `packed` do not
+interfere with a struct's eligibility for being passed in registers according
+to the rules below, i.e. `struct { int i; double d; }` and `+struct
 __attribute__((__packed__)) { int i; double d }+` are treated the same, as are
 `struct { float f; float g; }` and `+struct { float f; float g __attribute__
 ((aligned (8))); }+`.
 
+NOTE: One exceptional case for the flattening rule is the array with empty
```
nick-knight, line 238: Minor English suggestion:

```suggestion
NOTE: One exceptional case for the flattening rule is an array of empty
```

riscv-cc.adoc
```diff
@@ -227,14 +227,22 @@ hierarchy flattened, including any array fields.  That is, `struct { struct
...
+structs or unions; C language treats it as an empty field, but {Cpp} language
```
nick-knight, line 239: Minor English suggestion:

```suggestion
structs or unions; C treats it as an empty field, but {Cpp}
```

riscv-cc.adoc
```diff
@@ -227,14 +227,22 @@ hierarchy flattened, including any array fields.  That is, `struct { struct
...
+treats it as a non-empty field since {Cpp} defines the size of an empty struct
+or union as 1. However, C defines it as 0, that result from the array with an
+empty struct or union having a different size between C and {Cpp}. i.e.
```
nick-knight, line 242: Delete redundant sentence and s/i.e./e.g.

```suggestion
or union as 1. E.g.,
```

riscv-cc.adoc
```diff
@@ -227,14 +227,22 @@ hierarchy flattened, including any array fields.  That is, `struct { struct
...
+`struct { struct {} e[1]; float f; };` C will passed by `f0`, but {Cpp} will
+passed by `a0`
+
```
nick-knight, line 245: `f0` is `ft0`: I think you mean `fa0`. 

Also, we're not using a fixed-width font for register mnemonics elsewhere, so I suggest not introducing it here and now.

Also a minor English suggestion.

```suggestion
for `struct { struct {} e[1]; float f; };`, C will pass `f` in fa0, but {Cpp} will
pass it in `a0`.
```

nick-knight: My experiment on LLVM for this example is showing the same result for C and C++. But having just embarrassed myself over at #366 I am very hesitant to say any more.

nick-knight (changes requested):

riscv-cc.adoc
asb: I opened #358 as it seemed Clang/LLVM wasn't matching GCC. My patch https://reviews.llvm.org/D142327 _should_ match the clarification in this PR but it hasn't been committed yet.

nick-knight: Thanks @asb . I think I understand how things work. I've updated my suggestion above to clarify some of my confusions.

kito-cheng:
@nick-knight thanks for your comment and suggestion :)

kito-cheng:
@asb @jrtc27 @luismarques did you mind take a review for that? I would like to have your review before I merge that :)

luismarques (approved):

luismarques:
Sorry, I shouldn't have approved this. What is there LGTM but it doesn't explicitly cover all cases that @asb was asking about.

kito-cheng:
@luismarques  added more description for zero-length array of empty structs/union, thank for the review :)

kito-cheng:
Changes:

- Explicitly say the field with array of empty struct type will be ignored or not in the example.

riscv-cc.adoc
asb: This was marked resolved but doesn't seem to have been integrated.

asb:
@kito-cheng the last round of grammatical suggestions from @nick-knight are marked as resolved but the changes don't seem to be reflected in the current version of the patch. It looks like they might have got lot along the way?

riscv-cc.adoc
```diff
@@ -227,14 +227,27 @@ hierarchy flattened, including any array fields.  That is, `struct { struct
...
 
+NOTE: One exceptional case for the flattening rule is an array of empty
+structs or unions; C treats it as an empty field, but {Cpp}
+treats it as a non-empty field since {Cpp} defines the size of an empty struct
+or union as 1. i.e. for `struct { struct {} e[1]; float f; }` as the first
+argument, C will treat it like `struct { float f; }` and pass `f` in `fa0` as
+described below, whereas {Cpp} will pass the pass the entire aggregate in `a0` 
+(XLEN = 64) or `a0` and `a1` (XLEN = 32), as described in the integer calling
+convention.
+The zero-length arrays of empty structs or union will treat be
```
asb, line 246: The zero-length => Zero-length
will treat be => will be

riscv-cc.adoc
```diff
@@ -227,14 +227,27 @@ hierarchy flattened, including any array fields.  That is, `struct { struct
...
+ignored for bot C and {Cpp}. i.e. For `struct { struct {} e[0]; float f; };`,
```
asb, line 247: bot => both

asb (approved):
LGTM modulo the minor wording fixes suggested in line. Thanks Kito!

PR #366, opened 2023-02-16
Fix flattening example

nick-knight:
I've been confused for several years what it means to "flatten" an array, due to this example. I finally bothered to type it into a compiler, and now think the example is wrong.

I was inspired to look at this by recent discussion over at #365.

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -224,7 +224,7 @@ floating-point registers than the ABI.
 
 For the purposes of this section, "struct" refers to a C struct with its
 hierarchy flattened, including any array fields.  That is, `struct { struct
-{ float f[1]; } g[2]; }` and `struct { float f; float g; }` are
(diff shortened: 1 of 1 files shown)


asb:
I think you're right this is an error, but perhaps changing `g[2]` to `g[1]` would make it a better example, as it shows an examples of cases that will be passed according to the special case ABI rules being discussed.

jrtc27:
The case is correct, it's an array of two structs, each of which contains an array of a single float. It's just confusing because the f and g don't correlate on either side.

jrtc27:
Renaming it to be less confusing:
> [...] `struct { struct { float f[1]; } a[2]; }` and `struct { float f0; float f1; }` are [...]

asb:
Oh good, I was worried I'd missed (and perhaps introduced?) a mistake via typo here. I like Jessica's suggested renaming.

nick-knight:
:facepalm: Sorry team.

PR #367, opened 2023-03-01
Add ABI information for __bf16

asb:
At this point, posted just for comment. We'll need something covering this with the upcoming [RISC-V bfloat16 extension](https://github.com/riscv/riscv-bfloat16). I think this is all that is needed. For reference, see the [equivalent change in the x86-64 psABI](https://gitlab.com/x86-psABIs/x86-64-ABI/-/merge_requests/35/diffs).

I haven't carefully followed `__fp16` vs `_Float16` and `_BFloat16` vs `__bf16`. It looks like `__bf16` is preferred on other targets. Though this would also be a good moment to discuss that further if people feel that's not the right choice for RISC-V.

My understanding is that `_Float16` and `__bf16` will be passed in FPRs in exactly the same cases an FPR would be used for a `float` parameter or struct field. Again - please speak up if you feel that isn't what the psABI states.

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -449,6 +449,7 @@ alignments (based on the ILP32 convention):
 | long                 |  4            |  4
 | long long            |  8            |  8
 | void *               |  4            |  4
+| +++__bf16+++         |  2            |  2
 | _Float16             |  2            |  2
 | float                |  4            |  4
 | double               |  8            |  8
@@ -475,6 +476,7 @@ alignments (based on the LP64 convention):
 | long long            |  8            |  8
 | +++__int128+++       | 16            | 16
 | void *               |  8            |  8
+| +++__bf16+++         |  2            |  2
 | _Float16             |  2            |  2
 | float                |  4            |  4
 | double               |  8            |  8
@@ -501,6 +503,8 @@ arguments are either `0` (`false`) or `1` (`true`).
 A null pointer (for all types) has the value zero.
 
 `_Float16` is as defined in the C ISO/IEC TS 18661-3 extension.
+
+`__bf16` has the same parameter passing and return rules as for `_Float16`.
 
 `_Complex` types have the same layout as a struct containing two fields of the
 corresponding real type (`float`, `double`, or `long double`), with the first
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -145,7 +145,8 @@ the _Itanium {Cpp} ABI_ <<itanium-cxx-abi>>;
 there are no RISC-V specific mangling rules.
 
 See the "Type encodings" section in _Itanium {Cpp} ABI_
-for more detail on how to mangle types.
+for more detail on how to mangle types. Note that `__bf16` is mangled in the
+same way as `std::bfloat16_t`.
 
 == ELF Object Files
 


nick-knight:
> I haven't carefully followed `__fp16` vs `_Float16` and `_BFloat16` vs `__bf16`. It looks like `__bf16` is preferred on other targets. Though this would also be a good moment to discuss that further if people feel that's not the right choice for RISC-V.

I have a strong preference to `_Float16` over `__fp16` because the former will (almost certainly) be part of C23 (annex H), while the latter has a somewhat [sordid legacy](https://gcc.gnu.org/onlinedocs/gcc/Half-Precision.html). And I think it makes sense to prefer an "interchange format" (IEEE-754 terminology) in the context of specifying an ABI calling convention. 

Regarding `_Bfloat16` vs. `__bf16`, I don't have a strong preference (yet). The x86-64 ABI folks had a similar discussion and [settled on `__bf16`](https://groups.google.com/g/x86-64-abi/c/FHNHJYuKBMM). AFAIK, C23 is not planning to introduce a "`_Bfloat16`", although C++23 is proposing `std::bfloat16_t`. I'm slightly uncomfortable referencing a type that isn't yet specified in the C language, or our RISC-V C-language extensions, so I'd suggest adding some comment to indicate that what `__bf16` actually means is still under deliberation.

kito-cheng:
Unlike `__fp16` vs `_Float16`, the latter one has a spec (`ISO/IEC TS 18661-3:2015`) to specify the type and semantic, and it also will included into the C lang spec as @nick-knight mention, so `__fp16`  vs `_Float16` is less controversial since RISC-V are relative new target, we didn't formally support `__fp16` (clang has support by *default* that as clang extension), so it easier for us.

But `__bf16` vs `_Bfloat16` is different situation, no ISO spec defined `_Bfloat16` yet, so IMO we should align to other target like x86 and aarch64 if possible to prevent the unnecessary porting effort for those codebase which already use `__bf16`.

So the PR generally LGTM, but I would defer the formal LGTM until we've PoC implementation posted :)

NOTE: actually I was preferring `__fp16` before, see https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/172 , https://lists.llvm.org/pipermail/cfe-dev/2021-March/067847.html :P

kito-cheng:
@Liaoshihua good catch, thanks for raising that issue!

That's the list for related discussion I collected so far (not complete list but I put most important 3 here):
- [AArch64 bfloat16 mangling (GCC mailing list)](https://gcc.gnu.org/pipermail/gcc-patches/2023-February/611099.html)
- [[X86][RFC] Change mangle name of __bf16 from u6__bf16 to DF16b (LLVM)](https://reviews.llvm.org/D136919)
- [Add mangling for _FloatNx and bfloat16_t (itanium-cxx-abi/cxx-abi)](https://github.com/itanium-cxx-abi/cxx-abi/pull/147)

kdockser:
While BFloat16 (BF16) is not a standard format specified in IEEE-754, it is a floating-point format as defined by the standard. As such, it is characterized by its radix (2), precision (8-bit mantissa) and its exponent rage (emax=127). It follows all of the rules for IEEE conformance.

In RISC-V (and many other architectures), BF16 is both an _interchange_ and an _arithmetic format_.
Some architectures or extensions may choose to specify simplifications such as flushing subnormals to zero. While this is not IEEE compliant, this has been done on many IEEE standard formats.
In RISC-V, 16-bit floating-point values that are held in FPRs are NaN boxed. The least-significant 16 bits of the register hold the encoded value whether it is FP16 or BF16. In RISC-V vector, these values are held in 16-bit elements.

asb:
Thank you for pointing out the various discussions about mangling. Currently the RISC-V psABI doesn't specify its own mangling rules and instead points to the Itanium C++ ABI. I think it would make sense to maintain this property, so perhaps his PR is blocked until https://github.com/itanium-cxx-abi/cxx-abi/pull/147 is resolved?

kito-cheng:
@asb Yes, I would prefer wait https://github.com/itanium-cxx-abi/cxx-abi/pull/147 get resolved, at least for the `std::bfloat16_t` part, and I saw you've suggested them splited out `std::bfloat16_t`, thanks :)

asb:
I'd like to question one thing that's come up in these discussions - both your psABI meeting slides and Ken's response indicate that bf16 is both an interchange and an arithmetic format on RISC-V. Is that really the case though? Zfbfmin only defines load/store and conversion instructions.

kdockser:
Yes, it is an arithmetic format too; the proposed Zvfbfwma extension defines a widening floating-point fused multiply-add.

[Draft BF16 20230322 for AR](https://github.com/riscv/riscv-bfloat16/releases/tag/20230322) was released to the Architectural Review committee as the first step towards freeze.

asb:
> Yes, it is an arithmetic format too; the proposed Zvfbfwma extension defines a widening floating-point fused multiply-add.
> 
> [Draft BF16 20230322 for AR](https://github.com/riscv/riscv-bfloat16/releases/tag/20230322) was released to the Architectural Review committee as the first step towards freeze.

Yes, but that is of course a very limited set of operations compared to 32-bit floats, and a system might only implement Zfbfmin. Different groups are perhaps using differing terminology here, but on some of the linked issues the "storage format" vs "arithmetic format" distinction appears is used with respect to whether you can perform arithmetic operations on `__bf16`. While we could support such operations, it would in general require promotion to a wider float type or a libcall.

kdockser:
I would argue that Zvfbfwma does operate on BF16 in that it consumes values in this format.
Perhaps more significant is the call within RISC-V for non-widening BF16 arithmetic instructions. This is a hot topic in the Floating-Point SIG.

kito-cheng:
After reading more material, I prefer to define bf16 as an interchange and an arithmetic format, even NO HW instructions are provided.

Let me use x86's ABI bf16 motivation[1] here:

>Currently __bfloat16 is a typedef of short, which creates a problem where the compiler does not raise any alarms if it is used to add, subtract, multiply or divide, but the result of the calculation is actually meaningless.
>
>To solve this problem, a real scalar type __Bfloat16 needs to be introduced. It is mainly used for intrinsics, not available for C standard operators. __Bfloat16 will also be used for movement like passing parameter, load and store, vector initialization, vector shuffle, and .etc. It creates a need for a corresponding psABI.

For the same reason as x86, we should define _bf16 and it could also be used as C++23 std::bfloat16 type.

So here are questions about arithmetic, should we provide _bf16 operation libcall? Let's break that into several parts:
- Promotion from BF16 to F16 is cheap - just a shift left by 16.
- Truncation from F16 to BF16 might be more complex than the promotion operation since it needs to consider the rounding mode, which should go through the libcall if no bf16 extension is enabled.
- Operations: According to the C++23 spec, the spec doesn't specify the accuracy according to P1467[3]; it's valid to provide more accurate results, so promoting to float32 and then do the operation is a valid way, and actually, we already did that for _Float16 when no zfh.

One concern might be that the clang/LLVM is forbidden any operations on __bf16. Still, I believe it will relax soon since C++23 allows operation on std::bfloat16 unless clang/LLVM defines a different underlying type for std::bloat16.

And also GCC already allow the operation for __bf16, and provide libgcc function for bfloat16 type.

So, in summary - defining bf16 as an interchange and an arithmetic format is the right way to go, I believe - it's the foundation of implementing C++ std::bfloat16 type, and referring to AArch64 and x86 experience, we should go to define that in that way. Otherwise, we might change that later.


Refs:
[1] https://discourse.llvm.org/t/patch-add-optional-bfloat16-support/63149
[2] https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p1467r9.html
[3] https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p1467r9.html#extended-wording

nick-knight:
> Promotion from BF16 to F16 is cheap - just a shift left by 16.

I think you mean FP32, but the comment still needs clarification regarding behavior with NaNs.

There are competing definitions of BF16, and BF16 isn't (yet) explicitly defined in IEEE-754. But IEEE-754 does provide a framework for defining BF16 as a new binary arithmetic and interchange type. And in that framework, a BF16 signalling NaN must raise NV on conversion to FP32 and must return a quiet NaN. Additionally, following RISC-V (implementation-defined) behavior for other floating-point conversions, if the input to a BF16 to FP32 conversion is any kind of NaN, the output should be `0x7fc00000`, the canonical (quiet) NaN.

joshua-arch1:
It seems that the std::bfloat16_t part in https://github.com/itanium-cxx-abi/cxx-abi/pull/147 has been resolved. Can we now continue addressing this PR?

joshua-arch1:
It seems that the std::bfloat16_t part in https://github.com/itanium-cxx-abi/cxx-abi/pull/147 has been merged. Can we now continue addressing this PR and finalize it? @kito-cheng @asb

kito-cheng (approved):
As Nick pointed out, I do have some concerns regarding the bfloat16 arithmetic. However, this PR primarily describes the alignment and size, which I believe can be separated from the computational details.

So  LGTM.

kito-cheng:
Oh, wait, @asb could you add one more sentence in C++ Name Mangling to mention `__bf16` is following the mangling rule of `std::bfloat16_t`?

asb:
Yes I saw that linked PR finally got resolved - thanks for the nudge. I'll re-review and update this patch tomorrow.

kito-cheng:
Thanks!

PR #368, opened 2023-03-14
Use "Procedure Linkage Table" for PLT

lbmeng:
The conventional name for PLT is Procedure Linkage Table.


jrtc27 (approved):
Not sure how that crept in...

kito-cheng (approved):
Some place also use `Program Linkage Table` as PLT, but `Procedure Linkage Table` is more widely use, so LGTM.

Ref:
# of Google result "Program Linkage Table":  ~600
# of Google result "Procedure Linkage Table":  ~17800

Issue #305, opened 2022-07-27
[meta] ABI deviations


kito-cheng:
It's meta issue for tracking all ABI deviations.

Anders@IAR has released the first version of the proposal of EABI deviations, the full document is attached, we will use this  proposal as base to start the discussion.

## Background

A question about an embedded ABI was raised in the RISC-V community.
The main reason why a separate embedded ABI was requested, was the fear that a rather high number of temporary (caller save) registers would have an intolerable impact on interrupt functions. 

An EABI task group was formed to sort it out, but it turned out that the problem was more complex than anticipated. Later on the work in the EABI TG was transferred to the psABI TG, and the EWRISCV team at IAR Systems AB was appointed to investigate how an ABI for embedded applications should be designed.

## The IAR investigation

We have concluded that an embedded ABI in a traditional sense should not be needed. A separate EABI (with fewer temporary registers) would result in both slower and larger regular code, and the problem with many temporary registers for interrupt functions can be handled. 

By offering a set of deviations from a main ABI, all of the needs for embedded applications can be fulfilled.

The main objectives in our work have been:
- The deviations must be defined by what can be done in the future. We cannot let current limitations in today’s tools being the
guideline in our work. The RISC-V architecture will probably be used for at least 50+ years to come, so the future must be our guideline.
- The deviations must be defined in a way so we tool vendors can implement and improve support for each deviation in a reasonable pace. It must be ok to not support one or more of the proposed deviations.
- Keeping the number of main/top ABIs as low as possible in order to minimize the complexity of the tool ecosystem. This means that the calling conventions for regular non-embedded applications and embedded applications will be the same.
- Handel the deviations as well-defined sidesteps from an ABI. The deviations should be seen as a number of features one can use for fine tuning primarily embedded applications.
- Deviations from an ABI should always be safe and result in smaller and/or faster code when applicable. No sloppy solutions
or dirty fixes are allowed!
- The proposed solutions should also be useful for non-embedded applications when applicable.

## What is the process?

- Identify and collect possible deviations.
  - First version is proposed by IAR.
  - Create github issue for each deviation to tracking and collect feedbacks
- Discuss the deviations and then follow up in meetings, we may decide to change, accept or reject the proposed deviations individually
- Define the syntax, semantic and encoding.

## What is ABI deviation?

Each ABI deviation means an option of ABI design, so that users could pick up the best deviation combination for their own embedded application.

## Deviation

- Reservation of registers
[...]

Issue #298, opened 2022-07-09
Unallocatable global pointer (GP)


MaskRay:
GNU ld can relax PC-relative addressing to GP-relative addressing in `-no-pie` mode (disable for `-pie`/`-shared`).
```
# a.s
.globl _start
_start:
  lui a0,%hi(var)
  addi a0,a0,%lo(var)

  lla a0,var
  ret

.data
.space 8
.globl var
var:
.zero 4
.size var, . - var
```
=> (`riscv64-linux-gnu-gcc -nostdlib -no-pie a.s`)
```
addi a0,gp,st_value(var)-__global_pointer$
addi a0,gp,st_value(var)-__global_pointer$
```
There are some bugs, e.g. if I remove `.space 8`, it is not relaxable. If I use `.space 4` and keep `.size`, it is not relaxable.

IMO this is the de facto reason that GP is unallocable.
Questions:

1. Is GP relaxation useful? How much does it save?
2. Shall we make GP allocatable? If so, callee-saved, temporary, or reserved for a custom ABI?
3. If GP is made allocatable, how to handle the compatibility issue with GNU ld which performs GP relaxation?

For 1, I am strongly opinionated that it is not, and we should remove GP relaxation from GNU ld.
Main arguments:

(a) Global data access instructions typically take a very small portion of the code size and should not be a performance bottleneck. Having one more register at proposal (especially in RVC) likely provides more benefits.
(b) The relaxable range `[__global_pointer$-0x800, __global_pointer$+0x800)` is too small. On the other side, if some embedded usage really thinks GP relaxation is useful, they can perform GP relaxation in their custom ABI.

For 3, if a new relocatable object file uses GP for a different purpose, it can use a new relocation type to indicate the use which is incompatible with a linker performing GP relaxation. An old linker will report an error for the unrecognized relocation type.

Note: The resolution to #205 removed R_RISCV_GPREL_I/R_RISCV_GPREL_S used internally by GNU ld.

jrtc27:
GP relaxation is too entrenched in the GNU world to break; that'd be one major ABI break. Whether we think it's a good idea to have done or not it's what we have and have to deal with.

MaskRay:
> GP relaxation is too entrenched in the GNU world to break; that'd be one major ABI break. Whether we think it's a good idea to have done or not it's what we have and have to deal with.

3 is indeed a problem. We can remove GP relaxation from GNU ld now to make the compatibility problem less severe.

[...]

Issue #370, opened 2023-03-20
Specify a platform reserved register


appujee:
ARM reserves x18 for platform but RISC-v doesn't specify a platform register.

https://developer.arm.com/documentation/den0024/a/The-ABI-for-ARM-64-bit-Architecture/Register-use-in-the-AArch64-Procedure-Call-Standard/Parameters-in-general-purpose-registers

Recent changes in llvm has reserved the same `x18` for Android and Fuschia platforms to align with the **name** w.r.t. ARM spec. But x18 may not be an ideal choice for RISC-V because:
- it is the 3rd callee saved register
- `-msave-restore` does not interact well if `x18` gets reserved.

Some discussion has been conitnuing here: https://github.com/google/android-riscv64/issues/72 but it would be ideal to have a spec or at least a recommendation on choice of platform registers and trade-offs around that

In the current setup `x27` appears to be a better choice as this would reduce:
- interaction with `-msave-restore`
- amount of work in assembler code out there. `X18` is used in more places (3rd callee saved register) than X27.

kito-cheng:
It would be little bit later to reserve a reserved register for RISC-V since the ABI already used for a while, so adding one general reserved register would be annoying to the current ecosystem especially for the linux distro guys...

However here is an **_evil_** thought in my mind, GP relaxation might not very useful on non-baremetal environment so we might use that as special use for fuchsia and Android other than GP relaxation, that come with one more advantage is: all existing code case didn't use `gp` register in program including Assembly codes.

nick-knight:
As I understand it, this issue is not a request to change the Linux ABI, but to solicit recommendations on minimally disruptive specializations of it. It is understood that this would result in a new, incompatible ABI. Aside from the technical question, there is also the procedural question of how/where to standardize this. I think my suggestion to @appujee to jump-start the discussion here was misguided, and it would be better to start at [sig-toolchains](https://lists.riscv.org/g/sig-toolchains).

FWIW, I think the Android folks would be perfectly happy with a nonnormative note in the psABI --- e.g., "variant ABIs that add a platform register are encouraged to throw away GP-relaxation and use `x3` for this purpose" --- so they could just point to this to push their LLVM patches through. Of course, it'll eventually be necessary to have an actual spec.

appujee:
[...]

PR #371, opened 2023-03-22
Relax gp could be platform specific register rather than reserved for…

kito-cheng:
… linker relaxation

The usage of gp register has discussed serveral times (e.g. #298 and #305), and it reserved as special register used for linker relaxation, that could be improve perfomance and code size.

However it come with some limitation, like it can't applicable on shared libraries, and it also not work well when program come with large datas since the relaxable range is only +-2KiB.

Some platform like Haiku never use gp in the whole system, and also this might not useful in some baremetal system with specialized memory layout, so we might consider to release the gp usage with a non-hard-ABI-breakage way.

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -28,6 +28,16 @@ endif::[]
 In the standard ABI, procedures should not modify the integer registers tp and
 gp, because signal handlers may rely upon their values.
 
+The presence of a frame pointer is optional.  If a frame pointer exists,
+it must reside in x8 (s0); the register remains callee-saved.
+
+If a platform requires use of a dedicated general-purpose register for a
+platform-specific purpose, it is recommended to use gp (x3). The platform ABI
+specification must document the use of this register. For such platforms, care
+must be taken to ensure all code (compiler generated or otherwise) avoids using
+gp in a way incompatible with the platform specific purpose, and that global
+pointer relaxation is disabled in the toolchain.
+
 === Frame Pointer Convention
 
 The presence of a frame pointer is optional.  If a frame pointer exists,


appujee:
SGTM. Thanks for proposing this.

For reference: More discussion around this has been going on in: https://lists.riscv.org/g/sig-toolchains/topic/97736679#548 and https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/370

rui314:
Different platforms might want to use the "platform register" for different purposes, no? If so, just defining `1` for the platform register cannot prevent incompatible object files to be linked to a single executable. We might want to define a value for each specific usage.

kito-cheng:
@rui314 good point, I add 2 range for standard and non-stanadrd use used range, and also one for undefined one.

rui314:
I'm not sure what the undefined purpose is. Maybe we just want to reserve 2-1024 for the standard uses and 1025-2047 for nonstandard uses?

kito-cheng:
My first thought about that undefined purpose is kind of reserved for nonstandard use, but we already defined one range for nonstandard, so yeah, that should be just 2-1024 for the standard uses and 1025-2047 for nonstandard uses

riscv-elf.adoc
```diff
@@ -1164,6 +1165,25 @@ Merge policy:::
 The linker should report errors if object files of different privileged
 specification versions are merged.
 
+===== Tag_RISCV_gp_reg_usage, 14, uleb128=value
+
+Tag_RISCV_gp_reg_usage indicate the usage of gp register, gp could be used for
+global pointer relaxation, reserved platform register or a temporary register.
+
+[horizontal]
+0:: This object use gp as global pointer relaxation.
```
MaskRay, line 1174: What value is defined for unused gp?

appujee: Even if gp is unused, is it safe to assume that gp is used for global pointer relaxation? As this has been the default behavior.

Can there be cases where gp is allowed for global pointer relaxation and no such opportunities were found in the binary?

kito-cheng: > Can there be cases where gp is allowed for global pointer relaxation and no such opportunities were found in the binary?

In theory we can made one program with a symbol with huge size (large then 4k) and point `__global_pointer$` to the middle of the symbol, then the global pointer relaxation will have no way to relax anything, but that case should not happened in generally since `__global_pointer$` is default computed by a magic (or heuristic) expression with some arithmetic with sbss/sdata.

kito-cheng:
Changes:
- Add `unused` case
- Drop `undefined` case
- Made 3~1023 as reserved for future standard defined platform register.

NOTE: GP as SCS will defined in separated PR.

kito-cheng:
> Honestly I don't feel good to add Tag_RISCV_gp_reg_usage now. There are many C++ dialect affecting -f options, yet we don't have compatibility attributes for them. AArch64 has SHT_AARCH64_ATTRIBUTES, but it isn't really used.

IMO that could be used to sanity check and identity some information how this object/executable compiled, otherwise the only way is writing some magic script/tools to check that like https://github.com/google/android-riscv64/issues/77 mentioned, a binary validator...

`-frecord-command-line ` might be another way, but some info. might not able to know from command line since some setting might be implied by default configuration or target triple, so that's not a reliable way.

There is many C++ dialect, one recent happened example is [relative vtables ABI for C++](https://youtu.be/9HGKlDiJy8E), I didn't ask to add a tag to identify that is because it's still `experimental` (`-fexperimental-relative-c++-abi`) (I didn't express this at that PR I know, but that's what I thought and I should say that out at that time...).

Anyway, my point is we should track/record those information if possible, adding attributes could help tool to having better diagnosis message instead of silently broken, otherwise people don't have too much clue to tracking what happened when screw up things.

MaskRay:
If a source file uses inline assembly to initialize gp, we can't know how the user intends to use gp. For GP relaxation, the use case is done in the final link phase. The compile phases don't know and should not need to use an option to tell the compiler *not* to create an attribute/note indicating GP relaxation.

I don't deny that a compatibility detecting feature may be useful, but we don't need to do it now. (For this case I suspect I slightly lean toward never having it.) If we want to do it, it can be a separate proposal. I have seen a lot of inconvenience in binutils/glibc adding some compatibility detecting feature and I strongly believe waiting a bit will have a large value.

kito-cheng:
@MaskRay Thanks for sharing your thought , split `Tag_RISCV_gp_reg_usage` out, and that also made this PR become no PoC required, and that's kind of big change to RISC-V ABI, so I would like to get you explicitly LGTM :)

kito-cheng:
@asb @luismarques although `gp` only used in linker relaxation stage, so per our policy that only require consensus from linker gurus, but it's a big change so I would like to make sure RISC-V LLVM maintainer is OK with this.

@rui314 you are linker expert, and I would like all linker expert is happy with this change.

@Nelson1225 I need your opinion as binutils expert :) 

@jrtc27 FreeBSD didn't use gp relaxation, so I guess you are fine with this change, but I would like you give explicitly LGTM to this one if possible :)

asb:
> @asb @luismarques although `gp` only used in linker relaxation stage, so per our policy that only require consensus from linker gurus, but it's a big change so I would like to make sure RISC-V LLVM maintainer is OK with this.

I think although platforms can of course deviate from the standard ABI however they wish, having a suggested register for platform specific purposes makes sense. GP is convenient for this purpose and there seem to be no strong concerns about it, so all LGTM.

I think you need to update the commit message so it no longer claims to introduce a new tag (as this was dropped for now I believe).

riscv-cc.adoc
```diff
@@ -31,6 +31,9 @@ gp, because signal handlers may rely upon their values.
 The presence of a frame pointer is optional.  If a frame pointer exists,
 it must reside in x8 (s0); the register remains callee-saved.
 
+Platform might use gp register as platform specific register other than used for
```
asb, line 34: How about "If a platform requires use of a dedicated general-purpose register for a platform-specific purpose, it is recommended to use gp (x3). The platform ABI specification must document the use of this register and care must be taken to disable global pointer relaxation in toolchains when targeting that platform."

asb:
I've left a suggested wording change for the actual patch part of this (inspired partially by #374). I don't think we should equate gp just with gp relaxation, as it might e.g. be used in hand-written assembly.

riscv-cc.adoc
appujee: Should we also specify that, when **gp** is reserved for a platform-specific purpose:
- Toolchain may not allocate **gp** for any other purpose than for platform-specific purpose.
- Care must be taken while compiling assembly code as it may be using **gp**  in undesired ways.

IMO It would help to add a note that as per the current spec, the compiler does not add a tag in ELF that would indicate the way **gp** is being used. given a pre-built binary, **gp** might be getting used either for platform-specific purpose or for global pointer relaxation. The onus of ensuring **gp** is being used as intended lies entirely on the consumer of a pre-built binary.

aswaterman (changes requested):
I disagree with the premise of this PR.  A platform might choose to designate _any_ register as used for an alternate purpose.  No matter which register is chosen, it affects the ABI somehow.

jrtc27:
> No matter which register is chosen, it affects the ABI somehow.

Which is why we're constraining it so it's *not* a free-for-all. Otherwise you end up in a situation where something like OpenSSL needs a different copy of its crypto assembly routines for each platform because it needs to avoid each platform's reserved register. The whole point of a specification like a psABI is to impose order on the chaos so everyone agrees on what is going on; if you want to allow any register to be reserved by any platform and remain standards-compliant then why bother having an ABI at all?

aswaterman:
@jrtc27 indeed I perceive the intent to reserve `gp` to be part of a free-for-all promulgated by people who hate linker relaxation as a concept.  This is not a process; it's a mob.

jrtc27:
> @jrtc27 indeed I perceive the intent to reserve `gp` to be that of a free-for-all promulgated by people who hate linker relaxation as a concept. This is not a process; it's a mob.

This is not constructive, nor is it factually accurate. Please take this rhetoric elsewhere; it has no place in the RISC-V community.

aswaterman (approved):

aswaterman:
> Before further discussing which register to specify, I wanted to check if you disagree that's there's any value at all in giving such a hint?

Of course, I agree there's value in doing so.

> is it that choosing gp for this platform specific register might be seized upon to force a decision in a particular direction in any future debate about gp relaxations?

That was precisely my point, although I failed to articulate it politely.

asb:
> > is it that choosing gp for this platform specific register might be seized upon to force a decision in a particular direction in any future debate about gp relaxations?
> 
> That was precisely my point, although I failed to articulate it politely.

It's probably a good thing to discuss this concern explicitly though I think.

So for my part, my LGTM on this specific change is not indicative of any point of view on the best use of GP in the standard ABI as already implemented. I suspect there will be recurrent discussions about it because if you want to scavenge a register for another purpose, GP is going to be near the top of the list. But I see this PR as separate to any such discussion.

appujee:
> @jrtc27 indeed I perceive the intent to reserve gp to be that of a free-for-all promulgated by people who hate linker relaxation as a concept. This is not a process; it's a mob.

AIUI, the range of gp is small (+/- 2kb) and probably that's why we haven't seen compiler optimizations leveraging that.  To the best of my knowledge, none of the Android libraries use gp. If you have ideas on how gp can help in platforms like Android, it will be good to share such that we can realize the trade-offs being made here.

jrtc27:
> > @jrtc27 indeed I perceive the intent to reserve gp to be that of a free-for-all promulgated by people who hate linker relaxation as a concept. This is not a process; it's a mob.
> 
> AIUI, the range of gp is small (+/- 2kb) and probably that's why we haven't seen compiler optimizations leveraging that. To the best of my knowledge, none of the Android libraries use gp. If you have ideas on how gp can help in platforms like Android, it will be good to share such that we can realize the trade-offs being made here.

GP relaxation is only usable by executables, not libraries, since it takes a constant value throughout execution and thus libraries would compete for its use (though I guess there's an interesting question of whether one could designate a specific library as being the One True GP User, like libc). If I remember correctly, someone from Google in the call on Monday (Elliott?) said that ~everything is a library on Android, which would explain why GP relaxation is not of much use to them? (Though perhaps something like ART could then be blessed as the One True GP User...)

ilovepi:
@aswaterman I'm not sure where you stand on this change, given that you've approved it now, but your comments seem to conflict with that sentiment. Can you provide some clarification on your stance and any lingering concerns?

aswaterman:
Although I remain concerned that this choice might be speciously used as a pretext to stop using gp relaxations in existing ABIs, I agree with @asb that this PR can go through without addressing that matter.

appujee (approved):
LGTM the intent of the patch. As I mentioned in the review, it will be nice to have some explanation on the a) use case b) requirements on the toolchain, and c) risks of using gp for platform-specific purpose without a tag.

riscv-cc.adoc
asb: I don't feel particularly strongly, but I'm not sure it's worth going into details about the lack of an ELF tag. I don't think ELF tags are a good match for something at this level (after all, they're no help at all for inline assembly). I'm normally in favour of being a bit verbose to be more easily understandable, but I'm not sure further spelling out how gp can't be used for anything other than the platform-specific purpose adds a lot. But happy to go along with it if others feel differently.

Perhaps:
"If a platform requires use of a dedicated general-purpose register for a platform-specific purpose, it is recommended to use gp (x3). The platform ABI specification must document the use of this register. For such platforms, care must be taken to ensure all code (compiler generated or otherwise) avoids using gp in a way incompatible with the platform specific purpose, and that global pointer relaxation is disabled in the toolchain."

kito-cheng:
@aswaterman I guess I might give you impression that I am not support GP relaxation in #298, but I believe you know that I did know the value of GP relaxation in specific areas, especially on the small bare-metal system, and it also gives us some extra spec score, so I am not intending to deprecate the GP relaxation. 

For the SCS, the ideal case is using Zisslpcfi extension, but that would be an issue since it’s not ready now, and also it still require several psABI works (not only SCS, but also landing pad stuffs), so we need to figure out a way to make it moving forward, and GP is the relative best candidate so far.

And of course that’s kind of a mistake that we didn’t learn from other platforms - we didn’t reserve a platform register at the beginning, picking any other general purpose register as platform register would be more painful at this moment.

riscv-cc.adoc
kito-cheng: @asb thanks a lot for improving the text! hope you don't mind I listed you as co-author in this patch :)

kito-cheng:
Changes:
- Rebase
- Apply @asb's suggesion

kito-cheng:
Thanks everyone, It seems we have converged on this topic, I would like to wait few more day to make sure no further comment before merge :)

riscv-elf.adoc
X547: > Even if gp is unused, is it safe to assume that gp is used for global pointer relaxation?

It seems not. Operating system may not initialize GP with `__global_pointer` and leave it with zero or some undefined value. Haiku for example do not define or use `__global_pointer` because it do not distinguish executables and shared libraries.

X547:
> GP relaxation is only usable by executables, not libraries, since it takes a constant value throughout execution and thus libraries would compete for its use (though I guess there's an interesting question of whether one could designate a specific library as being the One True GP User, like libc).

PowerPC ABI have similar concept to GP register (TOC register R2) that also works in shared libraries by using special library calling convention with switching TOC register to address of called shared library and switch back on return.

> If I remember correctly, someone from Google in the call on Monday (Elliott?) said that ~everything is a library on Android, which would explain why GP relaxation is not of much use to them?

The same for Haiku. Executables and shared libraries are the same except NULL entry point is allowed for shared libraries, but not for executables.

MaskRay (approved):
Thanks for the update. The new wording looks good to me.

kito-cheng:
> PowerPC ABI have similar concept to GP register (TOC register R2) that also works in shared libraries by using special library calling convention with switching TOC register to address of called shared library and switch back on return.

@X547  thanks for provide this info, that's interesting idea we could consider in embedded ABI I think.

kito-cheng:
Thanks everyone again :)

MaskRay:
> > PowerPC ABI have similar concept to GP register (TOC register R2) that also works in shared libraries by using special library calling convention with switching TOC register to address of called shared library and switch back on return.
> 
> @X547 thanks for provide this info, that's interesting idea we could consider in embedded ABI I think.

@X547 I don't think PowerPC64 ELFv2 TOC is a good design:) (I have many lld patches in this area) See my summary of TOC on https://maskray.me/blog/2023-02-26-linker-notes-on-power-isa

X547:
> the compiler inserts a NOP after the branch instruction.

Can be relaxed, no?

MaskRay:
> > the compiler inserts a NOP after the branch instruction.
> 
> Can be relaxed, no?

It can, but I am unsure about the linker complexity.

Their mere existence makes a tail call no longer a tail cal. I have summarized all sorts of maintenance costs on https://maskray.me/blog/2023-02-26-linker-notes-on-power-isa

kito-cheng:
https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/387 for adding new tag for recording gp usage

578881718:
Hello, I would like to ask why not add an add instruction to expand the range of gp to 32bits, but it is only +/-2k.

Issue #94, opened 2019-05-26
Support Thread-Local Storage Descriptors (TLSDESC)


MaskRay:
TLSDESC (`-mtls-dialect=gnu2`) improves traditional General Dynamic and Local Dynamic TLS models (`-mtls-dialect=gnu`). In the most common case that TLS variables are defined in initially-loaded modules, it simplifies the work in `__tls_get_addr` and (probably more importantly) switches to a custom calling convention that doesn't clobber any register ("preserve any registers they modify"). This speedup is significant.

The linker may relax TLSDESC code sequence to Initial Exec (targeting an executable, the symbol is preemptable) or Local Exec (target an executable, the symbol is non-preemptable) if applicable.

The initial (static) and outstanding (dynamic) relocation types for TLSDESC have to be defined, as well as how static relocation types are relaxed to Initial Exec and Local Exec models.

TLSDESC is currently available on x86, x86-64, arm and aarch64.

x86: https://www.fsfla.org/~lxoliva/writeups/TLS/RFC-TLSDESC-x86.txt
ARM: https://www.fsfla.org/~lxoliva/writeups/TLS/RFC-TLSDESC-ARM.txt (the published paper was referenced by ELF for the Arm 64-bit Architecture (AArch64) and ARM 32-bit)

As I understand it, TLSDESC is a strict improvement, so it might be worth defaulting to TLSDESC and probably deprecating relocation types for General Dynamic/Local Dynamic.

aswaterman:
Seems like a good idea, though a big chunk of work to specify and implement.  Any chance you're volunteering? :)

MaskRay:
I'm still getting my feet wet with riscv toolchain (tracking the progress of musl-riscv, reading `llvm/lib/Target/RISCV` `binutils-gdb/bfd/elf*-riscv.c`, etc)... but I would be happy if I can help. I have some contribution to lld so I may help on the lld (llvm linker) side.

richfelker:
One note based on comments from sorear on `#musl` where this was discussed earlier: since TLSDESC calls have a custom calling convention (usually with few or no clobbers) anyway, it would probably make sense to use a non-default link register for the call. This would avoid the need for the caller to save and restore it.

aswaterman:
Indeed, `t0` is intended for this purpose.

rui314:
[...]

PR #373, opened 2023-03-27
Add specification for TLS descriptors

ishitatsuyuki:
This is a draft to propose a TLS descriptor scheme as requested in #94.

Currently working on a prototype on GNU toolchain to iron out potential issues, and until that's done this will remain a draft.

[Rendered](https://github.com/ishitatsuyuki/riscv-elf-psabi-doc/blob/tlsdesc/riscv-elf.adoc)

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -805,7 +805,7 @@ thread local storage models:
 
 [[tls-model]]
 .TLS models
-[cols="1,2,3"]
+[cols="1,2"]
 [width=70%]
 |===
 | Mnemonic | Model
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -368,6 +368,8 @@ Description:: Additional information about the relocation
                                             <| S + A + TLSOFFSET
 .2+| 11      .2+| TLS_TPREL64   .2+| Dynamic | _word64_          .2+|
                                             <| S + A + TLSOFFSET
+.2+| 12      .2+| TLSDESC       .2+| Dynamic | See <<TLS Descriptors>> .2+|
+                                            <| TLSDESC(S+A)
 .2+| 16      .2+| BRANCH        .2+| Static  | _B-Type_          .2+| 12-bit PC-relative branch offset
                                             <| S + A - P
 .2+| 17      .2+| JAL           .2+| Static  | _J-Type_          .2+| 20-bit PC-relative jump offset
@@ -452,7 +454,15 @@ Description:: Additional information about the relocation
                                             <| S + A
 .2+| 61      .2+| SUB_ULEB128   .2+| Static  | _ULEB128_         .2+| Local label subtraction <<uleb128-note,*note>>
                                             <| V - S - A
-.2+| 62-191  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
+.2+| 62      .2+| TLSDESC_HI20      .2+| Static  | _U-Type_          .2+| High 20 bits of a 32-bit PC-relative offset into a TLS descriptor entry, `%tlsdesc_hi(symbol)`
+                                            <| S + A - P
+.2+| 63      .2+| TLSDESC_LOAD_LO12 .2+| Static  | _I-Type_          .2+| Low 12 bits of a 32-bit PC-relative offset into a TLS descriptor entry, `%tlsdesc_load_lo(address of %tlsdesc_hi)`, the addend must be 0
+                                            <| S - P
+.2+| 64      .2+| TLSDESC_ADD_LO12  .2+| Static  | _I-Type_          .2+| Low 12 bits of a 32-bit PC-relative offset into a TLS descriptor entry, `%tlsdesc_add_lo(address of %tlsdesc_hi)`, the addend must be 0
+                                            <| S - P
+.2+| 65      .2+| TLSDESC_CALL      .2+| Static  |                   .2+| Annotate call to TLS descriptor resolver function, `%tlsdesc_call(address of %tlsdesc_hi)`, for relaxation purposes only
+                                            <|
+.2+| 66-191  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
                                             <|
 .2+| 192-255 .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for nonstandard ABI extensions
                                             <|
@@ -940,6 +950,55 @@ typedef struct
   unsigned long int ti_offset;
 } tls_index;
 ----
+
+==== TLS Descriptors
+
+TLS Descriptors (TLSDESC) are an alternative implementation of the Global Dynamic model
+that allows the dynamic linker to achieve performance close to that
+of Initial Exec when the library was not loaded dynamically with `dlopen`.
+
+The linker reserves a consecutive pair of pointer-sized entry in the GOT for each `TLSDESC`
+relocation. At runtime, the dynamic linker fills in the TLS descriptor entry as defined below:
+
+[,c]
+----
+typedef struct
+{
+  unsigned long (*entry)(tls_descriptor *);
+  unsigned long arg;
+} tls_descriptor;
+----
+
+Upon accessing the thread local variable, the `entry` function is called with the address
+of `tls_descriptor` containing it, returning `<address of thread local variable> - tp`.
+
+The TLS descriptor `entry` is called with a special calling convention, specified as follows:
+
+- `a0` is used to pass the argument and return value.
+- `t0` is used as the link register.
+- Any other registers are callee-saved. This includes any vector registers when the vector extension is supported.
+
+Example assembler load and store of a thread local variable `i` using the `%tlsdesc_hi`, `%tlsdesc_load_lo`, `%tlsdesc_add_lo` and `%tlsdesc_call`
+assembler functions. The emitted relocations are in the comments.
+
+[,asm]
+----
+label:
+	auipc tX, %tlsdesc_hi(symbol)         // R_RISCV_TLSDESC_HI20 (symbol)
+	lw    tY, tX, %tlsdesc_load_lo(label) // R_RISCV_TLSDESC_LOAD_LO12_I (label)
+	addi  a0, tX, %tlsdesc_add_lo(label)  // R_RISCV_TLSDESC_ADD_LO12_I (label)
+	jalr  t0, tY, %tlsdesc_call(label)    // R_RISCV_TLSDESC_CALL (label)
+----
+
+`tX` and `tY` in the example may be replaced with any combination of two general purpose registers.
+
+The `%tlsdesc_call` assembler function does not return a value and is used purely
+to associate the `R_RISCV_TLSDESC_CALL` relocation with the `jalr` instruction.
+
+The linker can use the relocations to recognize the sequence and to perform relaxations. To ensure correctness, only the following changes to the sequence are allowed:
+
+- Instructions outside the sequence that do not clobber the registers used within the sequence may be inserted in-between the instructions of the sequence (known as instruction scheduling).
+- Instructions in the sequence with no data dependency may be reordered. In the preceding example, the only instructions that can be reordered are `lw` and `addi`.
 
 === Sections
 
@@ -1610,6 +1669,91 @@ Relaxation result:
 [,asm]
 ----
     lw t1, <tp-offset-for-symbol>(tp)
+----
+--
+
+==== TLS Descriptors -> Initial Exec Relaxation
+
+Target Relocation:: R_RISCV_TLSDESC_HI20, R_RISCV_TLSDESC_LOAD_LO12_I, R_RISCV_TLSDESC_ADD_LO12_I, R_RISCV_TLSDESC_CALL
+
+Description:: This relaxation can relax a sequence loading the address of a thread-local symbol reference into a GOT load instruction.
+
+Condition::
+- Linker output is an executable.
+
+Relaxation::
+
+- Instruction associated with `R_RISCV_TLSDESC_HI20` or `R_RISCV_TLSDESC_LOAD_LO12_I` can be removed.
+- Instruction associated with `R_RISCV_TLSDESC_ADD_LO12_I` can be replaced with load of the high PC-relative offset of the symbol's GOT entry.
+- Instruction associated with `R_RISCV_TLSDESC_CALL` can be replaced with load of the low PC-relative offset of the symbol's GOT entry.
+Example::
++
+--
+Relaxation candidate (`tX` and `tY` can be any combination of two general purpose registers):
+
+[,asm]
+----
+label:
+	auipc tX, <hi>      // R_RISCV_TLSDESC_HI20 (symbol), R_RISCV_RELAX
+	lw    tY, tX, <lo>  // R_RISCV_TLSDESC_LOAD_LO12_I (label), R_RISCV_RELAX
+	addi  a0, tX, <lo>  // R_RISCV_TLSDESC_ADD_LO12_I (label), R_RISCV_RELAX
+	jalr  t0, tY        // R_RISCV_TLSDESC_CALL (label), R_RISCV_RELAX
+----
+
+Relaxation result:
+
+[,asm]
+----
+	lui a0, <pcrel-got-offset-for-symbol-hi>
+	{ld,lw} a0, <pcrel-got-offset-for-symbol-lo>(a0)
+----
+--
+
+==== TLS Descriptors -> Local Exec Relaxation
+
+Target Relocation:: R_RISCV_TLSDESC_HI20, R_RISCV_TLSDESC_LOAD_LO12_I, R_RISCV_TLSDESC_ADD_LO12_I, R_RISCV_TLSDESC_CALL
+
+Description:: This relaxation can relax a sequence loading the address of a thread-local symbol reference into a thread-pointer-relative instruction sequence.
+
+Condition::
+
+- Short form only: Offset between thread-pointer and thread-local symbol is within +-2KiB.
+- Linker output is an executable.
+- Target symbol is non-preemptible.
+
+Relaxation::
+
+- Instruction associated with `R_RISCV_TLSDESC_HI20` or `R_RISCV_TLSDESC_LOAD_LO12_I` can be removed.
+- Instruction associated with `R_RISCV_TLSDESC_ADD_LO12_I` can be replaced with the high TP-relative offset of symbol (long form) or be removed (short form).
+- Instruction associated with `R_RISCV_TLSDESC_CALL` can be replaced with the low TP-relative offset of symbol.
+
+Example::
++
+--
+Relaxation candidate (`tX` and `tY` can be any combination of two general purpose registers):
+
+[,asm]
+----
+label:
+	auipc tX, <hi>      // R_RISCV_TLSDESC_HI20 (symbol), R_RISCV_RELAX
+	lw    tY, tX, <lo>  // R_RISCV_TLSDESC_LOAD_LO12_I (label), R_RISCV_RELAX
+	addi  a0, tX, <lo>  // R_RISCV_TLSDESC_ADD_LO12_I (label), R_RISCV_RELAX
+	jalr  t0, tY        // R_RISCV_TLSDESC_CALL (label), R_RISCV_RELAX
+----
+
+Relaxation result (long form):
+
+[,asm]
+----
+	lui a0, <tp-offset-for-symbol-hi>
+	addi a0, a0, <tp-offset-for-symbol-lo>
+----
+
+Relaxation result (short form):
+
+[,asm]
+----
+	addi a0, zero, <tp-offset-for-symbol>
 ----
 --
 


ishitatsuyuki:
Main changes since https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/94#issuecomment-1473756420:
- Scheduling / interleaving is allowed.
- The compiler may choose the temporary registers used in the instruction sequence.
- The descriptor structure is now 4-word long, which is enough space to accommodate `resolver, module_id, module_offset, module_generation` as used in glibc. This makes the GOT entries larger but avoids the need to separately allocate descriptor memory for `dlopen`ed modules. (Thanks Rui for the idea.)

enh-google:
@rprichard --- any thoughts from the bionic implementation perspective?

riscv-elf.adoc
```diff
@@ -1521,6 +1580,91 @@ Relaxation result:
...
+----
+--
+
+==== TLS Descriptors -> Local Exec Relaxation
+
+Target Relocation:: R_RISCV_TLSDESC_HI20, R_RISCV_TLSDESC_LOAD_LO12_I, R_RISCV_TLSDESC_ADD_LO12_I, R_RISCV_TLSDESC_CALL
+
+Description:: This relaxation can relax a sequence of the load address of a symbol or load/store with a thread-local symbol reference into a thread-pointer-relative instruction.
+
+  **TODO:** How do we handle load/stores? Should we annotate the load/store instruction with an additional relocation so that the low offset can be fused into the load/store?
```
rui314, line 1625: I don't think we need to try to handle load/stores for TLSDESC because we don't handle even usual GOT-indirect load/stores. I think we can just forget about it.

riscv-elf.adoc
```diff
@@ -1521,6 +1580,91 @@ Relaxation result:
...
+--
+Relaxation candidate:
+
+[,asm]
+----
+label:
+	auipc a0, 0     // R_RISCV_TLSDESC_HI20 (symbol), R_RISCV_RELAX
+	lw    t0, a0, 0 // R_RISCV_TLSDESC_LOAD_LO12_I (label), R_RISCV_RELAX
+	addi  a0, a0, 0 // R_RISCV_TLSDESC_ADD_LO12_I (label), R_RISCV_RELAX
+	jalr  t0, t0    // R_RISCV_TLSDESC_CALL (label), R_RISCV_RELAX
```
rui314, line 1607: Don't you want to generalize registers with tX and tY?

riscv-elf.adoc
```diff
@@ -929,6 +941,53 @@ typedef struct
...
+	lw    tY, tX, %tlsdesc_lo_load(label) // R_RISCV_TLSDESC_LOAD_LO12_I (label)
+	addi  a0, tX, %tlsdesc_lo_add(label)  // R_RISCV_TLSDESC_ADD_LO12_I (label)
+	jalr  t0, tY                          // R_RISCV_TLSDESC_CALL (label)
+----
+
+`tX` and `tY` in the example may be replaced with any combination of two general purpose registers.
+
+The assembly sequence may be arbitrarily interleaved with other instructions for the purpose of
+instruction scheduling, but the order of the 4 instructions must not change. The linker can use
+the relocations to recognize the sequence and to perform relaxations.
```
rui314, line 989: I don't think we want to add a constraint on the order of these four instructions unless it is technically required. The linker can perform optimizations even if these instructions are given in an arbitrary order, no?

riscv-elf.adoc
ishitatsuyuki: The pc-rel values in the examples are replaced with a concrete value, so I've also opted for concrete registers here. I don't have a strong opinion on this though.

riscv-elf.adoc
ishitatsuyuki: I've reworded the constraints and explicitly stated that `lw` and `addi` may be reordered with respect to each other.

riscv-elf.adoc
ishitatsuyuki: Removed the TODO, although I'll keep this thread open in case other reviewers want to share their opinion.

riscv-elf.adoc
```diff
@@ -929,6 +941,54 @@ typedef struct
...
+----
+typedef struct
+{
+  unsigned long (*resolver)(tls_descriptor *);
+  unsigned long data[3]; /* defined by the dynamic linker */
+} tls_descriptor;
+----
+
+The `resolver` function is called with the address to the `tls_descriptor` itself,
+and returns the address the TLS variable resolves to.
```
rui314, line 963: Doesn't it return the offset from TP?

riscv-elf.adoc
```diff
@@ -1521,6 +1581,89 @@ Relaxation result:
...
+----
+	lui a0, <tp-offset-for-symbol-hi>
+	addi a0, a0, <tp-offset-for-symbol-lo>
+----
+
+Relaxation result (short form):
+
+[,asm]
+----
+	addi a0, a0, <tp-offset-for-symbol>
```
rui314, line 1663: I think it should be `addi a0, x0, <tp-offset-for-symbol>`.

riscv-elf.adoc
ishitatsuyuki: The intention was to save a redundant subtraction and addition when DTV is used.

But now I realized that the relaxation only works if it's specified as the TP offset. So I've changed it to say offset from TP.

riscv-elf.adoc
ishitatsuyuki: Fixed.

riscv-elf.adoc
```diff
@@ -367,6 +367,10 @@ Description:: Additional information about the relocation
                                             <| S + A + TLSOFFSET
 .2+| 11      .2+| TLS_TPREL64   .2+| Dynamic | _word64_          .2+|
                                             <| S + A + TLSOFFSET
+.2+| 12      .2+| TLSDESC32     .2+| Dynamic | See <<TLS Descriptors>> .2+|
```
ilovepi, line 370: Is there a particular reason we need 32 + 64 suffix on these?

1. I think other platforms just use TLSDESC as the reloc type for both 32-bit and 64-bit
2. These map to the same value
3. AFAICT the only differences in this case are related to the size of a GOT slot, so any differences in how the relocation would be calculated should already be accounted for, right?

ishitatsuyuki: Thanks.

I didn't know that the other relocs with 32/64 suffix was considered a mistake, and you're right, it's only related to the size and that should be determinable from the binary target arch. I've removed the variants.

(By the way, mapping to the same value was unintentional and a copy-paste mistake.)

riscv-elf.adoc
```diff
@@ -929,6 +939,54 @@ typedef struct
 } tls_index;
 ----
 
+==== TLS Descriptors
+
+TLS Descriptors (TLSDESC) is an alternative implementation of the Global Dynamic model
```
ilovepi, line 944:
```suggestion
TLS Descriptors (TLSDESC) are an alternative implementation of the Global Dynamic model
```

riscv-elf.adoc
```diff
@@ -929,6 +939,54 @@ typedef struct
...
+that additionally allows the dynamic linker to achieve performance close to that
```
ilovepi, line 945:
```suggestion
that allows the dynamic linker to achieve performance close to that
```

riscv-elf.adoc
```diff
@@ -929,6 +939,54 @@ typedef struct
...
+label:
+	auipc tX, %tlsdesc_hi(symbol)         // R_RISCV_TLSDESC_HI20 (symbol)
+	lw    tY, tX, %tlsdesc_lo_load(label) // R_RISCV_TLSDESC_LOAD_LO12_I (label)
+	addi  a0, tX, %tlsdesc_lo_add(label)  // R_RISCV_TLSDESC_ADD_LO12_I (label)
+	jalr  t0, tY                          // R_RISCV_TLSDESC_CALL (label)
+----
+
+`tX` and `tY` in the example may be replaced with any combination of two general purpose registers.
+
+The linker can use the relocations to recognize the sequence and to perform relaxations. To ensure this correctly, only the following changes to the sequence are allowed:
```
ilovepi, line 985:
```suggestion
The linker can use the relocations to recognize the sequence and to perform relaxations. To ensure correctness, only the following changes to the sequence are allowed:
```

riscv-elf.adoc
```diff
@@ -929,6 +939,54 @@ typedef struct
...
+
+- Other instructions that does not clobber the registers used by the sequence may be inserted in between (known as instruction scheduling).
```
ilovepi, line 987:
```suggestion
- Instructions outside of the sequence that do not clobber the registers used within the sequence may be inserted in-between the instructions of the sequence (known as instruction scheduling).
```

riscv-elf.adoc
```diff
@@ -929,6 +939,54 @@ typedef struct
...
+- Instructions in the sequence with no data dependency may be reordered. For this sequence, the only instructions that can be reordered are `lw` and `addi`.
```
ilovepi, line 988:
```suggestion
- Instructions in the sequence with no data dependency may be reordered. In the preceding example, the only instructions that can be reordered are `lw` and `addi`.
```

riscv-elf.adoc
```diff
@@ -1521,6 +1579,89 @@ Relaxation result:
 ----
 --
 
+==== TLS Descriptors -> Initial Exec Relaxation
+
+Target Relocation:: R_RISCV_TLSDESC_HI20, R_RISCV_TLSDESC_LOAD_LO12_I, R_RISCV_TLSDESC_ADD_LO12_I, R_RISCV_TLSDESC_CALL
+
+Description:: This relaxation can relax a sequence loading address of a thread-local symbol reference into a GOT load instruction.
```
ilovepi, line 1586:
```suggestion
Description:: This relaxation can relax a sequence loading the address of a thread-local symbol reference into a GOT load instruction.
```

riscv-elf.adoc
```diff
@@ -1521,6 +1579,89 @@ Relaxation result:
...
+	lui a0, <pcrel-got-offset-for-symbol-hi>
+	{ld,lw} a0, <pcrel-got-offset-for-symbol-lo>(a0)
+----
+--
+
+==== TLS Descriptors -> Local Exec Relaxation
+
+Target Relocation:: R_RISCV_TLSDESC_HI20, R_RISCV_TLSDESC_LOAD_LO12_I, R_RISCV_TLSDESC_ADD_LO12_I, R_RISCV_TLSDESC_CALL
+
+Description:: This relaxation can relax a sequence loading address of a thread-local symbol reference into a thread-pointer-relative instruction sequence.
```
ilovepi, line 1622:
```suggestion
Description:: This relaxation can relax a sequence loading the address of a thread-local symbol reference into a thread-pointer-relative instruction sequence.
```

ishitatsuyuki:
Thanks Paul for the spec text suggestions, it reads a lot more natural now.

riscv-elf.adoc
```diff
@@ -929,6 +939,54 @@ typedef struct
...
+  unsigned long data[3]; /* defined by the dynamic linker */
+} tls_descriptor;
+----
+
+The `resolver` function is called with the address to the `tls_descriptor` itself,
+and returns `<address of TLS variable> - tp`.
+
+Up to 3 `unsigned long` may be stored inline within the descriptor. Dynamic linker
+implementations may use this to avoid a separate allocation to store data associated
+with the symbol.
```
ishitatsuyuki, line 965: To address a question from the last meeting (https://github.com/riscv-admin/psabi/blob/master/MINUTES/meeting-20230403.adoc):

glibc, bionic and BSD libc uses (ti_module, ti_offset, gen_count). musl uses (ti_module, ti_offset). Both cases should happily fit in the 3-slot specification.

riscv-elf.adoc
```diff
@@ -929,6 +939,54 @@ typedef struct
...
+of Initial Exec when the library was not loaded dynamically with `dlopen`.
+
+For each `TLSDESC` relocation inside the GOT, the dynamic linker fills in the
+TLS descriptor entry as defined below:
+
+[,c]
+----
+typedef struct
+{
+  unsigned long (*resolver)(tls_descriptor *);
```
ishitatsuyuki, line 955: glibc uses `ptrdiff_t` for the return signature in its implementation. Should we use `ptrdiff_t` (better clarity that it's tp-offset) or `unsigned long` (simpler)?

rui314: I think `unsigned long` is a better choice because `long` is defined by https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc but `ptrdiff_t` isn't.

ishitatsuyuki: You're right.

riscv-elf.adoc
```diff
@@ -929,6 +939,54 @@ typedef struct
...
+The `resolver` function is called with the address to the `tls_descriptor` itself,
+and returns `<address of TLS variable> - tp`.
+
+Up to 3 `unsigned long` may be stored inline within the descriptor. Dynamic linker
+implementations may use this to avoid a separate allocation to store data associated
+with the symbol.
+
+The TLS descriptor `resolver` is called with a special calling convention where all
+registers all callee-saved, except `a0` which is used to pass the argument and `t0`
+which is used as the alternate link register.
```
ishitatsuyuki, line 969: Question: Should vector registers be caller or callee saved? Callee save is better for the fast path, but for the slow path it can suffer from over-conservative register saving.

On x86, SSE/AVX registers follow standard calling convention instead.  
On AArch64, Q (NEON) registers are callee saved, but SVE registers follow standard calling convention instead.

If we're gonna follow suit, then vector registers will be caller saved as in the provisional Vector ABI docs.

riscv-elf.adoc
kito-cheng: Defined by #377 now, and `ptrdiff_t` might be more portable when we have ilp32 on rv64?

jrtc27: ILP32 by definition has 32-bit long so that makes no difference (that's why there's an L before the 32 in there)

kito-cheng: oh yeah, you are right, thanks :)

ishitatsuyuki: Yeah, I think we should keep the `long`. As another point, the TP-offset arithmetic may wraparound — `ptrdiff_t` being signed is a minor annoyance as signed integer overflow is technically UB.

riscv-elf.adoc
kito-cheng: Is there a guarantee that the TLS resolver will be a leaf function? vector register should be all caller-save *IF* it's not leaf function since `mem*` or `str*` might use vector version, and you never know the which function has call them or not, then you still need to save all register before any function call in TLS resolver.

ishitatsuyuki: The fast path is a leaf function while the slow path almost definitely is not. The resolver is written in assembly and conditionally saves registers if it needs to enter the slow path. In that sequence, we could save vector registers as well, if callee-saved vector registers is desired.

In glibc the slow path is only hit once per (module / thread combination) to allocate the storage, so actually it might make more sense to make vector regs callee saved as well.

kito-cheng: Sound like it worth to having a special calling convention, I guess one challenge here is how to detect the presence of vector extensions, but anyway those stuffs should be addressed on upstream glibc soon by hw probing.

ishitatsuyuki:
I wonder if the use of labels might pose inconvenience for compiler implementations. I'm currently trying to implement this on GCC, but AFAIU the way to introduce labels in GCC also creates a jump target, which in turn results in more blocks in control flow graph (and inhibits many optimizations).

Label-utilizing relocations also exists for `la` family of instructions, but they can be handled by assembler in those cases since scheduling is not needed there. Doing the same for TLSDESC means giving up on scheduling opportunities as well as regalloc flexibility in some cases, but both might be negligible after all.

If anyone here is familiar with LLVM or other compilers, what do you think?
- Does the compiler have a way to introduce labels that do not affect control flow?
- If labels are un-ergonomic, is there a better way to convey information necessary for relaxation?

jrtc27:
GCC's opaque blocks of textual assembly for instructions is pretty terrible if that's what you're referring to, yes. LLVM does not suffer from such a limitation. It will gladly split up things like LLA into its constituent instructions and optimise them (https://godbolt.org/z/53xobnvxz). I don't know if there are mechanisms to do similar things in GCC, but if not it's long overdue.

ishitatsuyuki:
Rebased. As a note for anyone who is potentially implementing, the relocation magic numbers have been shifted since some numbers have been taken by other spec updates.

Other changes:
- Clarified that vector registers are also callee-saved.
- `.tlsdesc_call` uses a directive instead of assembler function, since it does not relate to a immediate constant within the instruction.

riscv-elf.adoc
```diff
@@ -929,6 +939,54 @@ typedef struct
...
+Example assembler load and store of a thread local variable `i` using the `%tlsdesc_hi`, `%tlsdesc_lo_add`, `%tlsdesc_lo_load` and `%tlsdesc_call`
+assembler functions. The emitted relocations are in the comments.
+
+[,asm]
+----
+label:
+	auipc tX, %tlsdesc_hi(symbol)         // R_RISCV_TLSDESC_HI20 (symbol)
+	lw    tY, tX, %tlsdesc_lo_load(label) // R_RISCV_TLSDESC_LOAD_LO12_I (label)
+	addi  a0, tX, %tlsdesc_lo_add(label)  // R_RISCV_TLSDESC_ADD_LO12_I (label)
+	jalr  t0, tY                          // R_RISCV_TLSDESC_CALL (label)
```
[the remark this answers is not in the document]
ishitatsuyuki: Changed back to `%tlsdesc_call` and it uses a syntax consistent with what we're already doing in `%tprel_add`.

ishitatsuyuki:
One small spec update, the %tlsdesc_lo family functions are unified since the assembler can determine the relocation to use from opcode. It also fixes some errors in the spec where %tlsdesc_lo_load and %tlsdesc_load_lo were both appearing in examples.

I have my WIP glibc tree here: https://github.com/ishitatsuyuki/glibc/tree/rv-tlsdesc. It doesn't handle vector calling conventions yet (I'm generally unsure about what's the state of vector integration in toolchain). For testing, I hand crafted some assembly and compiled it with a quickly hacked around GNU assembler, and linked with mold (linker support is required to generate the GOT entries, code is at https://github.com/rui314/mold/pull/1041). It should work with both static TLS and dlopen TLS, but if there are bugs please let me know (I should also be able to do in depth tests after I finish work on GCC side).

ilovepi:
I've uploaded an implementation in LLVM to phabricator.

https://reviews.llvm.org/D151458  
https://reviews.llvm.org/D151462  
https://reviews.llvm.org/D151464

ilovepi:
> One small spec update, the %tlsdesc_lo family functions are unified since the assembler can determine the relocation to use from opcode. It also fixes some errors in the spec where %tlsdesc_lo_load and %tlsdesc_load_lo were both appearing in examples.
> 

I'm not sure that's desirable. Just because the assembler can determine this, I don't see how it improves anything over just emitting the correct reloc directly. Can you elaborate on why you think this is a better approach?

But, even if we do want to go this route, I would expect the directive should be something other than `tlsdesc_lo` if you're gong to merge the add into it.

ilovepi:
> > I'm not sure that's desirable. Just because the assembler can determine this, I don't see how it improves anything over just emitting the correct reloc directly. Can you elaborate on why you think this is a better approach?
> 
> It's more consistent with existing relocations like `tprel_lo` and `pcrel_lo`, although in that case there's no semantic difference. AArch64 also uses the same `:tlsdesc_lo12` syntax within the assembler for `ldr` and `add`.
> 

we have `TPREL_ADD`, don't we?

> > But, even if we do want to go this route, I would expect the directive should be something other than tlsdesc_lo if you're gong to merge the add into it.
> 
> Why? Considering the examples above, `tlsdesc_lo` feels like the name that is most in line with others.

Because if you fuse it w/ the add, it's no longer solely about the low bits?

ishitatsuyuki:
> we have TPREL_ADD, don't we?

`%tprel_add` is different though, because the relocation is purely for relaxation purpose, and it does not affect the low bits. We cannot call it `%tprel_lo` in that case.

> Because if you fuse it w/ the add, it's no longer solely about the low bits?

I agree that encoding the underlying relocation (and more importantly relaxation) semantics into the directive makes sense. Re-thinking about this, I think having different name is fine in terms of consistency too (since the existing I/S variants does not affect relaxation semantics). It will still diverge from AArch64, but that's probably fine.

Does `tlsdesc_load_lo` and `tlsdesc_add_lo` sound good to you? I see in your patch that `tlsdesc_lo` is used for the former, but I'm not sure if that's intentional.

ilovepi:
> > we have TPREL_ADD, don't we?
> 
> `%tprel_add` is different though, because the relocation is purely for relaxation purpose, and it does not affect the low bits. We cannot call it `%tprel_lo` in that case.
> 

Ah, that's a good point. 

> > Because if you fuse it w/ the add, it's no longer solely about the low bits?
> 
> I agree that encoding the underlying relocation (and more importantly relaxation) semantics into the directive makes sense. Re-thinking about this, I think having different name is fine in terms of consistency too (since the existing I/S variants does not affect relaxation semantics). It will still diverge from AArch64, but that's probably fine.
> 

I think the RISC-V psABI seems to diverge from AArch64 whenever it makes sense, so I agree that's probably OK.

> Does `tlsdesc_load_lo` and `tlsdesc_add_lo` sound good to you? I see in your patch that `tlsdesc_lo` is used for the former, but I'm not sure if that's intentional.

I'm happy to change that, since it was just an oversite on my part.  I'll update those a bit later today.

ishitatsuyuki:
From https://github.com/rui314/mold/pull/1041#issuecomment-1667590523, we are considering dropping the 4-slot TLSDESC entry in favor of the traditional indirect layout to avoid diverging cross-platform code in toolchains. (It's technically more optimized, but arguably not significant enough to justify the divergence. The extra slots are also wasted if using an initial-exec type resolver.)

For the reference, the current proposed layout looks like this:

```
typedef struct
{
  unsigned long (*resolver)(tls_descriptor *);
  unsigned long data[3]; /* defined by the dynamic linker */
} tls_descriptor;
```

While the traditional layout looks like this:
```
typedef struct
{
  unsigned long (*resolver)(tls_descriptor *);
  void *arg;
}
```

If there are no objections, I'll be updating the spec and prototype implementations to use the traditional layout in the following days.

enh-google:
> If there are no objections, I'll be updating the spec and prototype implementations to use the traditional layout in the following days.

sgtm for Android!

ilovepi:
From our perspective, having 3 GOT slots (2 private after the call target) is a worthwhile improvement, and I'm not aware of any downsides to that approach. Existing implementations can ignore the extra slot to maintain consistency with other machine, but it's arguable that every machine would have been better off w/ a 3-slot ABI. In our experience, it's not that much work to take advantage of it for RISC-V, and leads to a simpler implementation. Even the conversation in the mold issue linked above mentions that benefit. 

Existing implementations using two slots have to allocate a two-word object dynamically when doing the relocation for non-static-TLS cases and also pay for the indirection on every access. With three slots, you don't need to use a pointer to two words in the second slot.  In our view, we'd prefer to move towards a simpler, more efficient implementation rather than try to conform to the status quo. This seems doubly true, since so much of the implementation here is specific to the target and are not shared anyway.

enh-google:
yeah, to be clear i don't have a strong opinion either way --- "some TLSDESC is better than no TLSDESC", and i'm actually more concerned that getting _any_ TLSDESC will take long enough that we won't be able to assume it in Android's riscv64 ABI. my _real_ vote is "whichever we can have quicker, please" :-)

rui314:
As a person who ported the mold linker to 10+ different targets, I wouldn't make that kind of change to deviate from the other psABIs unless there's a strong reason to do so. In this case, I think the performance benefit is not proved and therefore not strong enough to deviate from the other psABIs.

In fact, my guess is it could be a slight negative impact in terms of performance. For the local exec and initial exec (i.e. if a thread-local variable is in the main executable or an initially-loaded DSO), we would use only the first two GOT slots to store a function pointer and a TP-relative offset. The third and fourth slot would be just a waste of memory. The only case that we need the two extra slots is for dloped'ed DSO. But for dlopen'ed libraries, the runtime does complex things such as allocating a TLS block lazily anyway, so I doubt if placing the extra slots in .got will make any measurable difference. I also want to point out that dlopen'ed library is not the most common use case -- most TLS variables are in the main executable or loaded-on-process-startup DSOs, in which case extra slots are not used.

I'd stick with the usual two-slot design.

kito-cheng:
I don't have strong opinion here since I not a real linker expert :P, so I guess we need to call @MaskRay, and maybe @JeffreyALaw.

@MaskRay: sorry for bother you (again :P), I know you are out of town, but I think you are the best guy to review this.

enh-google:
> I also want to point out that dlopen'ed library is not the most common use case -- most TLS variables are in the main executable or loaded-on-process-startup DSOs, in which case extra slots are not used.

fwiw, dlopen() _is_ the common case on Android --- what you think of as an "app" is actually a .so loaded into an existing clone of a "zygote" process (so you effectively only pay for all the runtime startup once on boot, not on every app start).

but, yes, from a pragmatic perspective i assume that "do what the others do" gets us TLSDESC sooner, and with the fewest bugs.

rui314:
@enh-google I think you guys can improve the TLS access performance for dlopen'ed libs on Android with TLSDESC with the two-slot approach. Here is an idea: you over-allocate the initial TLS block when the "zygote" process starts up and place dlopen'ed libraries' TLS blocks to the remaining space of the initial TLS block instead of to a separate TLS block. Then you can set a function pointer and a TP-relative offset to the GOT slots in the dlopen'ed DSO, which should make the TLVs in dlopen'ed DSO to work as fast as ones in a non-dlopen'ed DSO. That's a separate topic, though.

enh-google:
> @enh-google I think you guys can improve the TLS access performance for dlopen'ed libs on Android with TLSDESC with the two-slot approach. Here is the idea: you over-allocate the initial TLS block when the "zygote" process starts up and place dlopen'ed libraries' TLS blocks to the remaining space of the initial TLS block instead of to a separate TLS block. Then you can set a function pointer and a TP-relative offset to the GOT slots in the dlopen'ed DSO, which should make the TLVs in dlopen'ed DSO to work as fast as ones in a non-dlopen'ed DSO. That's a separate topic, though.
> 
> (As long as you do not touch the over-allocated unused TLS block, that wouldn't use the physical memory, so I don't think that would waste any resources. With the 64-bit address space, I believe you can reserve plenty of space at the end of .tbss for future dlopen calls.)

@rprichard for that.

but just having TLSDESC for riscv64 from day one would put riscv64 in a better position than the other Android architectures, regardless of the specific details. so i continue to be firmly in the "what can i have soonest?" camp (as long as it's not _worse_ than the others) :-)

ishitatsuyuki:
GNU toolchain patches posted:
- [glibc](https://sourceware.org/pipermail/libc-alpha/2023-August/150968.html)
- [binutils](https://sourceware.org/pipermail/binutils/2023-August/129076.html)
- [gcc](https://gcc.gnu.org/pipermail/gcc-patches/2023-August/627788.html)

riscv-elf.adoc
```diff
@@ -940,6 +950,55 @@ typedef struct
...
+TLS Descriptors (TLSDESC) are an alternative implementation of the Global Dynamic model
+that allows the dynamic linker to achieve performance close to that
+of Initial Exec when the library was not loaded dynamically with `dlopen`.
+
+For each `TLSDESC` relocation inside the GOT, the dynamic linker fills in the
+TLS descriptor entry as defined below:
+
+[,c]
+----
+typedef struct
```
MaskRay, line 964: I expect that the section mentions that we reserve a consecutive pair of pointer-sized entry in the GOT for each `TLSDESC(S+A)`.

riscv-elf.adoc
MaskRay: I think it's useful to generalize the register as it's important for relaxation.

riscv-elf.adoc
```diff
@@ -1540,6 +1599,89 @@ Relaxation result:
 ----
 --
 
+==== TLS Descriptors -> Initial Exec Relaxation
+
+Target Relocation:: R_RISCV_TLSDESC_HI20, R_RISCV_TLSDESC_LOAD_LO12_I, R_RISCV_TLSDESC_ADD_LO12_I, R_RISCV_TLSDESC_CALL
+
+Description:: This relaxation can relax a sequence loading the address of a thread-local symbol reference into a GOT load instruction.
+
+Condition:: None.
```
MaskRay, line 1608: Here should state the condition when the relaxation is applicable. (producing an executable)

riscv-elf.adoc
```diff
@@ -1540,6 +1599,89 @@ Relaxation result:
...
+
+==== TLS Descriptors -> Local Exec Relaxation
+
+Target Relocation:: R_RISCV_TLSDESC_HI20, R_RISCV_TLSDESC_LOAD_LO12_I, R_RISCV_TLSDESC_ADD_LO12_I, R_RISCV_TLSDESC_CALL
+
+Description:: This relaxation can relax a sequence loading the address of a thread-local symbol reference into a thread-pointer-relative instruction sequence.
+
+Condition::
+
+- For the short form: Offset between thread-pointer and thread-local symbol is within +-2KiB.
```
MaskRay, line 1646: Here should state the condition when the relaxation is applicable. (producing an executable and defined symbol)

MaskRay:
TLSDESC is designed to optimize for the static case. Its dynamic case is slower than TLS GD for every access.
Paying a one-shot `dlopen` overhead for each symbol in rtld doesn't look terribly bad for me.
Adopting the traditional two slot mechanism is the correct approach.

richfelker:
> TLSDESC is designed to optimize for the static case. Its dynamic case has quite a bit overhead, similar to the traditional `__tls_get_addr`, just without the calling convention overhead. The TLS accesses dominate the overhead of one symbol. Paying a one-shot `dlopen` overhead for each symbol in rtld doesn't look terribly bad for me. Adopting the traditional two slot mechanism is the correct approach.

That's the case for glibc but not for musl, where the dynamic case is also very efficient and dominated by the call and indirection layers overhead. I would love to see the ABI reserve 3 or 4 slots in the GOT just so implementations have the freedom to do this more efficiently if they want to.

Of course, as was discussed on #musl, there's already some such freedom: for example, on 64-bit archs, if both modid and offset fit in 32 bits, we could repurpose the 2 halves of the second slot as modid/offset and use a variant of the function without the extra indirection layer that loads them as 32-bit values. The same is possible on 32-bit for cases that fit in 8/24 bits or 16/16, but the narrow accesses and width extension might be sufficiently costly to offset any benefit.

rui314:
@richfelker Did you measure how much faster/slower it actually is? As we discussed above, you need to demonstrate how fast the TLV accesses could be with one or two extra GOT slots and that performance benefit outperforms the cost of the extra GOT slots _and_ the cost of diverging from the other psABIs.

I'm also not convinced that decomposing 8/24 bits or 16/16 is costly. Obviously, you can just read two 16-bit values for 16/16. Decomposing a 32 bit value into 8 and 24 signed values needs only a few integer arithmetic machine instructions.

rui314:
The spec looks good to me.

rui314:
Now that we have a set of working patches for the GNU toolchain. Can this proposal be ratified?

ishitatsuyuki:
I think we're pending maintainer approval, per [policy](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/policy.md).

- For binutils, I have [v2](https://inbox.sourceware.org/binutils/20230831171345.49052-1-ishitatsuyuki@gmail.com/) pending review.
- For gcc and glibc, I'll work on a v2 with possibly more tests included.

rui314:
My understanding is that we need a PoC implementation, which we already have, and approvals from LLVM/GCC/binutils/LLD on this proposal. Or do we need approvals on the implementation? I wanted this to be ratified because I wanted to merge the patch to the mold linker but that's not safe to do unless we at least reserve the relocation numbers.

ishitatsuyuki:
As Rui said, does anyone know what kind of approval or ratification do we need on this to move forward?

For the reference, I posted v2 patches for v2 posted, [gcc](https://inbox.sourceware.org/gcc-patches/20230908104923.31154-1-ishitatsuyuki@gmail.com/) and [glibc](https://inbox.sourceware.org/libc-alpha/20230908105500.32072-1-ishitatsuyuki@gmail.com/) too. There are no extra tests, as I believe gcc already covers TLS with portable C tests and glibc is not testable alone.

kito-cheng (approved):
@rui314 @ishitatsuyuki Yes, we only require PoC for the proposal, so this proposal is almost ready to go, but before merge I would like to have @MaskRay's LGTM :)
And could you squash commits into one?
but anyway give my LGTM here, thanks for contributing this, this has discussed since 2019, and now we finally have it!!! 

@maskRay Could you take a look to see any comments?

ishitatsuyuki:
I squashed and rebased the changes. Thanks a lot, Kito, for your review and your help!

enh-google:
> I squashed and rebased the changes. Thanks a lot, Kito, for your review and your help!

yeah, thanks kito-cheng, and everyone else who contributed, but thanks especially to ishitatsuyuki for driving the initial implementation!

MaskRay (approved):

kito-cheng:
Merged, @ishitatsuyuki thanks!

ishitatsuyuki:
I posted [v3 of the glibc patch](https://inbox.sourceware.org/libc-alpha/20230913172633.39229-1-ishitatsuyuki@gmail.com/) with a few critical bug fixes discovered during running the mold test suite.

PR #374, opened 2023-03-28
Define s11(x27) as a platform specific register

### LINKED ISSUE #370: Specify a platform reserved register

ilovepi:
The Procedure Call Standard for ARM(AArch64) defines x18 as a "Platform Register" if needed. This is useful for platforms to implement certain inter-procedural state. For instance, x18 is used on AArch64 as the shadow call stack register. Today, RISC-V has no analogous concept, but features like shadow call stack have already been implemented in LLVM using x18, which is problematic on RISC-V, since it prevents save/restore ranges from working as expected. Defining s11 as a platform register allows platforms to use features like shadow call stack without conflicting with planed feature like save/restore libcalls. 

resolves #370

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -31,6 +31,12 @@ gp, because signal handlers may rely upon their values.
 The presence of a frame pointer is optional.  If a frame pointer exists,
 it must reside in x8 (s0); the register remains callee-saved.
 
+Platforms are free to reserve `s11(x27)` for use as a platform register. If a
+platform has need for a dedicated general-purpose register, then it should use
+`s11` register for that purpose. Should the platform forgo using `s11` in this
+role, then it should use `s11` as an additional callee-saved register. The
+platform ABI specification must document the use of this register.
+
 === Floating-point Register Convention
 
 .Floating-point register convention


ilovepi:
https://reviews.llvm.org/D146463 is a proposed change to fix the shadow call stack register in LLVM.

https://discourse.llvm.org/t/rfc-psa-changing-the-shadow-call-stack-register-on-risc-v/69537 is a related discussion on LLVM discourse.

appujee:
There was a discussion at the toolchain-sig today and people were of the view that re-purposing gp is better patch forward. https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/371

ilovepi:
I'm pretty sure we're going to go w/ `gp` as the platform register, but I'll leave this open until we finalize that.

ilovepi:
Since it seems that we will be using `x3` instead, I'm going to close this PR now, in favor of #371.

PR #375, opened 2023-04-03
Add placeholder for runtime library ABI

kito-cheng:
It's a placeholder of runtime library ABI, this document plan to writing down the behavior and naming scheme for current compiler helper funciton implementation first.


kito-cheng:
it's just a placeholder, so gonna merge this for now,

PR #377, opened 2023-04-14
Define size_t and ptrdiff_t

### LINKED ISSUE #376: Should calling convention also define ptrdiff_t?
[cc-type-sizes-and-alignments](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc#cc-type-sizes-and-alignments) has the definitions of types for both IPL32 and LP64 systems. Would it help to define the size of `ptrdiff_t`?

MaskRay:
Close #376

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -426,6 +426,10 @@ A null pointer (for all types) has the value zero.
 corresponding real type (`float`, `double`, or `long double`), with the first
 member holding the real part and the second member holding the imaginary part.
 
+The type `size_t` is defined as `unsigned int` for RV32 and `unsigned long` for RV64.
+
+The type `ptrdiff_t` is defined as `int` for RV32 and `long` for RV64.
+
 === va_list, va_start, and va_arg
 
 The `va_list` type is `void*`. A callee with variadic arguments is responsible


riscv-cc.adoc
```diff
@@ -426,6 +426,10 @@ A null pointer (for all types) has the value zero.
 corresponding real type (`float`, `double`, or `long double`), with the first
 member holding the real part and the second member holding the imaginary part.
 
+The type `size_t` is defined as `unsigned int` for RV32 and `unsigned long` for RV64.
+
+The type `ptrdiff_t` is defined as `int` for RV32 and `long` for RV64.
```
appujee, line 431: Should we put them in table above as well? I feel like it is easier to follow in a table.

jrtc27: They're not basic types, so no

appujee (approved):
LTGM, but i'd let other reviewers take the final call.

kito-cheng (approved):
LGTM, this is document what we have, checked it consistent with clang and GCC.

kito-cheng:
@appujee @MaskRay thanks!

PR #380, opened 2023-05-18
[Vector ABI] Type size and alignment for vector types.

### LINKED ISSUE #347: ABI alignment of vector types
We should define the ABI alignment of (scalable) vector types, that could be separated sub-item from the full vector ABI.

Vector extension only require vector load/store align to the element width, e.g. require 8 byte alignment for element width=64, however some RISC-V core implementation might require larger alignment for best performance.

So we have following options:

1. Alignment as the element width.
2. Require higher alignment for vector types, e.g. align to 8 byte or 16 byte for all vector type for any element width.

And we have LMUL in the vector extension, we might also consider that in the alignment if needed.

NOTE: ABI alignment is the minimal requirement, compiler/programmer could set that alignment to larger than the ABI alignment is also conformance to the ABI

kito-cheng:
The issue of Vector alignment is discussed in #347. It is mentioned that aligning to 128 bits might deliver better performance on some RISC-V cores, but this behavior could lead to considerable stack wastage on zve32 and zve64 cores. For instance, in order to ensure a vector value in the stack conforms to the ABI specification, we could potentially waste up to 96 bits per vector object in stack for zve32, and the performance difference isn't always evident across all core implementations.

Therefore, this proposal sets the alignment of vector types to element alignment, to avoid wasting a significant amount of stack space in zve32 and zve64 configurations. Also, since the ABI only specify the minimum alignment and doesn't limit the compiler from adopting higher alignment for specific CPUs.

Fix #347.

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -626,6 +626,38 @@ of the vararg save area.  The `va_arg` macro will increment its `va_list`
 argument according to the size of the given type, taking into account the
 rules about 2×XLEN aligned arguments being passed in "aligned" register pairs.
 
+=== Vector type sizes and alignments
+
+This section defines the sizes and alignments for the vector types defined in
+the _RISC-V Vector Extension Intrinsic Document_<<rvv-intrinsic-doc>>.
+The actual size of each type is determined by the hardware configuration, which
+is based on the content of the `vlenb` register.
+
+There are three classes of vector types: the vector mask types, the vector
+data types and the vector tuple types.
+
+.Type sizes and alignments for vector mask types
+[cols="4,3,>3,>2"]
+[width=80%]
+|===
+| Internal Name              | Type                 | Size (Bytes)       | Alignment (Bytes)
+
+| __rvv_vbool1_t             | vbool1_t             |  VLENB             |  1
+| __rvv_vbool2_t             | vbool2_t             |  VLENB / 2         |  1
+| __rvv_vbool4_t             | vbool4_t             |  VLENB / 4         |  1
+| __rvv_vbool8_t             | vbool8_t             |  ceil(VLENB / 8)   |  1
+| __rvv_vbool16_t            | vbool16_t            |  ceil(VLENB / 16)  |  1
+| __rvv_vbool32_t            | vbool32_t            |  ceil(VLENB / 32)  |  1
+| __rvv_vbool64_t            | vbool64_t            |  ceil(VLENB / 64)  |  1
+|===
+
+include::vector_type_infos.adoc[]
+
+NOTE: The vector mask types utilize a portion of the space, while the remaining
+content may be undefined, both in the register and in memory.
+
+NOTE: Size must be a positive integer.
+
 [appendix]
 == Linux-specific ABI
 
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -142,12 +142,45 @@ any vector registers.
 
 {Cpp} name mangling for RISC-V follows
 the _Itanium {Cpp} ABI_ <<itanium-cxx-abi>>;
-there are no RISC-V specific mangling rules.
+plus mangling for RISC-V vector data types and vector mask types,
+which are defined in the following section.
 
 See the "Type encodings" section in _Itanium {Cpp} ABI_
 for more detail on how to mangle types. Note that `__bf16` is mangled in the
 same way as `std::bfloat16_t`.
 
+=== Name Mangling for Vector Data Types, Vector Mask Types and Vector Tuple Types.
+
+The vector data types and vector mask types, as defined in the section
+<<Vector type sizes and alignments>>, are treated as vendor-extended types in
+the _Itanium {Cpp} ABI_ <<itanium-cxx-abi>>. These mangled name for
+these types is `"u"<len>"__rvv_"<type-name>`. Specifically,
+prefixing the type name with `__rvv_`, which is prefixed by
+a decimal string indicating its length, which is prefixed by "u".
+
+For example:
+
+[,c]
+----
+    void foo(vint8m1_t x);
+----
+
+is mangled as
+[,c]
+----
+    _Z3foou15__rvv_vint8m1_t
+----
+[source,abnf]
+----
+mangled-name = "u" len "__rvv_" type-name
+
+len = nonzero *DIGIT
+nonzero = "1" / "2" / "3" / "4" / "5" / "6" / "7" / "8" / "9"
+
+type-name = identifier-nondigit *identifier-char
+identifier-nondigit = ALPHA / "_"
+identifier-char = identifier-nondigit / "_"
+----
 == ELF Object Files
 
 The ELF object file format for RISC-V follows the
@@ -1977,3 +2010,6 @@ RISC-V International.
 
 * [[[riscv-zc-extension-group]]] "ZC* extension specification"
 https://github.com/riscv/riscv-code-size-reduction
+
+* [[[rvv-intrinsic-doc]]] "RISC-V Vector Extension Intrinsic Document"
+https://github.com/riscv-non-isa/rvv-intrinsic-doc
--- a/Makefile
+++ b/Makefile
@@ -19,3 +19,6 @@ $(NAME).pdf: $(NAME).adoc $(wildcard *.adoc) resources/themes/risc-v_spec-pdf.ym
 	    -a pdf-fontsdir=resources/fonts \
 	    -v \
 	    $< -o $@
+
+regen_vector_type_infos:
+	python3 gen_vector_type_infos.py > vector_type_infos.adoc
--- a/gen_vector_type_infos.py
+++ b/gen_vector_type_infos.py
@@ -1,0 +1,112 @@
+import subprocess
+import os
+import sys
+import math
+
+LMUL=["mf8", "mf4", "mf2", "m1", "m2", "m4", "m8"]
+NF=[2, 3, 4, 5, 6, 7, 8]
+SEW=[8, 16, 32, 64]
+TYPES = ["int", "uint", "float", "bfloat"]
+
+def lmul2nreg(lmul):
+    if (lmul.startswith("mf")):
+        return 1
+    else:
+        return int(lmul[1:])
+
+def lmul2num(lmul):
+    if (lmul.startswith("mf")):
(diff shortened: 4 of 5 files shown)


kito-cheng:
cc. @rofirrim @aswaterman @zhongjuzhe @JeffreyALaw @topperc @preames

riscv-elf.adoc
```diff
@@ -142,11 +142,34 @@ any vector registers.
...
 
 See the "Type encodings" section in _Itanium {Cpp} ABI_
 for more detail on how to mangle types.
 
+=== Name Mangling for Vector Data Types and Vector Masking Types
+
+The vector data types and vector masking types, as defined in the section
+<<Vector type sizes and alignments>>, are treated as vendor-extended types in
+the _Itanium {Cpp} ABI_ <<itanium-cxx-abi>>. The mangled name for these types
+employs the internal name, which is prefixed by `u`, and then followed by the
```
rofirrim, line 157: Is the internal name specified somewhere? I failed to find it in the rvv intrinsics doc.

If it is a compiler implementation-specific one we might risk incompatibilities, e.g, an object built using `g++` and then linking it with another file built using `clang++`. Not sure if this is a desirable scenario.

rofirrim (commented):
Thanks @kito-cheng for this.

> The issue of Vector alignment is discussed in https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/347. It is mentioned that aligning to 128 bytes might deliver better performance on some RISC-V cores,

I think you meant 128 bits? (later on you say 96 bits which would match with bits).

Just to confirm the intent of this change: this is always minimal alignment and a compiler implementation/platform may choose to use a larger value. Also, a user could use `__attribute__((aligned(16)))` (or similar mechanisms) to enforce larger alignments.

Now this probably only impacts stack variables but if we ever have aggregates of vector type, the alignment will be used there when computing offsets. Maybe I'm missing some other case.

riscv-elf.adoc
kito-cheng: I missed the internal on the table, currently gcc and clang are using same internal name, so we don't suffer any ABI breakage on either compiler on this :)

kito-cheng: Updated

kito-cheng:
> I think you meant 128 bits? (later on you say 96 bits which would match with bits).

Yeah, should be `bits`.

> Just to confirm the intent of this change: this is always minimal alignment and a compiler implementation/platform may choose to use a larger value. Also, a user could use __attribute__((aligned(16))) (or similar mechanisms) to enforce larger alignments.
> 
> Now this probably only impacts stack variables but if we ever have aggregates of vector type, the alignment will be used there when computing offsets. Maybe I'm missing some other case.

Yes, compiler always has freedom to use bigger/higher alignment, and this mostly affect stack object, another one that will be affected is `attribute(riscv_rvv_vector_bits(N))`[1], since with that attribute, rvv type could also live as global variable and live within struct.

[1] https://reviews.llvm.org/D145088

zhongjuzhe:
>> | __rvv_vbool1_t             | vbool1_t             |  VLENB        |  1
>>| __rvv_vbool2_t             | vbool2_t             |  VLENB        |  1
>>| __rvv_vbool4_t             | vbool4_t             |  VLENB        |  1
>>| __rvv_vbool8_t             | vbool8_t             |  VLENB        |  1
>>| __rvv_vbool16_t            | vbool16_t            |  VLENB        |  1
>>| __rvv_vbool32_t            | vbool32_t            |  VLENB        |  1
>>| __rvv_vbool64_t            | vbool64_t            |  VLENB        |  1

I understand LLVM mask type size is adjusted as doc said.

However, GCC byte size of these mask type is not adjusted like this.

GCC configuration for BOOL type
vbool1_t VLENB
vbool2_t VLENB / 2
vbool4_t VLENB / 4
vbool8_t VLENB / 8
vbool16_t CEIL (VLENB / 16)
vbool32_t CEIL (VLENB / 32)
vbool64_t CEIL (VLENB / 64)

I am not sure whether this inconsistency between GCC and LLVM is necessary to be documented in the ABI doc.

Otherwise, LGTM. Thanks.

kito-cheng:
> > > | __rvv_vbool1_t             | vbool1_t             |  VLENB        |  1
> > > | __rvv_vbool2_t             | vbool2_t             |  VLENB        |  1
> > > | __rvv_vbool4_t             | vbool4_t             |  VLENB        |  1
> > > | __rvv_vbool8_t             | vbool8_t             |  VLENB        |  1
> > > | __rvv_vbool16_t            | vbool16_t            |  VLENB        |  1
> > > | __rvv_vbool32_t            | vbool32_t            |  VLENB        |  1
> > > | __rvv_vbool64_t            | vbool64_t            |  VLENB        |  1
> 
> I understand LLVM mask type size is adjusted as doc said.
> 
> However, GCC byte size of these mask type is not adjusted like this.
> 
> GCC configuration for BOOL type vbool1_t VLENB vbool2_t VLENB / 2 vbool4_t VLENB / 4 vbool8_t VLENB / 8 vbool16_t CEIL (VLENB / 16) vbool32_t CEIL (VLENB / 32) vbool64_t CEIL (VLENB / 64)
> 
> I am not sure whether this inconsistency between GCC and LLVM is necessary to be documented in the ABI doc.
> 
> Otherwise, LGTM. Thanks.

Yeah, I know GCC has did some more compact way, that's the reason I add the note at the end:

`NOTE: The vector mask types utilize a portion of the space, while the remaining content may be undefined, both in the register and in memory.`

Thanks for review :)

kito-cheng:
Changes:
- Adjust size for vector mask types

kito-cheng:
The motivation of updating the vector mask type to more compact size, this allow compiler to store that in compact way without ABI violation.

kito-cheng:
ping @rofirrim @asb @topperc @preames

I would like to get some feedback from LLVM community before merge :)

rofirrim:
> ping @rofirrim @asb @topperc @preames
> 
> I would like to get some feedback from LLVM community before merge :)

No objections on my side, everything looks reasonable to me.

riscv-cc.adoc
```diff
@@ -524,6 +524,95 @@ of the vararg save area.  The `va_arg` macro will increment its `va_list`
...
+| __rvv_vint64m2_t     | vint64m2_t           |  VLENB * 2    |  8
+| __rvv_vuint64m2_t    | vuint64m2_t          |  VLENB * 2    |  8
+| __rvv_vfloat64m2_t   | vfloat64m2_t         |  VLENB * 2    |  8
+| __rvv_vint64m4_t     | vint64m4_t           |  VLENB * 4    |  8
+| __rvv_vuint64m4_t    | vuint64m4_t          |  VLENB * 4    |  8
+| __rvv_vfloat64m4_t   | vfloat64m4_t         |  VLENB * 4    |  8
+| __rvv_vint64m8_t     | vint64m8_t           |  VLENB * 8    |  8
+| __rvv_vuint64m8_t    | vuint64m8_t          |  VLENB * 8    |  8
+| __rvv_vfloat64m8_t   | vfloat64m8_t         |  VLENB * 8    |  8
+|===
```
topperc, line 612: Do we need to add bfloat types to this list?

kito-cheng: Added fp16 and bfloat16 related types :)

riscv-elf.adoc
```diff
@@ -142,12 +142,35 @@ any vector registers.
...
-there are no RISC-V specific mangling rules.
+there are no RISC-V specific mangling rules,
+except for the vector data types and vector masking types, which are defined in
+the following section.
 
 See the "Type encodings" section in _Itanium {Cpp} ABI_
 for more detail on how to mangle types. Note that `__bf16` is mangled in the
 same way as `std::bfloat16_t`.
 
+=== Name Mangling for Vector Data Types and Vector Masking Types
```
lhtin, line 153: Ditto.

riscv-elf.adoc
```diff
@@ -142,12 +142,35 @@ any vector registers.
...
+
+The vector data types and vector masking types, as defined in the section
```
lhtin, line 155: Ditto.

riscv-elf.adoc
```diff
@@ -142,12 +142,35 @@ any vector registers.
 
 {Cpp} name mangling for RISC-V follows
 the _Itanium {Cpp} ABI_ <<itanium-cxx-abi>>;
-there are no RISC-V specific mangling rules.
+there are no RISC-V specific mangling rules,
+except for the vector data types and vector masking types, which are defined in
```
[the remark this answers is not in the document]
kito-cheng: Updated :)

riscv-cc.adoc
```diff
@@ -524,6 +524,107 @@ of the vararg save area.  The `va_arg` macro will increment its `va_list`
...
```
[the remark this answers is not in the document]
kito-cheng: I was plan added in follow up PR, but I guess just added within this should reduce the review cycle :)

kito-cheng:
Changes:

- Add simple generator script for gen the table.
- Add tuple type.

workingjubilee (commented):
I was reviewing this and I had a few thoughts.

riscv-cc.adoc
```diff
@@ -524,6 +524,38 @@ of the vararg save area.  The `va_arg` macro will increment its `va_list`
...
+| Internal Name              | Type                 | Size (Bytes)       | Alignment (Bytes)
+
+| __rvv_vbool1_t             | vbool1_t             |  VLENB             |  1
+| __rvv_vbool2_t             | vbool2_t             |  VLENB / 2         |  1
+| __rvv_vbool4_t             | vbool4_t             |  VLENB / 4         |  1
+| __rvv_vbool8_t             | vbool8_t             |  ceil(VLENB / 8)   |  1
+| __rvv_vbool16_t            | vbool16_t            |  ceil(VLENB / 16)  |  1
+| __rvv_vbool32_t            | vbool32_t            |  ceil(VLENB / 32)  |  1
+| __rvv_vbool64_t            | vbool64_t            |  ceil(VLENB / 64)  |  1
```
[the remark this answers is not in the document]
kito-cheng: Updated, use VLEN instead now, thanks!

kito-cheng:
Going to merge this, thanks everyone :)

PR #386, opened 2023-06-07
Deprecate PT_RISCV_ATTRIBUTES

MaskRay:
#71 added a PT_RISCV_ATTRIBUTES segment to contain the SHT_RISCV_ATTRIBUTES section. The SHT_RISCV_ATTRIBUTES section does not have the SHF_ALLOC flag and is therefore not in a PT_LOAD segment. Dynamic loaders cannot assume the PT_RISCV_ATTRIBUTES part is present (it may happen to be mapped due to page alignment, but there is no guarantee).

To the best of my knowledge, PT_RISCV_ATTRIBUTES is the only open-source usage of a segment that contains a non-SHF_ALLOC section.

I think the existence of PT_RISCV_ATTRIBUTES just complicates analysis tools and binary manipulation tools like strip. E.g. if the SHT_RISCV_ATTRIBUTES section is removed, what should we do with PT_RISCV_ATTRIBUTES?

I think we should remove PT_RISCV_ATTRIBUTES, but ensure that the program header type is not reused by a future program header type.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1092,22 +1092,6 @@ The defined processor-specific section types are listed in <<rv-section-type>>.
 +++.riscv.jvt+++ is a linker-created section to store table jump
 target addresses. The minimum alignment of this section is 64 bytes.
 
-=== Program Header Table
-
-The defined processor-specific segment types are listed in <<rv-seg-type>>.
-
-[[rv-seg-type]]
-.RISC-V-specific segment types
-[cols="3,2,3"]
-[width=80%]
-|===
-| Name                 | Value       | Meaning
-
-| PT_RISCV_ATTRIBUTES  | 0x70000003  | RISC-V ELF attribute section.
-|===
-
-`PT_RISCV_ATTRIBUTES` describes the location of RISC-V ELF attribute section.
-
 === Note Sections
 
 There are no RISC-V specific definitions relating to ELF note sections.
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1092,6 +1092,26 @@ The defined processor-specific section types are listed in <<rv-section-type>>.
 +++.riscv.jvt+++ is a linker-created section to store table jump
 target addresses. The minimum alignment of this section is 64 bytes.
 
+=== Program Header Table
+
+The defined processor-specific segment types are listed in <<rv-seg-type>>.
+
+[[rv-seg-type]]
+.RISC-V-specific segment types
+[cols="3,2,3"]
+[width=80%]
+|===
+| Name                 | Value       | Meaning
+
+| PT_RISCV_ATTRIBUTES  | 0x70000003  | RISC-V ELF attribute section.
+|===
+
+`PT_RISCV_ATTRIBUTES` describes the location of RISC-V ELF attribute section.
+
+WARNING: `PT_RISCV_ATTRIBUTES` is deprecated. The build attributes section does
+not contain the `SHF_ALLOC` flag. Dynamic loaders cannot assume that the region
+described by `PT_RISCV_ATTRIBUTES` is present.
+
 === Note Sections
 
 There are no RISC-V specific definitions relating to ELF note sections.


jrtc27:
How is a run-time linker supposed to find it then in order to check compatibility if there's no program headers entry?

jrtc27:
I see two options:

1. Leave it as-is, and have run-time linkers go and explicitly map the segment somewhere temporarily
2. Make it SHF_ALLOC like all the other metadata sections needed at run time (which really don't need to be; things like .rela.dyn could quite easily not be and then just be mapped by the run-time linker as needed)

IMO 1 is totally fine though.

kito-cheng:
I think we might consider option 2, we don't have check anything from riscv attribute yet, but there is appeared some demand on that, https://github.com/riscv/riscv-cfi will need tags to the binary, and dynamic linker need to read those tags; we already has riscv-attribute, so IMO we should use riscv attribute.

But the we already define `SHT_RISCV_ATTRIBUTES` without `SHF_ALLOC` before, so that will require dynamic linker still implement option 1 way for existing programs.

MaskRay:
Making `SHT_RISCV_ATTRIBUTES` SHF_ALLOC would be quite unusual and deviate from existing build attributes practice. Perhaps we can consider removing `PT_RISCV_ATTRIBUTES` from GNU ld for now.

apazos:
I would like to pick up this conversation again. Was a decision made on marking it as allocatable or removing it from GNU ld / LLVM lld? It might be confusing for users that rely on linker scripts and do not expect the linker to be creating segments by default.

MaskRay:
> I would like to pick up this conversation again. Was a decision made on marking it as allocatable or removing it from GNU ld / LLVM lld? It might be confusing for users that rely on linker scripts and do not expect the linker to be creating segments by default.

The build attributes sections for other architectures do not have the `SHF_ALLOC` flag. Adding `SHF_ALLOC` would be very confusing.

shankarke:
is there any use of PT_RISCV_ATTRIBUTES segment ? I think it wastes few bytes in the program header space. 

Making it allocatable also may impact backwards compatibility. 

Overall I like this proposal of removing PT_RISCV_ATTRIBUTES.

MaskRay:
I am strongly in favor of removing PT_RISCV_ATTRIBUTES, which I believe is unused. I regretted that I added PT_RISCV_ATTRIBUTES (https://sourceware.org/bugzilla/show_bug.cgi?id=30237). The non-SHF_ALLOC section within a segment doesn't play well under `llvm-objcopy` as well.

jrtc27:
The current implementation may have issues, but the question remains: how does a run-time linker obtain information about the compatibility of libraries in the absence of such a segment?

apazos:
For static images the segment is not used. Entities like debuggers who need that info can simply read the it from the elf file.
For dynamic images it is yet to be seen how it will be used, but in that case, the segment would have to be loadable.
So should we only add to the spec and the tools when the dynamic use case is clear to us?

MaskRay:
> The current implementation may have issues, but the question remains: how does a run-time linker obtain information about the compatibility of libraries in the absence of such a segment?

The requirement of run-time capability doesn't seem that strong. I believe Arm's experience is that build attributes are only useful for static linking. A very small set of properties that really benefit from loader checking use GNU program properties. When SHT_AARCH64_ATTRIBUTES is introduced in the future, its spec will say that build Attributes are outside the scope of the specification.

jrtc27:
> > The current implementation may have issues, but the question remains: how does a run-time linker obtain information about the compatibility of libraries in the absence of such a segment?
> 
> The requirement of run-time capability doesn't seem that strong. I believe Arm's experience is that build attributes are only useful for static linking. A very small set of properties that really benefit from loader checking use GNU program properties. When SHT_AARCH64_ATTRIBUTES is introduced in the future, its spec will say that build Attributes are outside the scope of the specification.

Arm doesn't have the configurability and mutually-exclusive extensions that RISC-V has though. That's why we've ended up here, and the powers that be were very vocal that we needed a way to catch these issues.

MaskRay:
> > > The current implementation may have issues, but the question remains: how does a run-time linker obtain information about the compatibility of libraries in the absence of such a segment?
> > 
> > 
> > The requirement of run-time capability doesn't seem that strong. I believe Arm's experience is that build attributes are only useful for static linking. A very small set of properties that really benefit from loader checking use GNU program properties. When SHT_AARCH64_ATTRIBUTES is introduced in the future, its spec will say that build Attributes are outside the scope of the specification.
> 
> Arm doesn't have the configurability and mutually-exclusive extensions that RISC-V has though. That's why we've ended up here, and the powers that be were very vocal that we needed a way to catch these issues.

There might be such needs but the needs have to be satisfied using different mechanisms as non-SHF_ALLOC and SHF_ALLOC sections are very different. Being non-SHF_ALLOC makes SHT_RISCV_ATTRIBUTES not suitable for runtime checking. I'd also like to request the concrete needs.

kito-cheng (approved):
LGTM :)

PR #387, opened 2023-06-09
Introduce a new tag, `Tag_RISCV_x3_reg_usage` to record how gp/x3 begin …

kito-cheng:
…used.

This tag is designed to prevent the linker from merging objects that use `gp` for different purposes. Additionally, it can also serve to help users identify the usage within the executable.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1072,6 +1072,7 @@ non-standard ABI extensions.
 | Tag_RISCV_priv_spec                 |        8 | uleb128        | *Deprecated*, indicates the major version of the privileged specification.
 | Tag_RISCV_priv_spec_minor           |       10 | uleb128        | *Deprecated*, indicates the minor version of the privileged specification.
 | Tag_RISCV_priv_spec_revision        |       12 | uleb128        | *Deprecated*, indicates the revision version of the privileged specification.
+| Tag_RISCV_x3_reg_usage              |       16 | uleb128        | Indicates the usage definition of the X3 register.
 | Reserved for non-standard attribute | >= 32768 | -              | -
 |===
 
@@ -1176,6 +1177,27 @@ Merge policy:::
 The linker should report errors if object files of different privileged
 specification versions are merged.
 
+===== Tag_RISCV_x3_reg_usage, 14, uleb128=value
+
+Tag_RISCV_x3_reg_usage indicates the usage of `x3`/`gp` register. `x3`/`gp` could be used for
+global pointer relaxation, as a reserved platform register, or as a temporary register.
+
+[horizontal]
+0:: This object uses `x3` as a fixed register with unknown purpose.
+1:: This object uses `x3` as the global pointer, for relaxation purposes.
+2:: This object uses `x3` as the shadow stack pointer.
+3:: This object uses `X3` as a temporary register.
+4~1023:: Reserved for future standard defined platform register.
+1024~2047:: Reserved for nonstandard defined platform register.
+
+--
+
+Merge policy:::
+The linker should issue errors when object files with differing `gp` usage are
+combined. However, an exception exists: the value `0` can merge with `1` or `2`
+value. After the merge, the resulting value will be the non-zero one.
+
+--
 
 === Mapping Symbol
 
@@ -1387,7 +1409,8 @@ Relaxation result:
   load address of a symbol or load/store with a symbol reference into
   global-pointer-relative instruction.
 
-  Condition:: Offset between global-pointer and symbol is within +-2KiB,
+  Condition:: Global-pointer relaxation requires that Tag_RISCV_x3_reg_usage
+  must be 0 or 1, and offset between global-pointer and symbol is within +-2KiB,
   `R_RISCV_PCREL_LO12_I` and `R_RISCV_PCREL_LO12_S` resolved as indirect
   relocation pointer. It will always point to another `R_RISCV_PCREL_HI20`
   relocation, the symbol pointed by `R_RISCV_PCREL_HI20` will be used in
@@ -1441,6 +1464,8 @@ NOTE: The global pointer is referred to as the global offset table pointer in
 many other targets, however, RISC-V uses PC-relative addressing rather than
 access GOT via the global pointer register (`gp`), so we use `gp` register to
 optimize code size and performance of the symbol accessing.
+
+NOTE: Tag_RISCV_x3_reg_usage is treated as 0 if it is not present.
 
 ==== Zero-page Relaxation
 


riscv-elf.adoc
```diff
@@ -1176,6 +1177,27 @@ Merge policy:::
...
+===== Tag_RISCV_gp_reg_usage, 14, uleb128=value
+
+Tag_RISCV_gp_reg_usage indicate the usage of gp register, gp could be used for
+global pointer relaxation, reserved platform register or a temporary register.
+
+[horizontal]
+0:: This object use gp as fixed register with unknown purpose.
+1:: This object use gp as global pointer relaxation.
+2:: This object use gp as shadow stack pointer.
+3:: This object use gp as temporary register.
```
luismarques, line 1189:
```suggestion
0:: This object uses `gp` as a fixed register with unknown purpose.
1:: This object uses `gp` as the global pointer, for relaxation purposes.
2:: This object uses `gp` as the shadow stack pointer.
3:: This object uses `gp` as a temporary register.
```

riscv-elf.adoc
```diff
@@ -1176,6 +1177,27 @@ Merge policy:::
 The linker should report errors if object files of different privileged
 specification versions are merged.
 
+===== Tag_RISCV_gp_reg_usage, 14, uleb128=value
+
+Tag_RISCV_gp_reg_usage indicate the usage of gp register, gp could be used for
+global pointer relaxation, reserved platform register or a temporary register.
```
luismarques, line 1183:
```suggestion
Tag_RISCV_gp_reg_usage indicates the usage of `gp` register. `gp` could be used for
global pointer relaxation, as a reserved platform register or as a temporary register.
```

luismarques (commented):
Some editing suggestions.

kito-cheng:
@luismarques applied, thanks :)

riscv-elf.adoc
```diff
@@ -1072,6 +1072,7 @@ non-standard ABI extensions.
 | Tag_RISCV_priv_spec                 |        8 | uleb128        | *Deprecated*, indicates the major version of the privileged specification.
 | Tag_RISCV_priv_spec_minor           |       10 | uleb128        | *Deprecated*, indicates the minor version of the privileged specification.
 | Tag_RISCV_priv_spec_revision        |       12 | uleb128        | *Deprecated*, indicates the revision version of the privileged specification.
+| Tag_RISCV_gp_reg_usage              |       16 | uleb128        | Indicates the usage definition of the register.
```
ilovepi, line 1075:
```suggestion
| Tag_RISCV_gp_reg_usage              |       16 | uleb128        | Indicates the usage definition of the X3 register.
```
Should we be specific here?

ilovepi (commented):
I've left some minor suggestions for phrasing, and a question regarding merging.

This is the solution we were hoping for in Fuchsia, so I'm glad to see this PR. Thanks for taking the time to do this @kito-cheng.

riscv-elf.adoc
```diff
@@ -1176,6 +1177,27 @@ Merge policy:::
...
+3:: This object uses `gp` as a temporary register.
+4~1023:: Reserved for future standard defined platform register.
+1024~2047:: Reserved for nonstandard defined platform register.
+
+--
+
+Merge policy:::
+The linker should issue errors when object files with differing `gp` usage are
+combined. However, an exception exists: the value `0` can merge with any
+non-zero value. After the merge, the resulting value will be the non-zero one.
```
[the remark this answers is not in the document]
kito-cheng: Good point, value 3 would be danger to merge to all others.

kito-cheng:
Changes:
- Rename gp into x3 to prevent misleading, e.g. using gp as shadow stack pointer seems weird.
- Update merge rule: explicitly allow merge 0 with 1 and 2 only.

MaskRay:
Thank you for making GP more useful when many folks don't think think re-purposing GP is an "ABI break" on their platforms.

I still hold the opinion from https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/371#issuecomment-1489593011 that a tag in the attribute section is not going to be super useful, more so if we emit it by default for `-mrelax`.
It will waste some relocatable object file sizes, and if we make `SHT_RISCV_ATTRIBUTE` allocable, some bytes in the linked image.

Right now I don't have a strong opinion whether software shadow stack based on GP should emit a tag by default.
If a user does that and enables `--relax-gp` (current default in GNU ld), they will get an error.
For ld.lld which doesn't enable global pointer relaxation by default, this is going to be more like busy work.
(There are many linker option combinations that produce broken output. We simply cannot diagnose every invalid case.)

sorear:
Why would `-mrelax` generate the attribute tag when `-mrelax` doesn't generate explicit uses of `gp`?  The `gp` references are not, with the exception of the single point where `gp` is initialized, generated until link time so the attribute tag shouldn't be created until then either.

asb:
To ask a slightly higher level question - are we sure we want to go down the route of picking out every potential ABI option and specifying it in attributes (which seems to be the route this and the atomics PR seems to be going down). I can believe it's better this way, but it would be reassuring to know if people spent time considering an alternative that e.g. just specified the platform ABI targeted by a given ELF (potentially with an ABI version).

palmer-dabbelt:
> To ask a slightly higher level question - are we sure we want to go down the route of picking out every potential ABI option and specifying it in attributes (which seems to be the route this and the atomics PR seems to be going down). I can believe it's better this way, but it would be reassuring to know if people spent time considering an alternative that e.g. just specified the platform ABI targeted by a given ELF (potentially with an ABI version).

That's kind of a broader issue, though: coupling together features into a set requires getting consensus on what those sets should be and there's just no mechanism for doing that in RISC-V land.  It's essentially the same problem that's lead to an explosion of extensions, just more visible because it's ABI and we're all used to ABI proliferation being a mess.

So I think we just need to suck it up and live with the mess, it's already defacto there from the extensions.  We can just punt to the distros to group features into sets and then go deprecate everything else in a few years when it's sufficiently broken.

kito-cheng:
This issue could break into two separate parts IMO: bare-metal and UNIX-like platforms (Linux, *BSD, Android...)

For bare-metal: The answer is yes; we are going to identify potential ABI options and add attributes to let users configure what they want; that's is the scope of EABI; this part will drive by IAR folks, and I believe this is not the issue we concern here, so I would not spend too much word here :)

For UNIX-like platforms: I would like to minimalize this *IF* possible; however, as you mentioned, we already have atomic and this PR:
- atomic is unavoidable due to some...historical mistakes.
- For x3/gp, I would say that's a historical mistake too; we didn't reserve a platform register at the beginning, and the x3/gp is the best candidate for changing. Anyway, the full story described in https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/371, so I won't repeat again here :)

Do we have more ABI options for UNIX-like platforms? It's hard to guarantee to say no more, but this will be carefully considered when introducing new ABI options.

But could we prevent those EABI options from being applied on UNIX-like platforms? It's hard to stop people from doing so; from the psABI aspect, we can add NOTE to mention this is EABI, and not recommended applied on UNIX-like platforms.

> but it would be reassuring to know if people spent time considering an alternative that e.g. just specified the platform ABI targeted by a given ELF (potentially with an ABI version).

That would be useful; we could create a separated issue to track that

kito-cheng:
@asb @MaskRay did you mind give explicitly LGTM if it's OK from the LLVM land. we are OK with that in GNU land, and then I'll gonna prepare PoC for this :)

kito-cheng:
Changes:

-Rebase and squash patches

kito-cheng:
binutils PoC: https://github.com/kito-cheng/riscv-binutils-gdb/commit/d13849c84696f533f3e48fb1c5b147e0e01e2f59

kito-cheng:
ping @asb @MaskRay, this will be merged if no objection this week.

MaskRay:
> That's kind of a broader issue, though: coupling together features into a set requires getting consensus on what those sets should be and there's just no mechanism for doing that in RISC-V land. It's essentially the same problem that's lead to an explosion of extensions, just more visible becaus

This is exactly my concern. In another thread (I don't recall which) I mentioned that compilers cannot determine whether inline assembly uses x3 and in what way.

I am sorry that I haven't replied earlier. I think I am ok if psABI adds `Tag_RISCV_x3_reg_usage`, as long as if compilers don't emit `Tag_RISCV_x3_reg_usage`, as that just doesn't seem useful...

If folks loving gp wants to an opt-in option to emit `Tag_RISCV_x3_reg_usage`, I think I'm fine as well, as a compromise...

palmer-dabbelt:
> > That's kind of a broader issue, though: coupling together features into a set requires getting consensus on what those sets should be and there's just no mechanism for doing that in RISC-V land. It's essentially the same problem that's lead to an explosion of extensions, just more visible becaus
> 
> This is exactly my concern. In another thread (I don't recall which) I mentioned that compilers cannot determine whether inline assembly uses x3 and in what way.

We can't really expect compilers to be able to tell if inline assembly violates any ABI constraint.  It's inline assembly anything goes.

> I am sorry that I haven't replied earlier. I think I am ok if psABI adds `Tag_RISCV_x3_reg_usage`, as long as if compilers don't emit `Tag_RISCV_x3_reg_usage`, as that just doesn't seem useful... and purely unneeded code size for many systems.
> 
> If folks loving gp wants to an opt-in option to emit `Tag_RISCV_x3_reg_usage`, I think I'm fine as well, as a compromise...

kito-cheng:
>> This is exactly my concern. In another thread (I don't recall which) I mentioned that compilers cannot determine whether inline assembly uses x3 and in what way.
>
> We can't really expect compilers to be able to tell if inline assembly violates any ABI constraint. It's inline assembly anything goes.

Yeah, I think user could write any ABI violation code as they want, so I don't really concern about that - user should handle those inline asm very carefully since most compiler didn't really know what inline asm write - we all rely on the info what user give.

And my thought on that tag other than the binutils patch, the binutils patch is just basic tag parsing and merge rule, which not cover all thing we should do:
- Linker should NOT perform gp relaxation if any input object file contain Tag_RISCV_x3_reg_usage other than 0 (unknown or unset) and 1 (gp relax).
- Linker should set Tag_RISCV_x3_reg_usage to 1 if gp relaxation has perform.
- If -mno-gp-relax is given (assembler or compiler), then Tag_RISCV_x3_reg_usage should explicitly set to 0 for the final output.
- Compiler should emit Tag_RISCV_x3_reg_usage as 2 if it use as shadow stack.

aswaterman:
> If folks loving gp

You mean folks who are using the ABI as specified?  We all know you dislike the ABI, but you should characterize your position a bit more forthrightly.

asb:
> ping @asb @MaskRay, this will be merged if no objection this week.

Now I'm back from vacation, just noting for the record I've got no objection to this flag (and sorry for missing the earlier ping!).

MaskRay:
> You mean folks who are using the ABI as specified? We all know you dislike the ABI, but you should characterize your position a bit more forthrightly.

I am forthright that I don't think global pointer relaxation provides sufficient valuable.
I am forthright that I don't get why marking object files with `Tag_RISCV_x3_reg_usage` is useful, and my arguments have been provided above.

I do worry that Clang produced relocatable object files will be annotated as `Tag_RISCV_x3_reg_usage` and lld will somehow need to parse it. I think this is busy work. If the toolchain will not be patched to add `Tag_RISCV_x3_reg_usage` in the default mode, I am happy to take a compromise and not object to this tag in the ABI.

kito-cheng:
@MaskRay I feel there's two things mixed in your comment, one concern for `Tag_RISCV_x3_reg_usage` and one concern for ELF attribute, so let us break it down into two parts to discuss.

Let me describe the ELF attribute first. I think we have discussed this in different PRs/issues, and I know you are not really convinced that we need that.

There are many different containers in ELF format which can carry information like e_flags, properties, ELF attributes, and many different .note.* sections.

And we pick ELF attributes at beginning due to following reason:
1. It’s implemented in existing toolchain
2. It’s allow store string and arbitrary length of integer (by ULEB128 format)

However I know it’s come with some drawback like:
1. It is not easy to do random access, it must parse from the beginning.
2. NO existing parser implementation within dynamic linker (like glibc)

So we may consider introducing ELF properties to make dynamic linker easier to read, however I think we could keep using ELF attributes, and define some guidelines to clarify how to decide what information/tag should add to attribute or property.

The guideline in my mind is:
- The tag should go property if it must read the information at dynamic linking time, otherwise use ELF attribute.
- If the information is only used by static linker and other static time tools like disassembler - then use ELF attribute.  

Back to `Tag_RISCV_x3_reg_usage`, that info mostly used by the static linker to determine the GP relaxation should be performed or not, does it introduce lot of complexity of the linker implementation, I expect that won’t be terrible high cost, we already have RVC flag to determine if the compressed instruction can be used within linker relaxation.

The first use case I expect is (soft) shadow stack, compiler will emit `Tag_RISCV_x3_reg_usage` when they begin use x3 as shadows stack pointer, then it will suppress linker to perform GP relaxation even -mno-gp-relax is not given.

Without that tag, users must be careful to make sure `-mno-gp-relax` is given when (soft) shadow stack enabled - otherwise the program might cause random bugs due to gp relaxation, but that could be prevented at the beginning if we have the tag.

For inline asm and hand craft assembly files…I tend to ignore those cases since I believe no matter what mechanism is  used, users could play tricks to violate or ignore that.

MaskRay:
> @MaskRay I feel there's two things mixed in your comment, one concern for Tag_RISCV_x3_reg_usage and one concern for ELF attribute, so let us break it down into two parts to discuss.

My two concerns are:

* whether Tag_RISCV_x3_reg_usage is useful
* in what cases users will bear the (small) overhead of Tag_RISCV_x3_reg_usage 

I do agree that if we ever need an ABI marker, build attributes is the best choice. Actually, no other alternative is ever close to be competitive.

I am more concerned with the second point.
What relocatable object files are going to have Tag_RISCV_x3_reg_usage?

I re-read the comments and think my
commenting "OK, if -mno-gp-relax is introduced, value 1 will now be distinguishable with others."
was mistaken.

> Linker should NOT perform gp relaxation if any input object file contain Tag_RISCV_x3_reg_usage other than 0 (unknown or unset) and 1 (gp relax).

OK, a bit of busy work in the linker to me, but I think this is acceptable.

> Linker should set Tag_RISCV_x3_reg_usage to 1 if gp relaxation has perform.

I hope that in the absence of build attributes, the linker doesn't need to set Tag_RISCV_x3_reg_usage.

> If -mno-gp-relax is given (assembler or compiler), then Tag_RISCV_x3_reg_usage should explicitly set to 0 for the final output.

0 is compatible with global pointer relaxation, then what does the 0 value do?

In the default mode (whether -mrelax or -mno-relax, but there is no -m[no-]gp-relax), is there a Tag_RISCV_x3_reg_usage attribute?

> Compiler should emit Tag_RISCV_x3_reg_usage as 2 if it use as shadow stack.

This is small overhead, but I think it's fine.

PR #389, opened 2023-06-16
Proposal for Vector Calling Convention

lhtin:
Update 2023-6-29: Updated to a new proposal based on comments, please view the file changes directly.

This proposal is based on the prior work (#171) which is currently implemented in LLVM. The differences between them are:

1. v1-v15 and v24-v31 is callee save in this proposal for vector functions, but LLVM's one is all caller-save
2. No ABI for tuple type ABI implemented in LLVM yet
3. For variadic functions, the current LLVM does inconsistent passing between caller and callee. (https://godbolt.org/z/9qv9n3zYP)

PoCs:
- Link of PoC on Clang/LLVM (https://reviews.llvm.org/D154576)
- Link of PoC on GCC (https://gcc.gnu.org/pipermail/gcc-patches/2023-July/625042.html)

------
NOTE: **Original post, it doesn't matter if you don't read it.**

Hi,

Here's a proposal for vector calling convention. This proposal is based on the prior work (#171) which is currently implemented in LLVM. This proposal extends the range of registers that can be used for vector data types and added more descriptions of the allocation.

Update 2023-6-27: Here (https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/38) is a proposal that is very similar to the current one.

Reasons for the proposal:

* Reason 1: The vector register size tends to be very large, so all vector registers used to pass arguments can reduce memory usage.

  In particular, for arguments with LMUL equal to 8, it is possible to support passing in 3 such arguments (i.e. v8\-v15, v16\-v23, v24\-v31).

* Reason 2: In RVV, the mask operand of an instruction with a mask must be placed in the v0 register. Let the first vector mask type argument pass in v0 can reduce register move.

* Reason 3: Finding the right register segment each time starting from v1 allows more arguments to be passed through the vector registers.
  
  For example for the function `void foo(vint32m8_t a1, vint32m8_t a2, vint32m8_t a3, vint32m1_t a4 vint32m2_t a5, vint32m4_t a6);`. You can pass a1 into v8\-v15, a2 into v16\-v23, a3 into v24\-v31, a4 into v1, a5 into v2\-v3, a6 into v4\-v7. If not, a4, a5, and a6 need to be passed by reference.

* ~~Reason 4: Start from v1 to find LMUL-aligned vector registers segment for return value can reduce the overlap of different LMUL return value vector registers.~~

  (Update 2023-6-28: Kito pointed out that this reason is not true. https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/389#issuecomment-1610549714)

  For example, for the following code snippet. You can return x in v1, and y in v2-v3, they do not overlap.

  ```
  vint32m1_t x = foo1 ();
  vint32m2_t y = foo2 (c, d);
  vint32m2_t z = __riscv_vwadd_wv (x, y, vl);
  ```

I have a few points that I am not very sure about and would like your help in looking at.
1. Do we really need all ilp32v, ilp32fv, ilp32dv?
   
   Update 2023-6-22: After further understanding, there should be no need to add a new abi, but the function that uses the vector calling convention must be annotated with `STO_RISCV_VARIANT_CC`.

4. Will using all registers as argument registers to cause the resolver routine in the dynamic link lazy bind cost too much?
   
   Update:
   
   Maybe solved by #190 ? (point by @sorear ). From bellow two patches, it seems that the STO_RISCV_VARIANT_CC symbol is not yet used inside glibc.
   - https://sourceware.org/git/?p=glibc.git;a=commit;h=117e8b341c5c0ace8d65feeef136fececb3fdc9c
   - https://sourceware.org/git/?p=glibc.git;a=commit;h=bf88b47ecb54888a789c02fa81aa4ab81ec2f3a5

Feel free to ask any questions or suggestions.

Best,
Lehua (RiVAI)


riscv-cc.adoc
```diff
@@ -329,6 +325,30 @@ type would be passed.
...
+scalar arguments of the same type pass like the named scalar arguments of
+vector data type. See the following description about named scalar arguments
+of vector data type.
+
+Named scalar arguments of vector data type has property LMUL. If LMUL is less
+than 1, treat it as 1. If it is possible to find unused registers segment
+starting from v1 and the first register in segment is LMUL-aligned, use these
+registers to pass the argument. Otherwise, pass through the function stack.
+
+Aggregates cannot pass with vector registers.
```
nick-knight, line 340: Why can't we just flatten aggregates, like in the other calling conventions?

And aren't the RVV tuple types essentially "aggregates"?

It's true that GCC and LLVM currently don't support scalable vectors in aggregates, but GCC has much of the infrastructure in place, and the LLVM folks who opposed this previously (w.r.t. SVE) aren't so active any more. So I'm optimistic we will see this as the RVV C/C++ dialect evolves. It'd be nice if our ABI didn't rule it out.

nick-knight (commented):
Some prior work: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/171

riscv-cc.adoc
```diff
@@ -329,6 +325,30 @@ type would be passed.
...
+
+Variadic vector arguments are passed by reference.
+
+Vector mask value is rturned in the same manner as a first named argument of the
+same type would be passed.
+
+Vector data value is returnd in the same manner as a first named argument of the
+same type would be passed.
+
+NOTE: vector mask type and data type defined https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/master/rvv-intrinsic-rfc.md#type-system[here]. Data type include tuple type.
```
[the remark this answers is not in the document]
lhtin: Yes, the type extended by the tuple type is also included. The proposal only stipulates that the argument register allocated for the vector type is LMUL-aligned, and the number of registers occupied by the vector type itself is not fixed.

lhtin:
> Some prior work: #171

Thank you for sending the previous discussion about vector calling convention.

From this issue and related issues(#66), I found that if all vector registers are used as argument registers, the resolver routine in the dynamic link lazy bind needs to save all vector registers. This work will be huge because the vector register itself can be very large. This issue needs to be further resolved.

riscv-cc.adoc
lhtin: Thanks for the reminder, I think it makes sense, even if there is currently no way to implement the inclusion of vector scalar type filed in struct. I will add these as soon as possible.

lhtin: Hi, I've added the description of the struct type pass rule, thanks.

sorear:
Suppose XLEN = VLEN = 64 and we are passing an argument of type `struct { int a; int b; vint8m1_t c; }`.  Since the struct has three fields, this proposal causes the struct to be "passed according to the integer calling convention".  But since the aggregate is no larger than 2*XLEN, the struct is passed in a pair of integer registers.  If the same call occurs with VLEN = 128, the struct will be passed by reference in the proposal.  I consider it undesirable for the registers used to pass an argument to depend on the runtime VLEN, so I propose that the integer calling convention be modified to add "aggregates containing a value of vector type" to the list of aggregates always passed by reference in the integer calling convention.

> From this issue and related issues(https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/66), I found that if all vector registers are used as argument registers, the resolver routine in the dynamic link lazy bind needs to save all vector registers. This work will be huge because the vector register itself can be very large. This issue needs to be further resolved.

Wasn't this already resolved in #190?

> Do we really need all ilp32v, ilp32fv, ilp32dv

I don't think we should be defining any new ABIs for this; the vector calling convention is exactly the same as the integer calling convention for functions that don't take or return vector values, supporting scalable vector arguments or return values without the V extension is meaningless, and any proposal which prevents mixing vector and non-vector code and requires to recompile the world at a single time is not viable for major distros.

lhtin:
> Suppose XLEN = VLEN = 64 and we are passing an argument of type `struct { int a; int b; vint8m1_t c; }`. Since the struct has three fields, this proposal causes the struct to be "passed according to the integer calling convention". But since the aggregate is no larger than 2*XLEN, the struct is passed in a pair of integer registers. If the same call occurs with VLEN = 128, the struct will be passed by reference in the proposal. I consider it undesirable for the registers used to pass an argument to depend on the runtime VLEN, so I propose that the integer calling convention be modified to add "aggregates containing a value of vector type" to the list of aggregates always passed by reference in the integer calling convention.

I think the way to pass parameters does not depend on runtime VLEN, this compiler compile is also not aware of this information. Here when VLEN=64 and 128, the inconsistency is because you specify vl64 and vl128 at compile time and use the fixed-vlmax option. If you don't specify it, the compiler can only assume that it is VLEN agnostic and doesn't know if it is less than or equal to 2*XLEN. For safety reasons, it must be passed through memory. I think this is a bit like the difference between `struct { long long a; long long b; }` on top of ilp32 and lp64, where on ilp32 it pass in memory but use two registers in lp64.

> Wasn't this already resolved in #190?

Yes! Thank you for pointing that out.

According to my superficial understanding, this means that if you use vector registers to pass parameters, you need to set the STO_RISCV_VARIANT_CC attribute. So I understand that the resolver function does not use the vector register if it finds this attribute or saves it before using it and restores it afterward. Is it right?

> I don't think we should be defining any new ABIs for this; the vector calling convention is exactly the same as the integer calling convention for functions that don't take or return vector values, supporting scalable vector arguments or return values without the V extension is meaningless, and any proposal which prevents mixing vector and non-vector code and requires to recompile the world at a single time is not viable for major distros.

How do you know if a compiled library uses a vector register as an argument without adding a new abi? If one is used and one is not, then linking them together will cause problems, right?

sorear:
> I think the way to pass parameters does not depend on runtime VLEN, this compiler compile is also not aware of this information. Here when VLEN=64 and 128, the inconsistency is because you specify vl64 and vl128 at compile time and use the fixed-vlmax option.

To be clear, I'm talking about the text of the proposal, not the current gcc behavior.  I believe that the proposal as written could be interpreted as saying that the compiler needs to look at `vlenb` before deciding which registers to pass the argument in.

> For safety reasons, it must be passed through memory.

This is the behavior I want - I propose we make it clear in the ABI specification.

> So I understand that the resolver function does not use the vector register if it finds this attribute or saves it before using it and restores it afterward. Is it right?
>
> Update: After looking at the relevant patches again, it should be forced to bind directly instead of lazy bind when the function comes with STO_RISCV_VARIANT_CC.

I would argue that both of those are allowed implementations.

> How do you know if a compiled library uses a vector register as an argument without adding a new abi? If one is used and one is not, then linking them together will cause problems, right?

The calling convention is per-function.  If the caller of a function passes a vector register and the callee expects a vector argument, it works.  If the caller is using the integer convention and the callee expects only integers, it works since both sides are using the integer convention.  If the caller and the callee have mismatched prototypes, we are already looking at undefined behavior.

lhtin:
@sorear @nick-knight For the question of whether new ABIs need to be added. After re-understood the existing ABI specification, functions need to be marked as [STO_RISCV_VARIANT_CC](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#dynamic-linking)  for cases that are not compatible with the current ABI specification, so that the compiler, inker, and loader can distinguish them and prevent caller and callee from using different specifications.

Therefore, it should not be necessary to add a new abi. However, the compiler needs to add an option to enable the vector calling convention and annotate the function that uses vector registers to pass arguments or return value with STO_RISCV_VARIANT_CC.

lhtin:
> This is the behavior I want - I propose we make it clear in the ABI specification.

I think this proposal is reasonable. Currently, neither GCC nor LLVM supports putting vector type fields in struct structures. Adding this rule should not introduce a breaking change.

riscv-cc.adoc
```diff
@@ -329,6 +329,36 @@ type would be passed.
...
+
+A struct containing just one vector type value is passed as though it were a standalone vector type value.
+
+A struct containing two vector type values is passed in vector registers, if there are suitable unused continuous vector register sets for them.
+
+A struct containing one vector type value and one integer (or bitfield), in either order, is passed in vector registers and an integer register, provided the integer is no more than XLEN bits wide, and at least one unused continuous vector register set and at least one integer argument register are available.
+
+A struct containing one vector type value and one floating-point real, in either order, is passed in vector registers and an floating-point register, provided the floating-point real is no more than ABI_FLEN bits wide, and at least one unused continuous vector register set and at least one floating-point argument register are available.
+
+A struct is not passed in the above manner, then it is passed according to the integer calling convention.
```
kito-cheng, line 352: Plz drop the struct part first, psABI require PoC, and it's not implementable on LLVM land.

riscv-cc.adoc
```diff
@@ -329,6 +329,36 @@ type would be passed.
 Floating-point registers fs0-fs11 shall be preserved across procedure calls,
 provided they hold values no more than ABI_FLEN bits wide.
 
+=== Hardware Vector Calling Convention
```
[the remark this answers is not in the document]
lhtin: Get it.

riscv-cc.adoc
```diff
@@ -104,18 +104,14 @@ duration in accordance with C11 section 7.6 "Floating-point environment
...
 | vl      |              | Vector length                | No
 | vtype   |              | Vector data type register    | No
 | vxrm    |              | Vector fixed-point rounding mode register    | No
 | vxsat   |              | Vector fixed-point saturation flag register  | No
 |===
 
-
-Vector registers are not used for passing arguments or return values; we
-intend to define a new calling convention variant to allow that as a future
-software optimization.
```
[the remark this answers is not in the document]
lhtin: Get it.

riscv-cc.adoc
```diff
@@ -166,6 +162,10 @@ the high-order XLEN bits are passed on the stack.
 Scalars wider than 2×XLEN bits are passed by reference and are replaced in the
 argument list with the address.
 
+The size of the vector type is considered to be unknown. So vector type scalar always is passed by reference.
```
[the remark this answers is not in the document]
lhtin: Sorry, the statement is a bit wrong, it should be `scalar of vector type`.

riscv-cc.adoc
```diff
@@ -104,18 +104,14 @@ duration in accordance with C11 section 7.6 "Floating-point environment
 |===
 | Name    | ABI Mnemonic | Meaning                      | Preserved across calls?
 
-| v0-v31  |              | Temporary registers          | No
+| v0      |              | First vector mask argument or mask return value | No
+| v1-v31  |              | Other vector mask and data argument registers   | No
```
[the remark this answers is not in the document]
lhtin: Get it.

riscv-cc.adoc
```diff
@@ -329,6 +329,36 @@ type would be passed.
...
+
+A struct is not passed in the above manner, then it is passed according to the integer calling convention.
+
+NOTE: See the Hardware Floating-point Calling Convention section for the definition of struct.
+
+Unions are never flattened and are always passed according to the integer calling convention.
+
+Values are returned in the same manner as the first named argument of the same type would be passed.
+
+If a function uses vector registers to pass arguments or return value, the function needs to be annotated with `STO_RISCV_VARIANT_CC`. See <<Dynamic Linking>> for more details.
```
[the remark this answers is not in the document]
lhtin: Get it.

riscv-cc.adoc
lhtin: Get it.

riscv-cc.adoc
```diff
@@ -329,6 +330,22 @@ type would be passed.
 Floating-point registers fs0-fs11 shall be preserved across procedure calls,
 provided they hold values no more than ABI_FLEN bits wide.
 
+=== Standard Vector Calling Convention Variant
```
sorear, line 333: The section title is contributing significantly to the confusion here.  We are NOT defining a new calling convention variant, we are making a compatible extension of the existing calling conventions to handle a new type.  We should call the section something like "Rules for vector arguments and return values in the integer calling convention and the hardware floating point calling convention" or remove the section entirely and edit the contents into the other sections.

riscv-cc.adoc
```diff
@@ -166,6 +163,10 @@ the high-order XLEN bits are passed on the stack.
 Scalars wider than 2×XLEN bits are passed by reference and are replaced in the
 argument list with the address.
 
+The size of scalars with vector type is considered to be unknown, so these scalars always is passed by reference.
+
```
sorear, line 167: Since we aren't defining any new ABIs, there is no calling convention variant where scalars of vector type are passed by reference - appearance of any scalar of vector type requires the use of the rules in the "Rules for vector arguments ..." section.

riscv-cc.adoc
```diff
@@ -174,7 +175,7 @@ of registers; if only one register is available, the first XLEN bits are passed
 in a register and the remaining bits are passed on the stack. If no registers are
 available, the aggregate is passed on the stack. Bits unused due to
 padding, and bits past the end of an aggregate whose size in bits is not
-divisible by XLEN, are undefined.
+divisible by XLEN, are undefined. If the aggregate contains a field of vector type, it always is passed by reference.
```
sorear, line 178: ditto

riscv-cc.adoc
```diff
@@ -329,6 +330,22 @@ type would be passed.
...
+
+The hardware vector calling convention adds 1 argument register for vector mask type argument and 31 argument registers for vector data and tuple type argument which are v0 and v1-v31, respectively. v0 is used for the first vector mask type argument and the vector mask type return value, the rest of the mask type arguments are treated as vector data type arguments. v1-v31 are also used for the vector data and tuple type return value.
+
+Vector data type arguments have properties LMUL and NREGS, the current LMUL can be 1/8, 1/4, 1/2, 1, 2, 4, 8, the current NREGS can be 1, 2, 4, 8. For arguments with LMUL less than 1, their LMUL is treated as 1. The LMUL of the vector mask type argument is treated as 1. The NREGS property means the number of registers needed for this argument. For vector data type, NREGS is 1 when LMUL is less than 1, otherwise NREGS is equal to LMUL. If it is possible to find NREGS unused continuous vector register set starting from v1 and its first register is LMUL-aligned, use these registers to pass the argument. Otherwise, the argument is passed by reference.
```
sorear, line 339: Do vector arguments go in the lowest numbered free vector register set satisfying the criteria, the lowest numbered register set after those already used, or some other rule?

```c
void foo(vint8m2_t x, vint8m2_t y) {
    /* x is v2 because ??? */
    /* y is v1 because it's the lowest not already used by x,
       v4 because it's the lowest after x, or ??? */
}
```

riscv-cc.adoc
```diff
@@ -329,6 +330,22 @@ type would be passed.
...
+
+vector tuple type arguments have the same LMUL and NREGS properties as the vector data type, but also have the NF property. NREGS equals NF multiplied by LMUL, but cannot exceed 8. The process of finding the argument registers is the same as for the vector data type.
+
+NOTE: The vector mask type, data type and tuple type are defined https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/master/rvv-intrinsic-rfc.md#type-system[here].
+
+Values are returned in the same manner as the first named argument of the same type would be passed.
+
+If a function uses standard vector calling convention variant, the function needs to be annotated with `STO_RISCV_VARIANT_CC`. See xref:riscv-elf.adoc#dynamic-linking[Dynamic Linking] for more details.
+
```
sorear, line 348: If a function has any arguments or return values passed using vector registers,

riscv-cc.adoc
```diff
@@ -329,6 +330,22 @@ type would be passed.
...
+Vector data type arguments have properties LMUL and NREGS, the current LMUL can be 1/8, 1/4, 1/2, 1, 2, 4, 8, the current NREGS can be 1, 2, 4, 8. For arguments with LMUL less than 1, their LMUL is treated as 1. The LMUL of the vector mask type argument is treated as 1. The NREGS property means the number of registers needed for this argument. For vector data type, NREGS is 1 when LMUL is less than 1, otherwise NREGS is equal to LMUL. If it is possible to find NREGS unused continuous vector register set starting from v1 and its first register is LMUL-aligned, use these registers to pass the argument. Otherwise, the argument is passed by reference.
+
+vector tuple type arguments have the same LMUL and NREGS properties as the vector data type, but also have the NF property. NREGS equals NF multiplied by LMUL, but cannot exceed 8. The process of finding the argument registers is the same as for the vector data type.
+
+NOTE: The vector mask type, data type and tuple type are defined https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/master/rvv-intrinsic-rfc.md#type-system[here].
+
```
sorear, line 344: Since aggregates containing vector types were removed from the initial proposal, we should explicitly say that there is no ABI for aggregates containing vector types and passing them is illegal, so that if they are added in the future it will not be a compatibility break.

riscv-cc.adoc
jrtc27: I mean, this is already defined by the existing part of the spec that introduces STO_RISCV_VARIANT_CC...

riscv-cc.adoc
lhtin: The process of allocation:
1. For x, LMUL=2, NREGS=2. It can find v2-v3 which are unused and LMUL-aligned from the start of v1, so allocate v2-v3 for it.
2. For y, LMUL=2, NREGS=2. It can find v4-v5  which are unused and LMUL-aligned from the start of v1, so allocate v4-v5 for it.

kito-cheng: We need more document the allocation detail on the spec, and I would suggest drop the NREGS term, just using LMUL and say any fractional LMUL will treat as LMUL=1.

kito-cheng:
> Reason 4: Start from v1 to find LMUL-aligned vector registers segment for return value can reduce the overlap of different LMUL return value vector registers.
> 
> For example, for the following code snippet. You can return x in v1, and y in v2-v3, they do not overlap.
>
> vint32m1_t x = foo1 ();
> vint32m2_t y = foo2 (c, d);
> vint32m2_t z = __riscv_vwadd_wv (x, y, vl);

This is not true, because all return value register are all call-clobber, so it not really get benefit on this case.

riscv-cc.adoc
lhtin: Hmm, that makes sense. This rule will never be used if we mandate that arguments and return values with vector type must follow the vector calling convention. My confusion is how to pass these arguments and return values before the vector calling convention becomes standard. Or just assume that it doesn't matter until the vector calling convention is determined.

riscv-cc.adoc
sorear: There was a typo, I meant y to be `vint8_t`.  I just noticed that the answer is in the top comment but it should be in the commits as well.

riscv-cc.adoc
sorear: I would say that it doesn't matter because until the vector calling convention is determined, users have no expectation of ABI compatibility between compilers and compiler versions.

lhtin:
> > Reason 4: Start from v1 to find LMUL-aligned vector registers segment for return value can reduce the overlap of different LMUL return value vector registers.
> > For example, for the following code snippet. You can return x in v1, and y in v2-v3, they do not overlap.
> > vint32m1_t x = foo1 ();
> > vint32m2_t y = foo2 (c, d);
> > vint32m2_t z = __riscv_vwadd_wv (x, y, vl);
> 
> This is not true, because all return value register are all call-clobber, so it not really get benefit on this case.

Indeed, thank you very much for pointing that out. I've added a delete line to Reason 4.

kito-cheng:
```
|===
| Name    | ABI Mnemonic | Meaning                      | Preserved across calls?

|v0       |              | Argument register for mask*  | No
|v1-v7    |              | Temporary registers          | No
|v8-v23   |              | Argument registers           | No
|v24-v31  |              | Temporary registers          | Yes
|===
```

I would like add callee-save registers rather than all caller-save, this could benefit those kernel loop contain with vector function, 16+1 argument register can hold 2 LMUL=8  value and one vector mask, which is satisfy almost all math function, the only function take 3 argument in std c math is `fma`, which can be expanded to instruction.

e.g.
```
void func (){
  vector_type a, b, x, y, z;
  for (i = 0 ~ n)
  {
     a = foo (b); // vectorized foo
     // `a` can be moved into a callee-save register instead of spill to stack.
     x = bar (y); // vectorized bar
     z = a + x;
  }
}
```

Just emphasize one point is the outer function `func` still define all vector register as caller-save register, so no callee-save overhead for `func`, the rule only applied on `foo` and `bar`, SVE also define similar rule for their vector calling  convention.

sorear:
What's the use case for temporary registers?  We have integer temporary registers because they're needed for some kinds of stub code, and I think float temporaries exist only to align the float ABI register names with the integer ABI register names, but with LMUL=8 types it's much easier to write a function that needs 24 vector registers than it is for integer registers.  Making v1 - v7 non-arguments wouldn't affect most functions but I'm also not sure what this is accomplishing.

Far more hesitant about a block of call-preserved registers.  I remember, years and years ago before there even was a vector spec, a decision being made to never add new call-preserved U-mode state, does anyone remember the complete list of reasons why that was?   `setjmp` shouldn't be an issue because setjmp doesn't take any vector arguments, so the compiler won't try to have vectors live across it.  `.eh_frame` could be a problem will also need careful handing; my first impression is that if vectors are not allowed to be live at landing pad entry, then vector-related entries don't need to appear in the unwind info, but this needs a closer analysis.

kito-cheng:
@sorear updated the table, I just changed the column of `Preserved across calls?` but no update with `Meaning` column.

> What's the use case for temporary registers? We have integer temporary registers because they're needed for some kinds of stub code, and I think float temporaries exist only to align the float ABI register names with the integer ABI register names, but with LMUL=8 types it's much easier to write a function that needs 24 vector registers than it is for integer registers. Making v1 - v7 non-arguments wouldn't affect most functions but I'm also not sure what this is accomplishing.

caller can hold value on v1-v7 and v24-v31, which is one LMUL=8, one LMUL=4, one LMUL=2 and one LMUL=1 values, this might not easy to use when LMUL=8, but would be useful to hold multiple value across the function call with LMUL=4~1.

Making v1-v7 and v24-v31 could reduce the potential spill/reload around the vectorized function call, use same example with more comment to demonstrate the idea.

All callee-save:
```
void func (){
  vector_type a, b, x, y, z;
  for (i = 0 ~ n)
  {
     b = load
     a = foo (b); // vectorized foo
     // spill a, b
     y = load
     x = bar (y); // vectorized bar
     // reload a, b
     z = a + x + b;
  }
}
```

So `a` and `b` will always spill/reload N times where N is the number of iteration.

All callee-save:
```
void func (){
  // NO callee-save reg store/restore since func is NOT using vector cc
  vector_type a, b, x, y, z;
  for (i = 0 ~ n)
  {
     b = load
     // Put b into vector arg register
     // Copy b to callee-save register
     a = foo (b); // vectorized foo
     // Move a to callee-save register
     y = load
     x = bar (y); // vectorized bar
     z = a + x + b;
  }
}
```

So there is no spill/reload around within the loop.

You might ask what about `foo` and `bar`? they won't have callee-save reg overhead IF the vector reg pressure is less than 17, but what if the reg pressure larger than 17? save/restore callee-save reg at prologue and epilogue, that would be same situation as all callee-save.

So in this design we can gain some performance IF those vectorized function has lower reg pressure, the worst case is same as the baseline design, overall is net gain.

> Far more hesitant about a block of call-preserved registers. I remember, years and years ago before there even was a vector spec, a decision being made to never add new call-preserved U-mode state, does anyone remember the complete list of reasons why that was? setjmp shouldn't be an issue because setjmp doesn't take any vector arguments, so the compiler won't try to have vectors live across it. 

Adding call-preserved vector registers into integer or HW floating point cc will cause ABI breakage, it will cause same problem as passing argument in vector reg, and this issue has addressed by `STO_RISCV_VARIANT_CC` :)

cc @aswaterman might have memory on those early things.

> .eh_frame could be a problem will also need careful handing; my first impression is that if vectors are not allowed to be live at landing pad entry, then vector-related entries don't need to appear in the unwind info, but this needs a closer analysis.

Fortunately RVV is not forerunner on scalable vector, unwinding stuffs already resolved by ARM folks and we just follow them (patch for unwinding scalable vector https://github.com/gcc-mirror/gcc/commit/89367e794613bdeb21df3e6fc0215f0acd553ef8) :P

sorear:
> So there is no spill/reload around within the loop.

With the changed proposal to have 17 call-clobbered argument registers, 15 call-saved, and 0 temporary (call-clobbered non-argument) registers I no longer have a question about the purpose of the last category :)

I also see that saving 16+12 registers in any vectorized function that calls non-vectorized functions was implemented and found adequate for SVE, so it should be good for us.

(Feedback from people with real-world experience with SVE calling conventions would be useful - I'm going off first principles here.  Did 16 saved vector registers turn out to be a useful number or should a different one be picked?)

> Fortunately RVV is not forerunner on scalable vector, unwinding stuffs already resolved by ARM folks and we just follow them (patch for unwinding scalable vector https://github.com/gcc-mirror/gcc/commit/89367e794613bdeb21df3e6fc0215f0acd553ef8) :P

I was asking more about the case where a vectorized function contains a try/catch block.  Since the exception unwinding could pass through any number of frames with saved vector registers, either the unwinder needs to be able to restore the saved vector registers, or the compiler needs to be aware that saved vector registers are lost on the edge between a throwing call and the landing pad.

Incidentally this does NOT work on aarch64, with the following test program, qemu-aarch64 8.0.0, and gcc 12.2.0 or 13.1.0:

```c++
#include <arm_sve.h>
volatile svint8_t *space;
typedef svint8_t (*Ft)(svint8_t);
volatile Ft g;
inline Ft hide(Ft f) { asm("" : "=r"(f) : "0"(f)); return f; }
svint8_t F3(svint8_t arg) {
    throw 5;
}
svint8_t F2(svint8_t arg) {
    svint8_t temp = *space;
    svint8_t ret = hide(F3)(arg);
    *space = temp;
    return ret;
}
svint8_t F1(svint8_t arg) {
    try {
        return hide(F2)(arg);
    } catch (int x) {
        return arg;
    }   
}
svint8_t F0(svint8_t arg) {
    hide(F1)(arg);
    return arg;
}
int main() {
    volatile svint8_t sp = svdup_s8(3);
    space = &sp;
    svint8_t rv = hide(F0)(svdup_s8(2));
    return svclastb_n_s8(svdup_b8(1),6,rv);
}
```

riscv-cc.adoc
lhtin: Good idea, thank you.

riscv-cc.adoc
lhtin: I think what you mean is that the vector calling convention is equivalent to an undefined part of an existing calling convention. For the vector type defined in the RVV Intrinsic document, it can be seamlessly added to the existing calling convention. However, this is not the case for GNU-ext vector types, which will be passed using the scalar register until there is a vector calling convention (i.e., currently), and then will need to switch to using the vector register (if the GNU-ext vector type is also included in the new calling convention). In this case calling `variant` is necessary.

sorear: If @kito-cheng's call-preserved vector register proposal is adopted then there's more of a case for describing the calling convention with call-preserved vector registers a variant.  I would still prefer to discuss and merge this PR independently of #390, since this one discusses strictly new types and doesn't require programmer-visible annotation while #390 is the opposite.

kito-cheng: @sorear I think so, I would like break whole vector calling convention into 3 parts, first one is this PR which contain all scalable vector stuffs pass by vector register, and second part is the fixed size gnu extension vector and last part is the  `riscv_rvv_vector_bits` (`arm_sve_vector_bits` or `--param riscv-autovec-preference=fixed-vlmax` mode).

riscv-cc.adoc
```diff
@@ -329,6 +328,24 @@ type would be passed.
 Floating-point registers fs0-fs11 shall be preserved across procedure calls,
 provided they hold values no more than ABI_FLEN bits wide.
 
+=== Standard Vector Calling Convention Variant
+
+This section applies only to named arguments. Variadic arguments of vector type are passed by reference.
+
+The hardware vector calling convention adds 1 argument register for argument of vector mask type and 31 argument registers for arguments of vector data type and vector tuple type which are v0 and v1-v31, respectively. v0 is used for the first argument of vector mask type and return values of vector mask type, the rest of the arguments of mask type are treated as arguments of vector data type. v1-v31 are also used for return values of vector data type and vector tuple type.
```
kito-cheng, line 335: ` the rest of the arguments of mask type are treated as arguments of vector data type`  -> ` the rest of the arguments of mask type are treated as arguments of vector data type with LMUL 1`

lhtin: Ok. In addition, is it necessary to explain LMUL and NF in this proposal, or is it assumed that the reader knows the meaning of LMUL and NF?

lhtin:
> ```
> |===
> | Name    | ABI Mnemonic | Meaning                      | Preserved across calls?
> 
> |v0       |              | Argument register for mask*  | No
> |v1-v7    |              | Callee-saved registers          | Yes
> |v8-v23   |              | Argument registers           | No
> |v24-v31  |              | Callee-saved registers         | Yes
> |===
> ```
> 
> I would like add callee-save registers rather than all caller-save, this could benefit those kernel loop contain with vector function, 16+1 argument register can hold 2 LMUL=8 value and one vector mask, which is satisfy almost all math function, the only function take 3 argument in std c math is `fma`, which can be expanded to instruction.
> 
> e.g.
> 
> ```
> void func (){
>   vector_type a, b, x, y, z;
>   for (i = 0 ~ n)
>   {
>      a = foo (b); // vectorized foo
>      // `a` can be moved into a callee-save register instead of spill to stack.
>      x = bar (y); // vectorized bar
>      z = a + x;
>   }
> }
> ```
> 
> Just emphasize one point is the outer function `func` still define all vector register as caller-save register, so no callee-save overhead for `func`, the rule only applied on `foo` and `bar`, SVE also define similar rule for their vector calling convention.

This proposal is more sophisticated and reasonable. I think I should update based on this and wait for more people to review it.

Also passing arguments of vector type to the unprototype function should report an error, similar to [ARM SVE's ABI](https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst#102argument-passing-conventions). Is this rule to be stated in this document as well?

riscv-cc.adoc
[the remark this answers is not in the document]
lhtin: > Vector registers are not used for passing arguments or return values; use Standard Vector Calling Convention Variant if want to passing arguments or return values in vector register.

A small question, for functions that contain vector type parameters or return values, is it necessary to follow the Standard Vector Calling Convention Variant? If so, this means that vector type arguments must be passed through vector registers or memory (if there are not enough registers). I think so, so this sentence is not needed.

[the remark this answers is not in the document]
kito-cheng: Yeah, I think it could be removed.

riscv-cc.adoc
kito-cheng: Good point, let add some comment to explain that :)

kito-cheng:
@sorear 

> With the changed proposal to have 17 call-clobbered argument registers, 15 call-saved, and 0 temporary (call-clobbered non-argument) registers I no longer have a question about the purpose of the last category :)

Cool, I'm glad we have a consensus here!

> (Feedback from people with real-world experience with SVE calling conventions would be useful - I'm going off first principles here. Did 16 saved vector registers turn out to be a useful number or should a different one be picked?)

Yeah, let me tag few more people once @lhtin updated :)


> I was asking more about the case where a vectorized function contains a try/catch block. Since the exception unwinding could pass through any number of frames with saved vector registers, either the unwinder needs to be able to restore the saved vector registers, or the compiler needs to be aware that saved vector registers are lost on the edge between a throwing call and the landing pad.
>
> Incidentally this does NOT work on aarch64, with the following test program, qemu-aarch64 8.0.0, and gcc 12.2.0 or 13.1.0:

Thanks for providing the case, I can reproduce with linaro aarch64 toolchain, we (SiFive folks) resolved several unwinding issue, so I thought should be not issues around their, but seems like not :(

kito-cheng:
@lhtin 

> This proposal is more sophisticated and reasonable. I think I should update based on this and wait for more people to review it.

Thanks!

> Also passing arguments of vector type to the unprototype function should report an error, similar to [ARM SVE's ABI](https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst#102argument-passing-conventions). Is this rule to be stated in this document as well?

Yeah, forbid that would be easier, so let us follow SVE here :)

riscv-cc.adoc
lhtin: @sorear @kito-cheng  I've re-described the register allocation rules, would it be easier to understand them?

riscv-cc.adoc
lhtin: > I mean, this is already defined by the existing part of the spec that introduces STO_RISCV_VARIANT_CC...

Removed this statement, thank you.

riscv-cc.adoc
sorear: With the new note it's clear to me, will allow @kito-cheng to resolve.

riscv-cc.adoc
lhtin: After further discussion, we currently prohibit the use of scalable vector type in struct or union in the proposal first, because LLVM has not proven whether it can be implemented. If this is possible in the future, we will add this piece. This approach does not introduce incompatibilities.

riscv-cc.adoc
```diff
@@ -329,6 +331,34 @@ type would be passed.
...
+
+If a function takes at least one argument in scalable vector registers, or if it returns results in such registers, it must ensure that the entire contents of v1-v7 and v24-v31 are preserved across the call. In other cases it no need to preserve.
+
+Each scalable data and tuple type has an LMUL attribute that indicates a vector register group. The value of LMUL indicates the number of vector registers in the vector register group and requires the first vector register number in the vector register group must be a multiple of it. For example, the LMUL of `vint64m8_t` is 8, so v8-v15 vector register group can be allocated to this type, but v9-v16 can not because the v9 register number is not a multiple of 8. If LMUL is less than 1, it is treated as 1. If it is a scalable mask type, its LMUL is 1.
+
+Each scalable tuple type also has a NFIELDS attribute, which indicates how many vector register groups the type contains. Thus a scalable tuple type takes up LMUL×NFIELDS registers.
+
+The rules for passing the scalable vector argument are as follows:
+
+1. For the first scalable mask parameter, use v0 to pass it. The argument has now been allocated.
```
kito-cheng, line 350: use v0 to pass it if that has not yet been allocated

riscv-cc.adoc
```diff
@@ -329,6 +331,34 @@ type would be passed.
...
+
+2. For scalable data argument or rest scalable mask argument, if a vector register group can be found from v8 to v23 that has not yet been allocated and the first register number is a multiple of LMUL, then allocate the vector register group to this argument and mark these registers as allocated. Otherwise, pass it by reference. The argument has now been allocated.
+
+3. For scalable tuple argument, if NFIELDS vector register groups can be found from v8 to v23 that have not yet been allocated, and the first register number is a multiple of LMUL, then allocate these vector register groups to this argument and mark these registers as allocated. Otherwise, pass it by reference. The argument has now been allocated.
```
kito-cheng, line 354: I would like to swap 2 and 3, and saying scalable tuple type will break into (or treat as) NFIELDS-pieces scalable data argument.

This will introduce some complicated rule like partial tuple type passed in register and partial passed in memory, however my assumption is the vector register size might be large, the cost of memory traffic would be high, so introduce this can reduce that in some case, and utilize the vector arg. register.

e.g. 

`void foo (vint32m8_t a, vint32m8x2_t b)` a will passed in v8-v15 and b will passed in v16-v23 *AND* memory

riscv-cc.adoc
nick-knight: I understand the current deficiency of LLVM and I understand the task group policy, so I accept this decision. But I disagree with the comments about in/compatibilities: supporting aggregates containing vectors would require (yet) another calling convention variant.

lhtin: What I'm trying to say is that when we support this part, there will be no problems linking to the previous library.

riscv-cc.adoc
lhtin: Is this condition `has not yet been allocated` redundant? Since v0 can't be in the allocated state before it.

riscv-cc.adoc
sorear: My understanding is that the primary purpose of tuple types is to support the vector segment load/store instructions.  If that is correct, there's very little use for such a type split between registers and memory; does LLVM even support partially spilling a vector tuple type?  (Trying to pass one to inline asm crashes compiler explorer)

lhtin: For cases where there are enough registers, such as the function `void foo (vint32m1_t a, vint32m2_t b, vint32m1x2_t c)`. According to your rules, v8 is allocated to `a`, v10-v11 to `b`, and v9 and v12 to `c`. When using the `c` argument in segment store operation, it is necessary to move v9 and v12 to consecutive registers (e.g. move v9 to v13 and v12 to v14). Thus unnecessary instructions are added compared to the old rule.

kito-cheng: @sorear 
> My understanding is that the primary purpose of tuple types is to support the vector segment load/store instructions. If that is correct, there's very little use for such a type split between registers and memory;

Yeah, and also @lhtin's case convince me that's unnecessary complexity. 

> does LLVM even support partially spilling a vector tuple type? (Trying to pass one to inline asm crashes compiler explorer)

Per me LLVM experience, that should be implementable since it's modeled as struct, but anyway let drop my the latest proposal.

@lhtin 
> For cases where there are enough registers, such as the function void foo (vint32m1_t a, vint32m2_t b, vint32m1x2_t c). According to your rules, v8 is allocated to a, v10-v11 to b, and v9 and v12 to c. When using the c argument in segment store operation, it is necessary to move v9 and v12 to consecutive registers (e.g. move v9 to v13 and v12 to v14). Thus unnecessary instructions are added compared to the old rule.

Thanks for the case! the case convince me that my proposal here is bad idea :)

riscv-cc.adoc
```diff
@@ -104,17 +104,17 @@ duration in accordance with C11 section 7.6 "Floating-point environment
 |===
 | Name    | ABI Mnemonic | Meaning                      | Preserved across calls?
 
-| v0-v31  |              | Temporary registers          | No
+| v0      |              | Argument register            | No
+| v1-v7   |              | Callee-saved registers       | Yes*
+| v8-v23  |              | Argument registers           | No
+| v24-v31 |              | Callee-saved registers       | Yes*
```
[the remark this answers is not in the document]
lhtin: It does not feel easy to understand if two convention tables are in two separate places, because the two tables are for different functions. I think it would be better to just put the two tables in this place, and then state which table applies to functions with scalable vector arguments and return values, and which table applies to the other functions. like bellow:

```adoc
.Vector register convention for functions without scalable vector arguments and return values
[%autowidth]
|===
| Name    | ABI Mnemonic | Meaning                      | Preserved across calls?

| v0-v31  |              | Temporary registers          | No
| vl      |              | Vector length                | No
| vtype   |              | Vector data type register    | No
| vxrm    |              | Vector fixed-point rounding mode register    | No
| vxsat   |              | Vector fixed-point saturation flag register  | No
|===

.Vector register convention for functions with scalable vector arguments or return values
[%autowidth]
|===
| Name    | ABI Mnemonic | Meaning                      | Preserved across calls?

| v0      |              | Argument register            | No
| v1-v7   |              | Callee-saved registers       | Yes*
| v8-v23  |              | Argument registers           | No
| v24-v31 |              | Callee-saved registers       | Yes*
| vl      |              | Vector length                | No
| vtype   |              | Vector data type register    | No
| vxrm    |              | Vector fixed-point rounding mode register    | No
| vxsat   |              | Vector fixed-point saturation flag register  | No
|===

*: Only functions using vector registers to pass arguments or return values need to preserve these registers, other functions treat these registers as temporary registers.

Please refer to the <<Standard Vector Calling Convention Variant>> section for more details about functions with scalable vector arguments or return values.
```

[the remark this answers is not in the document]
kito-cheng: > .Vector register convention for functions with scalable vector arguments or return values

This statement seems not included fixed size vector, I would like use more neutral term here: 

Vector register convention for standard vector calling convention variant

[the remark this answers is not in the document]
lhtin: > Vector register convention for standard vector calling convention variant

Maybe `Vector register convention for functions that do not use the vector registers to pass arguments and return values` and `Vector register convention for functions that use the vector registers to pass arguments and return values` are more accurate?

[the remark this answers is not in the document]
kito-cheng: SVE has a case is you can annotate a function with aarch64_vector_pcs[1], but no vector argument/return value, I suspect that might be useful in some scenario, but explicitly say function arg or return passed in vector reg will  forbid such scenario, that's why I would like refer `standard vector calling convention variant` instead of referring pass/return in vector regs.

[1] https://github.com/gcc-mirror/gcc/blob/master/gcc/testsuite/gcc.target/aarch64/sve/pcs/saves_2_le_wrap.c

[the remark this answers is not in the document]
lhtin: Soga, make sense to me. A function with `#pragma omp declare simd` notate might be converted to a new function that uses the vector registers to pass arguments or return values. Therefore, these functions need to be explicitly marked when needed to indicate compliance with `the standard vector calling convention variant`.

riscv-cc.adoc
```diff
@@ -128,8 +140,7 @@ Any procedure that does explicitly write `vstart` to a nonzero value must zero
 
 == Procedure Calling Convention
 
-This chapter defines standard calling conventions, and describes how to pass
-parameters and return values.
+This chapter defines standard calling conventions and standard calling convention variants and describes how to pass arguments and return values.
```
kito-cheng, line 143:
```suggestion
This chapter defines standard calling conventions and standard calling convention variants,
and describes how to pass arguments and return values.
```

riscv-cc.adoc
```diff
@@ -111,10 +111,22 @@ duration in accordance with C11 section 7.6 "Floating-point environment
...
+| vl      |              | Vector length                | No
+| vtype   |              | Vector data type register    | No
+| vxrm    |              | Vector fixed-point rounding mode register    | No
+| vxsat   |              | Vector fixed-point saturation flag register  | No
+|===
 
-Vector registers are not used for passing arguments or return values; we
-intend to define a new calling convention variant to allow that as a future
-software optimization.
+Please refer to the <<Standard Vector Calling Convention Variant>> section for more details about standard vector calling convention variant.
```
kito-cheng, line 129:
```suggestion
Please refer to the <<Standard Vector Calling Convention Variant>> section for 
more details about standard vector calling convention variant.
```

riscv-cc.adoc
```diff
@@ -329,6 +340,36 @@ type would be passed.
 Floating-point registers fs0-fs11 shall be preserved across procedure calls,
 provided they hold values no more than ABI_FLEN bits wide.
 
+=== Standard Vector Calling Convention Variant
+
+The https://github.com/riscv/riscv-v-spec[RISC-V V Vector Extension] defines a set of thirty-two scalable vector registers, v0-v31. The https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/master/rvv-intrinsic-rfc.md#type-system[RISC-V V Vector Extension Intrinsics] defines scalable vector types which include scalable mask types, scalable data types, and scalable tuple types. A scalable vector value can be stored in vector register groups.
```
kito-cheng, line 345: I'm a bit hesitant to use the term "scalable vector" and "scalable" in the psABI spec since neither v-spec or rvv-intrinsic-doc didn't use that, so I would like to just "vector" rather than "scalable vector" in this section.

And `vector mask type` `vector data types` for consistent with https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/380/files

riscv-cc.adoc
```diff
@@ -329,6 +340,36 @@ type would be passed.
...
+
+The remainder of this section applies only to named scalable vector arguments, other named arguments and return values follow the standard calling convention. Variadic scalable vector arguments are passed by reference.
+
+For now, scalable vector types are not allowed in struct or union.
```
kito-cheng, line 349:
```suggestion
Vector types are disallowed in struct or union.
```

riscv-cc.adoc
```diff
@@ -329,6 +340,36 @@ type would be passed.
...
+
+v0 is used to pass the first scalable mask argument to a function, and to return scalable mask result from a function. v8-v23 are used to pass scalable data arguments, scalable tuple arguments and the rest scalable mask arguments to a function, and to return scalable data and scalable tuple results from a function.
+
+If a function takes at least one argument in scalable vector registers, or if it returns results in such registers, it must ensure that the entire contents of v1-v7 and v24-v31 are preserved across the call. In other cases it no need to preserve.
```
kito-cheng, line 353: Per discussion https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/389#discussion_r1251421684, I suggest change to:
```suggestion
It must ensure that the entire contents of v1-v7 and v24-v31 are preserved across the call.
```

riscv-cc.adoc
lhtin: I think the name `scalable` can use distinguishes it from the gnu-ext vector type. vector types include those defined by intrinsic doc, as well as gnu-ext vector types. But I think it's okay if there's one place where it's clear what it's referring to.

kito-cheng:
@lhtin Could you add few more info on the post?
- Link of PoC on clang/LLVM (https://reviews.llvm.org/D154576)
- Link of PoC on GCC (Add once you have updated one)
- Comparison with existing experimental vector calling convention on LLVM:
  - v1-v15 and v24-v31 is callee save in this proposal, but LLVM's one is all caller-save
  - No ABI for tuple type ABI implemented in LLVM yet.

kito-cheng:
cc @frasercrmck @rjiejie @aswaterman @rdolbeau @rofirrim @topperc @JeffreyALaw @palmer-dabbelt @preames

lhtin:
> @lhtin Could you add few more info on the post?
> 
> * Link of PoC on clang/LLVM (https://reviews.llvm.org/D154576)
> * Link of PoC on GCC (Add once you have updated one)
> * Comparison with existing experimental vector calling convention on LLVM:
>   
>   * v1-v15 and v24-v31 is callee save in this proposal, but LLVM's one is all caller-save
>   * No ABI for tuple type ABI implemented in LLVM yet.

Done, thank you for your guide.

aswaterman:
As a general statement, I remain deeply skeptical of the value of callee-saved vector registers. Given that they complicate the ABI, the default assumption should continue to be that all are caller-saved, and the onus should be on the folks proposing the partially-callee-saved ABI to demonstrate on existing designs that theirs has superior performance in practice.

aswaterman:
Hence the "demonstrate" in my remark.  I don't disagree with the premise that callee-saved registers might improve performance; I disagree with the notion of including them without proving they help in practice.

jnk0le:
> If the A and C loops vectorize and B has a call, then we may well end up wanting to carry a value from the A loop into the C loop over the call.

>The other cases would be things like loop fission/splitting were we could have 3 chunks of code after fission where the first and last chunks are vectorizable loops with a non-vectorizable chunk of code with a call in between. We might want to hold a value computed by the first vectorizable loop in a register over the call to then be used in the second vector loop.

IPRA allows to do this under current ABI (all caller saved), provided that the B calees are not clobbering all vector registers or doing syscalls.

kito-cheng:
@aswaterman We are working on building prototype and seeking benchmarks for that :)

riscv-cc.adoc
```diff
@@ -329,6 +347,76 @@ type would be passed.
...
+been allocated.
+
+2. For vector data arguments or rest vector mask arguments, starting from the
+v8 register, if a vector register group between v8-v23 that has not been
+allocated can be found and the first register number is a multiple of LMUL,
+then allocate this vector register group to the argument and mark these
+registers as allocated. Otherwise, pass it by reference. The argument has now
+been allocated.
+
+3. For vector tuple arguments, starting from the v8 register, if NFIELDS
```
4vtomat, line 395: One question on tuple type arguments. If the `NFIELDS=2` and we only find 1 register group for it, can we pass 1 group by register and the other by reference?
Or all of the group should be passed in the same way?

lhtin: The current proposal chooses the second way that either both go through registers or both go through memory.

4vtomat: Is it better to use first approach for better register usage?

lhtin: This approach needs a bit of refinement before we discuss it. For example, consider a case like the function `void foo (.... , vint32m4x2_t a, vint32m2_t b, vint32m2_t c);`, suppose the first approach is used and the remaining v18-v23 registers are available for the `a`, `b`, and `c` parameters. Then part of `a` is allocated to v20-v23 (4-aligned) and the other part is passed through memory, `b` is allocated to v18-v19, and `c` is passed through memory by reference. How is the other part of a going to be passed in this case? If something like 2xXLEN's scalar argument is passed through the stack, then the part passed through v20-v23 needs to be restored to the pretend_args area, which doesn't feel like an added benefit. Also, part of `a` is passed via stack and `b` is in registers, a situation I'm not sure can be implemented in the compiler, I feel like it can only be passed by reference and not by stack. But if it is passed by reference, it feels strange that this reference only references part of `a` argument.

4vtomat: Got it, after discussing with @kito-cheng , we thought that we can implement the original(non-partial) approach. Then we can find another possibility for increasing the register usage, thanks!

4vtomat: One other question, are the tuple type's `NFIELD` of register groups necessary to be consecutive?
I think we just need ensure all the groups are passed in a same way right?

lhtin: Please see if this previous discussion(https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/389#discussion_r1246717449) answers your question, thanks.

4vtomat: I see, it should better be consecutive since the instructions such as segment load/store need them to be, thanks!

riscv-cc.adoc
```diff
@@ -329,6 +347,76 @@ type would be passed.
...
+consecutive vector register groups between v8-v23 that have not been allocated
+can be found and the first register number is a multiple of LMUL, then allocate
+these vector register groups to the argument and mark these registers as
+allocated. Otherwise, pass it by reference. The argument has now been allocated.
```
4vtomat, line 399: Could you please add a note for example for `tuple type registers should be consecutive` constraint?
e.g. void foo(vint32m1_t a0, vint32m2_t a1, vint32m1x2_t a2)
a0 -> v8
a1 -> v10-v11
a2 -> v12-v13
but not 
a0 -> v8
a1 -> v10-v11
a2.0 -> v9
a2.1 -> v12
where `a2.0` is the first NFIELD of `a2` and `a2.1` is the second NFIELD of `a2`.

lhtin: Sure, how about this NOTE?:
```
NOTE: The registers assigned to the tuple vector data argument must be consecutive.
For example, for the function `void foo(vint32m1_t a, vint32m2_t b, vint32m1x2_t c)`,
v8 will be allocated to `a`, v10-v11 will be allocated to `b`,
v12-v13 instead of v9 and v12 will be allocated to `c`.
```

riscv-cc.adoc
```diff
@@ -329,6 +347,76 @@ type would be passed.
...
+Each vector tuple type also has an NFIELDS attribute that indicates how many
+vector register groups the type contains. Thus a vector tuple type needs to
+take up LMUL×NFIELDS registers.
+
+The rules for passing vector arguments are as follows:
+
+1. For the first vector mask argument, use v0 to pass it. The argument has now
+been allocated.
+
+2. For vector data arguments or rest vector mask arguments, starting from the
```
4vtomat, line 388: Is it necessary to allocate the register in order? For example,
`func foo(vint32m8 a, vint32m1 b, vint32m8 c)`,
if we allocate in order, we can just allocate `a` to `v8-v15`, `b` to `v16` and `c` by reference,
if we would like to maximize register usage, we have to allocate `a` to `v8-v15`, `c` to `v16-v23` and `b` by reference.

riscv-cc.adoc
4vtomat: Great! Thanks~

riscv-cc.adoc
lhtin: I had thought about this before, and I thought it would make the allocation rules very complicated (the allocation algorithm needs to choose the optimal allocation results while maintaining certainty), but I didn't see much benefit. Even these current rules are actually a little complicated, the actual user will write the interface when the function of such parameters? More discussion is welcome about it.

4vtomat: If we didn't sort the arguments, it's potentially violation for tuple type consecutive constraint as the example shown in [line 401](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/389/files#diff-3a934f00cffdb3e509722753126a2cf6082a7648ab3b9ca8cbb0e84f8a6a12edR401), since `c` is split into two `vint32m1_t` in backend, the normal ordered allocation would assign first `vint32m1_t` to `v9` and the second one to `v12`.

lhtin: > since c is split into two vint32m1_t in backend
Which backend? We treat `c` as a whole argument(not split), just like a `vint32m2_t` type, but without needing align 2.

4vtomat: For example, `void foo(vint32m1x2_t a) {}` is lowered to
`define void @foo(<vscale x 2 x i32> %a.coerce0, <vscale x 2 x i32> %a.coerce1) {
entry:
  ret void
}`
in LLVM.

lhtin: So LLVM can't allocate him as a whole argument? If this is the case, this proposal is now unimplementable for LLVM. The current specification allows a certain amount of ORDER, but not full ORDER. you can see an example of the second NOTE.

jrtc27: Without attempting to read all the details here:

LLVM can be made to implement whatever you want it to do (though some things are easier than others). Talking about whatever IR Clang happens to emit today and how the backend deals with that today isn't particularly meaningful.

4vtomat: @kito-cheng  what do you think? Do you think we should implement tuple type as a single type instead of splitting to multiple single type in LLVM IR?

4vtomat:
Since vector load/store has much more overhead than scalar one, I think it's worth it to design a rule for it.
Here is a strait forward recursive approach:
```
**Resources**:16 argument registers[v8, v9, v10, v11, v12, v13, v14, v15, v16, v17, v18, v19, v20, v21, v22, v23]
**Constraints**:
1. Arguments should be assigned in order, i.e. arg0 -> arg1 -> arg2 
2. Register starting number should be multiple of LMUL, e.g. vint32m4 can only be assigned to v8m4, v12m4, v16m4, v20m4.
3. Tuple type should be assigned to consecutive registers, e.g. vint32m1x2 can be assigned to (v8, v9), (v9, v10) or (v21, v22) etc. but not (v8, v10).
4. Register(s) that is first matched that is available should be assigned.

**Goal**: For each argument started from arg0, determine if it's passed by register or passed by reference that maximizes the total number of registers assigned.

**Approach**:
int NUM_RVV_REGS = 16;
bool CAN_ALLOCATE(X) {
    if (any register(s) can be found that meets the constraints2-4)
        return TRUE;
    return FALSE;
}

int NEED(X) {
    return registers needed for argument X;
}

int ALLOCATE(N) {
    if (N == ARG.size() || N == NUM_RVV_REGS)
        return 0;

    int dont_assign = ALLOCATE(N+1);
    int assign = -1;
    if (CAN_ALLOCATE(ARG[N]))
        assign = NEED(ARG[N]) + ALLOCATE(N+1);

    if (assign >= dont_assign) {
        mark ARG[N] as byreg;
        return assign;
    }
    mark ARG[N] as byref;
    return dont_assign;
}

Note: The additional N == NUM_RVV_REGS constraint in ALLOCATE(N) is to limit the recursion times to NUM_RVV_REGS to avoid blowing up.
```

lhtin:
> Since vector load/store has much more overhead than scalar one, I think it's worth it to design a rule for it. Here is a strait forward recursive approach:
> 
> Resources: 16 argument registers[v8, v9, v10, v11, v12, v13, v14, v15, v16, v17, v18, v19, v20, v21, v22, v23]
> Constraints:
> 1. Arguments should be assigned in order, i.e. arg0 -> arg1 -> arg2...
> 2. Register starting number should be multiple of LMUL, e.g. vint32m4 can only be assigned to v8m4, v12m4, v16m4, v20m4.
> 3. Tuple type should be assigned to consecutive registers, e.g. vint32m1x2 can be assigned to (v8, v9), (v9, v10) or (v21, v22) etc. but not (v8, v10).
> 4. Register(s) that is first matched that is available should be assigned.
> 
> Goal: For each argument started from arg0, determine if it's passed by register or passed by reference that maximizes the total number of registers assigned.
> 
> Approach:
> int NUM_RVV_REGS = 16;
> bool CAN_ALLOCATE(X) {
>     if (any register(s) can be found that meets the constraints2-4)
>         return TRUE;
>     return FALSE;
> }
> 
> int NEED(X) {
>     return registers needed for argument X;
> }
> 
> int ALLOCATE(N) {
>     if (N == ARG.size() || N == NUM_RVV_REGS)
>         return 0;
> 
>     int dont_assign = ALLOCATE(N+1);
>     int assign = -1;
>     if (CAN_ALLOCATE(ARG[N]))
>         assign = NEED(ARG[N]) + ALLOCATE(N+1);
> 
>     if (assign >= dont_assign) {
>         mark ARG[N] as byreg;
>         return assign;
>     }
>     mark ARG[N] as byref;
>     return dont_assign;
> }
> 
> Note: The additional N == NUM_RVV_REGS constraint in ALLOCATE(N) is to limit the recursion times to NUM_RVV_REGS to avoid blowing up.

Before I get into the details, I'd like to know what the difference is between your proposed allocation rules and the current ones?

4vtomat:
> > Since vector load/store has much more overhead than scalar one, I think it's worth it to design a rule for it. Here is a strait forward recursive approach:
> > Resources: 16 argument registers[v8, v9, v10, v11, v12, v13, v14, v15, v16, v17, v18, v19, v20, v21, v22, v23]
> > Constraints:
> > 
> > 1. Arguments should be assigned in order, i.e. arg0 -> arg1 -> arg2...
> > 2. Register starting number should be multiple of LMUL, e.g. vint32m4 can only be assigned to v8m4, v12m4, v16m4, v20m4.
> > 3. Tuple type should be assigned to consecutive registers, e.g. vint32m1x2 can be assigned to (v8, v9), (v9, v10) or (v21, v22) etc. but not (v8, v10).
> > 4. Register(s) that is first matched that is available should be assigned.
> > 
> > Goal: For each argument started from arg0, determine if it's passed by register or passed by reference that maximizes the total number of registers assigned.
> > Approach:
> > int NUM_RVV_REGS = 16;
> > bool CAN_ALLOCATE(X) {
> > if (any register(s) can be found that meets the constraints2-4)
> > return TRUE;
> > return FALSE;
> > }
> > int NEED(X) {
> > return registers needed for argument X;
> > }
> > int ALLOCATE(N) {
> > if (N == ARG.size() || N == NUM_RVV_REGS)
> > return 0;
> > ```
> > int dont_assign = ALLOCATE(N+1);
> > int assign = -1;
> > if (CAN_ALLOCATE(ARG[N]))
> >     assign = NEED(ARG[N]) + ALLOCATE(N+1);
> > 
> > if (assign >= dont_assign) {
> >     mark ARG[N] as byreg;
> >     return assign;
> > }
> > mark ARG[N] as byref;
> > return dont_assign;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > }
> > Note: The additional N == NUM_RVV_REGS constraint in ALLOCATE(N) is to limit the recursion times to NUM_RVV_REGS to avoid blowing up.
> 
> Before I get into the details, I'd like to know what the difference is between your proposed allocation rules and the current ones?

Current one just simply just allocate in order until we don't have any registers remain that meets the constraints.
For example,
if we have a function `func foo(vint32m8 a, vint32m1 b, vint32m8 c)`, current approach would assign `v8-v15` to `a` and `v16` to `b` and c would be passed by reference.

However the proposed approach would assign `v8-v15` to `a` and `v16-v23` to `c` and b would be passed by reference.

lhtin:
> Current one just simply just allocate in order until we don't have any registers remain that meets the constraints. For example, if we have a function `func foo(vint32m8 a, vint32m1 b, vint32m8 c)`, current approach would assign `v8-v15` to `a` and `v16` to `b` and c would be passed by reference.
> 
> However the proposed approach would assign `v8-v15` to `a` and `v16-v23` to `c` and b would be passed by reference.

Oh, it seems similar to your previous comment(https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/389#discussion_r1316075701). I don't have much to say about this at the moment, other than my previous comments. Wait for the others to see.

4vtomat:
> > Current one just simply just allocate in order until we don't have any registers remain that meets the constraints. For example, if we have a function `func foo(vint32m8 a, vint32m1 b, vint32m8 c)`, current approach would assign `v8-v15` to `a` and `v16` to `b` and c would be passed by reference.
> > However the proposed approach would assign `v8-v15` to `a` and `v16-v23` to `c` and b would be passed by reference.
> 
> Oh, it seems similar to your previous comment([#389 (comment)](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/389#discussion_r1316075701)). I don't have much to say about this at the moment, other than my previous comments. Wait for the others to see.

Yes, and I just discussed with @kito-cheng offline, we thought that if we can elaborate this that is easy to understand for everyone, and it's worth it to add the rule.
But like you said, we are welcome for comments.

workingjubilee:
Hmm. Does this proposal account for function pointers? I'm having a hard time imagining how this actually gets used by the C code that is supposed to be participating in this calling convention if the functions are taken by pointer. Do I have to use some kind of `[[riscv::attribute]]` (I guess?) on the function to indicate `STO_RISCV_VARIANT_CC`?

kito-cheng:
> Hmm. Does this proposal account for function pointers? I'm having a hard time imagining how this actually gets used by the C code that is supposed to be participating in this calling convention if the functions are taken by pointer. Do I have to use some kind of [[riscv::attribute]] (I guess?) on the function to indicate STO_RISCV_VARIANT_CC?

From high level aspect: yes, but actually psABI don't involved that implementation details, that's is mode like the scope of [riscv-c-api-doc](https://github.com/riscv-non-isa/riscv-c-api-doc)

workingjubilee:
Code pointer, not data pointer.

@kito-cheng In practice, "how do I call a function pointer?" is not _that_ high-level a concern, else languages that are _not_ C cannot fulfill cross-language ABI requirements. I am asking about implementation details for C compilers so I can get an idea of what sort of implementation details other compilers must implement for their *foreign-function interface*. Other psABIs certainly talk about how they *expect* language bindings to use them for external functions.

> Update 2023-6-22: After further understanding, there should be no need to add a new abi, but the function that uses the vector calling convention must be annotated with STO_RISCV_VARIANT_CC.

So I guess this _will_ add a new call ABI for Rust. An `extern "vector-cc" fn` will not be callable using the same call ABI as `extern "C" fn`, as C functions should default to passing vector arguments and returns on the stack, right?

That's somewhat regrettable, as on e.g. x86-64 SysV these types are (supposed to be) passed in a uniform way by `extern "C"`, so there is no split between the `extern "C" fn` argument/return by feature set... at least, hypothetically. Given how many C compilers fail to compile the x86-64 SysV calling convention correctly, and the fact [vectorcall is another traditional solution](https://learn.microsoft.com/en-us/cpp/cpp/vectorcall?view=msvc-170), it does seem like the safer choice.

lhtin:
> I am asking about implementation details for C compilers so I can get an idea of what sort of implementation details other compilers must implement for their foreign-function interface. Other psABIs certainly talk about how they expect language bindings to use them for external functions.

I understand that a function pointer will have a detailed declaration, right? This function declaration then implicitly states how the vector argument is passed. That is, if the compiler implementations corresponding to caller and callee respectively adhere to this proposal, then there is no problem. So why is it even necessary to explicitly indicate the attributes of a particular function pointer (e.g. `[[riscv::attribute]]`)?

workingjubilee:
@lhtin It is important because we have found that function pointers often **do not** contain all the information with respect to call ABI, especially once LLVM is done with everything and interprets things according to its own reasoning about ABI and features.

lhtin:
> @lhtin It is important because we have found that function pointers often **do not** contain all the information with respect to call ABI, especially once LLVM is done with everything and interprets things according to its own reasoning about ABI and features.

Can you give a detailed example?

jnk0le:
1. Were those benchmarks optimized by intrinsics or compilers autovectorizer?
2. Could you try all caller saved scenario with `[-Wl,]-mllvm,-enable-ipra` flag?

kito-cheng:
> Were those benchmarks optimized by intrinsics or compilers autovectorizer?

Both using intrinsics for Kepler's equation and Clohessy-Wiltshire equation:
- Kepler's equation: has uncountable loop, and current open source compilers for both GCC and LLVM can't auto-vectorize that yet.
- Clohessy-Wiltshire: Our downstream LLVM able to vectorized that, but it much worse than hand optimized intrinsic version yet.

> Could you try all caller saved scenario with [-Wl,]-mllvm,-enable-ipra flag?

Currently math libraries are separated from the benchmark, so I think it's not help by enable ipra, of cause we can rebuild benchmark with the library together, but that may be different from the real use case IMO?

cmuellner:
I second Andrews comment and I am looking forward to see this implemented.

The "best" balance between caller-saved and callee-saved registers only exists with additional assumptions (e.g. a given workload and a defined LMUL value). The evaluation that Brandon, Lehua and Kito made shows, that a half-half scheme is a reasonable choice. Many integer register CCs do this as well, but that's not fully applicable for a vector register CC.

Also, the way how the vector registers are grouped into caller-saved and callee-saved registers is done carefully, so that big LMUL values won't cause issues.

The preservation of the current calling convention for function calls without vector register usage is a great invitation to vectorize leaf functions (either with auto-vectorization, RVV intrinsics, or hand written assembly). And it has the nice side-effect that existing vectorized leaf functions (e.g. vectorized crypto routines in OpenSSL) don't have to be rewritten.

The only thing which I could not find is what will be done if we run out of vector registers (i.e. too many arguments). Will such arguments be passed as pointers to the stack, should compilers error out, or what is the expected behavior in this case?

aswaterman:
> The preservation of the current calling convention for function calls without vector register usage is a great invitation to vectorize leaf functions

My thoughts exactly.

rofirrim:
We, at BSC, implemented a similar vector calling convention on LLVM but in our case v1-v7 and v16-v31 are callee-saved. LLVM, in some circumstances, converts calls to some well-known math libraries into calls to internal intrinsics of LLVM. We have always lowered those into calls to a small vector math of ours using our vector convention.

In hindsight our split might have been a bit extreme.  For the functions present both in our vector math library and SKL, I informally checked the spills with those in the report of @4vtomat , @lhtin  and @kito-cheng  above (thanks a lot!). In our version, we have in many cases a larger number of spills in functions. This suggests the current proposal is more balanced for the use case of a vector math library. As reported in the document above, the callee does benefit from having extra non-callee-saved registers.

At BSC we want to be able to vectorise (usually side-effect free) functions that are called in vectorised loops via a mechanism such as OpenMP's `#pragma omp declare simd` directive. Those functions are typically called in vectorised loops (perhaps, again, due to a `#pragma omp simd` construct). As reported, we do risk having live vector values crossing function calls in those loops. In the default calling convention (where vector registers are all caller-saved) this means preserving and restoring them around the call. This may not be a big deal for a short running loop but it becomes more concerning if the invoked function may be implemented with few registers. E.g., both in our library and in the SKL, a function such `exp` may not require any spill when using the vector calling convention. This means that making the caller not prefer those registers (because they can be freely clobbered by the callee) may still be a good choice. Note, though, that this may not hold in general, so having the choice is always beneficial.

What we discarded early, though, was having the default calling convention to always have a set of callee-saved vector registers. As vectorized loops typically rely on hiding the memory latency by interspersing computation and memory accesses, it is not unusual to have high register pressure, with some loops adjusted to the brink of causing spilling. In those circumstances it is the norm to run out of non-callee-saved registers. This means that our functions, even leaf ones, had to pay the price of preserving and restoring those registers every time. This appeared in our traces and our users were not happy.  (This is particularly problematic for an implementation like ours where vectors are very long. Note that I am wary of suggesting our choices to the broad RVV ecosystem because our implementation is very particular and may not represent common implementation choices in this space). This problem might also ail a function being invoked from a vectorized loop, but those are, in my experience, less frequent, so it did not make sense to "pessimise" the common case.

In summary, I think this proposal makes a lot of sense and I support it.  The the proposal should not hurt existing codebases and at the same time provides an opt-in mechanism for those scenarios, like a vector math library, where it is seen a clear positive.

JeffreyALaw:
Generally makes sense.  My concerns are more about how this all interfaces with the rest of the system, say for example the glibc resolver trampoline and resolver, setjmp/longjmp and the like.  As long as provisions are made such that those routines can be properly implemented and we have some sensible mix of caller/callee saves, then I'm not going to complain.

kito-cheng (approved):
LGTM, thanks @lhtin and everyone!

I believe we gather enough feedback and address all major comments, so I think we are ready to land, however it's vacations season for most American and European, therefore I would like to wait few more time, my plan is merge this and #380 at 2024/1/5 IF no further strong comment/objection.

kito-cheng:
Thanks everyone, finally we have an official vector calling convention after few years :)

riscv-cc.adoc
```diff
@@ -329,6 +347,82 @@ type would be passed.
...
+be allocated to arguments in some cases.
+
+Vector values are returned in the same manner as the first named argument of
+the same type would be passed.
+
+Vector types are disallowed in struct or union.
+
+Vector arguments and return values are disallowed to pass to an unprototyped
+function.
+
```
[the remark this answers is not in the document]
jrtc27: This should not have been done like this. STO_RISCV_VARIANT_CC is an ELF-specific thing (a flag in ElfXX_Sym's st_other) detailed in riscv-elf.adoc, but riscv-cc.adoc is a separate document that describes the calling convention independently from the underlying file format (and could be used by Windows or macOS if they choose to adopt RISC-V, which use PE/COFF and Mach-O respectively).

jrtc27 (commented):
I note that this was merged despite being 21(!!) individual commits with all manner of fixups, making the history of the repository a mess. Can we please be more diligent in future about ensuring we merge a sensible series of commits?

PR #391, opened 2023-07-17
Ztso psABI atomics specification

patrick-rivos:
This PR defines an optimized atomics mapping ABI for Ztso targets. It is a strengthened version of Andrea Parri's proposed Ztso mappings ("Proposed Mapping") [1]. The proposal changes were discussed by Andrea Parri and Hans Boehm on the GCC mailing list [2] and are required in order to be compatible with the RVWMO A.7 ABI (mappings with note 3 in the psABI atomics table).

[1] https://github.com/preames/public-notes/blob/master/riscv-tso-mappings.rst
[2] https://inbox.sourceware.org/gcc-patches/ZFV8pNAstwrF2qBb@andrea/T/#t


patrick-rivos:
Patches:
GCC: https://inbox.sourceware.org/gcc-patches/20230717212836.23056-1-patrick@rivosinc.com/T/#u
LLVM: https://reviews.llvm.org/D155517

patrick-rivos:
@hboehm @kito-cheng @jrtc27 @ilovepi @nick-knight @asb @enh-google @daniellustig @aparri
This is atomics related and may be interesting to you.

hboehm:
Is there a reason to present these as an alternate mapping scheme, rather than as additional mappings that are enabled by Ztso? If not, I would prefer the latter presentation, for several reasons:

1) It involves less redundancy.
2) It makes it clear that these are compatible with the RVWMO mappings, which I think is extremely desirable.
3) I think it sets up a better precedent for the future. If we add more hardware support, like RCpc operations, that should be reflected here by adding mappings conditioned on that extension, not by inventing completely new mappings.
4) (I'm not enough of an ELF expert to be sure about this.) It makes it clear that we can indicate whether these mappings may be present by indicating in the object file which hardware extensions we rely on. I'm blindly assuming that there is a mechanism for that? If this is true, then we don't need additional atomics ABI bits for this in object files.

patrick-rivos:
To be clear, are you proposing a diff-only table like this?
|C/C++ Construct |Ztso Mapping |Notes
-|-|-
|`atomic_load(memory_order_acquire)` |`l{b\|h\|w\|d}` |
|`atomic_load(memory_order_seq_cst)` |`fence rw,rw; l{b\|h\|w\|d}` |
|`atomic_store(memory_order_release)` |`s{b\|h\|w\|d}` |
|`atomic_store(memory_order_seq_cst)` |`s{b\|h\|w\|d}; fence rw, rw` |
|`atomic_thread_fence(memory_order_acquire)` |`nop` |
|`atomic_thread_fence(memory_order_release)` |`nop` |
|`atomic_thread_fence(memory_order_acq_rel)` |`nop` |

|C/C++ Construct |Ztso AMO Mapping |Notes
-|-|-
|`atomic_<op>(memory_order_acquire)` |`amo<op>.{w\|d}` |4
|`atomic_<op>(memory_order_release)` |`amo<op>.{w\|d}` |4
|`atomic_<op>(memory_order_acq_rel)` |`amo<op>.{w\|d}` |4
|`atomic_<op>(memory_order_seq_cst)` |`amo<op>.{w\|d}` |4

|C/C++ Construct |Ztso LR/SC Mapping |Notes
-|-|-
|`atomic_<op>(memory_order_acquire)`|`loop:lr.{w\|d}; <op>; sc.{w\|d}; bnez loop` |4
|`atomic_<op>(memory_order_release)`|`loop:lr.{w\|d}; <op>; sc.{w\|d}; bnez loop` |4
|`atomic_<op>(memory_order_acq_rel)`|`loop:lr.{w\|d}; <op>; sc.{w\|d}; bnez loop` |4

I did separate mappings since it made it easier for me to verify my implementations were consistent with the table.
I'm not opposed to presenting a diff-only table, but I also think there is some value in discouraging a mix-and-match approach to atomic mappings. Clear discrete "blessed" mappings can help discourage bespoke collections that misinterpreted the notes and are broken in subtle ways.

Regarding point 4: I'm also not enough of an ELF expert, but I know we have a [TSO bit allocated in e_flags](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#file-header) for Ztso binaries.

hboehm:
Indeed. I would prefer the "diff-only" table, noting that it can contain only additions.

In my view, mix-and-match is completely unavoidable at the whole application level, so it would be better to be explicit about that in the spec. I also think we are likely to eventually end up with too many combinations to reasonably list all of them in the spec. Compiler writers will have to pick the right combination for the specified target hardware, and for their performance expectations. And we have to convince ourselves that they all play together (with the existing disclaimers about the A.6 mappings).

patrick-rivos:
I reformatted the table/reworded some of the surrounding text to make it more clear that these are additional mappings enabled by the Ztso extension.

I also changed the wording for the TSO bit to allow linking RVWMO/TSO binaries in a similar way to how RVC binaries are handled.

kito-cheng:
For the atomic flag stuffs, we will need add a `TSO_ATOMIC_ABI` and then allow other to merged with that (except the potential soft atomic ABI in future), and then the merge result is `TSO_ATOMIC_ABI`, does it make sense to you? @patrick-rivos @hboehm

patrick-rivos:
The merge result still needs to be marked as `A6C_ATOMIC_ABI` or `A6S_ATOMIC_ABI` in addition to TSO since the mix-and-match allows for TSO mappings that are not A6C compatible. 

Eg. `atomic_<op>(memory_order_seq_cst) | loop:lr.{w\|d}.aq; <op>; sc.{w\|d}.rl; bnez loop` is not A6C compatible - even with a TSO memory model.

IIUC, the `TSO_ATOMIC_ABI` case is already covered by the TSO bit in the `e_flags` and `{A6C|A6S}_ATOMIC_ABI` is needed in addition to that bit.

kito-cheng:
> The merge result still needs to be marked as `A6C_ATOMIC_ABI` or `A6S_ATOMIC_ABI` in addition to TSO since the mix-and-match allows for TSO mappings that are not A6C compatible.

`A6C_ATOMIC_ABI` merged with `TSO_ATOMIC_ABI` is not conformance with `A6C_ATOMIC_ABI` and not work on the machine without `TSO_ATOMIC_ABI`, so I think the merge result is `TSO_ATOMIC_ABI`, and should be same as `A6S_ATOMIC_ABI`.

> Eg. `atomic_<op>(memory_order_seq_cst) | loop:lr.{w\|d}.aq; <op>; sc.{w\|d}.rl; bnez loop` is not A6C compatible - even with a TSO memory model.
> 
> IIUC, the `TSO_ATOMIC_ABI` case is already covered by the TSO bit in the `e_flags` and `{A6C|A6S}_ATOMIC_ABI` is needed in addition to that bit.

My point is TSO mapping is not make sense to set either `A6C_ATOMIC_ABI` or `A6S_ATOMIC_ABI`, and set it to `UNKNOWN_ATOMIC_ABI` also not good idea since that will allow it can be merged with  `A6C_ATOMIC_ABI` or `A6S_ATOMIC_ABI`, and marking the object with `TSO_ATOMIC_ABI` can make sure that's using updated (this PR) TSO mapping.

asb:
Is the TSO bit in `e_flags` still useful, or is it effectively superseded by defining `TSO_ATOMIC_ABI`?

kito-cheng:
> Is the TSO bit in e_flags still useful, or is it effectively superseded by defining TSO_ATOMIC_ABI?

It's less useful, but it provide a way to do faster check for dynamic linker/program loader if needed, and it...allocated for a while, so I tend to keep as it.

sorear:
I would prefer to treat Ztso as a pure ISA extension which allows certain fences to be elided, and have object files which assume Ztso semantics for loads and stores claim conformance to the RVWMO memory models without adding additional complexity to the ATOMIC_ABI selection.

kito-cheng:
> I would prefer to treat Ztso as a pure ISA extension which allows certain fences to be elided, and have object files which assume Ztso semantics for loads and stores claim conformance to the RVWMO memory models without adding additional complexity to the ATOMIC_ABI selection.

But TSO mapping (this PR) is not compatible/conformance either `A6C_ATOMIC_ABI` or `A6S_ATOMIC_ABI`?

IMO that will increase the complexity of ATOMIC_ABI since we also need to consider e_flags.EF_RISCV_TSO when interpreter which atomic mapping are using for the object/executable.

hboehm:
I gather there have also been discussions about 8- and 16-bit variants of the AMO operations. If those materialize, that would be another instance in which we would want to extend the mappings in the presence of an ISA extension.  I think it would be great if such a future extension could just follow the model we use for Ztso here.

kito-cheng:
I am struggling what is the right merge behavior and what's the meaning of those flags (ATOMIC_ABI and EF_RISCV_TSO), so I trying to write my understanding and enumerate the situation (w/o adding TSO_ATOMIC_ABI and w/ A7_ATOMIC_ABI), please correct me if anything you feel wrong :)

EF_RISCV_TSO means the ELF require run on the machine has Ztso extension.
ATOMIC_ABI present which atomic mapping is begin used in the object/executable, if mixed more than one mapping, try to set it as minimal conformance one.

We have few objects using different atomic mapping.
A6C.o
A6S.o
TSO.o
A7.o

# ATOMIC_ABI + EF_RISCV_TSO

So when no mixing different mapping, those setting would be:

|  Input Objects | ATOMIC_ABI | EF_RISCV_TSO |  NOTE |
| --------------------- | -------------------- | --------------------- | ----- |
| A6C.o               | A6C_ATOMIC_ABI   | 0 | |
| A6S.o               | A6S_ATOMIC_ABI  | 0 |  |
| TSO.o               | UNKNWON_ATOMIC_ABI  | 1 |  |
| A7.o               | A7_ATOMIC_ABI  | 0 |  |


OK, let combination the situation:

|  Input Objects | ATOMIC_ABI | EF_RISCV_TSO |  NOTE |
| --------------------- | -------------------- | --------------------- | ----- |
| A6C.o + A6S.o              | A6C_ATOMIC_ABI   | 0 | Mixing A6C and A6S, made this won't conformance with A6S. |
| A6C.o + TSO.o               | UNKNWON_ATOMIC_ABI  | 1 | TSO mapping is not conformance with A6C, but it can run correctly on TSO machine, so set EF_RISCV_TSO=1 to require machine with TSO to run correctly. |
| A6C.o + A7.o               | x  | x | Report error, A6C and A7 mapping can't work together correctly. |
| A6S.o + TSO.o               | UNKNWON_ATOMIC_ABI  | 1 | TSO mapping is not conformance with A6S, but it can run correctly on TSO machine, so set EF_RISCV_TSO=1 to require machine with TSO to run correctly. |
| A6S.o + A7.o | A6S_ATOMIC_ABI | 0 | It conformance with A6S_ATOMIC_ABI but require `Zacas`. |
| TSO.o + A7.o |  UNKNWON_ATOMIC_ABI | 1  | TSO mapping is not conformance with A7, and it require run on machine has TSO *AND* `Zacas`. |
| A6C.o + A6S.o + TSO.o | UNKNWON_ATOMIC_ABI | 1 | TSO mapping is not conformance with A6C and A6S, but it can run correctly on TSO machine, so set EF_RISCV_TSO=1 to require machine with TSO to run correctly. |
| A6C.o + A6S.o + A7.o | x | x | Report error, A6C and A7 mapping can't work together correctly.
| A6S.o + A7.o + TSO.o | UNKNWON_ATOMIC_ABI |1 | A6S and A7 is compatible, but TSO mapping is not conformance with A7, so it require run on machine has TSO and `Zacas` |

---

# ATOMIC_ABI (w/ TSO_ATOMIC_ABI) + EF_RISCV_TSO

And try to add TSO_ATOMIC_ABI

|  Input Objects | ATOMIC_ABI | EF_RISCV_TSO |  NOTE |
| --------------------- | -------------------- | --------------------- | ----- |
| A6C.o + A6S.o              | A6C_ATOMIC_ABI   | 0 | Mixing A6C and A6S, made this won't conformance with A6S. |
| A6C.o + TSO.o               | TSO_ATOMIC_ABI  | 1 | TSO mapping is not conformance with A6C, but A6C is conformance/stronger than TSO, so result is TSO_ATOMIC_ABI |
| A6C.o + A7.o               | x  | x | Report error, A6C and A7 mapping can't work together correctly. |
| A6S.o + TSO.o               | TSO_ATOMIC_ABI  | 1 | TSO mapping is not conformance with A6S, but A6S is conformance/stronger than TSO, so result is TSO_ATOMIC_ABI |
| A6S.o + A7.o | A6S_ATOMIC_ABI | 0 | It conformance with A6S_ATOMIC_ABI but require `Zacas`. |
| TSO.o + A7.o |  TSO_ATOMIC_ABI | 1  | TSO mapping is not conformance with A7, but A7 is conformance/stronger than TSO, so result is TSO_ATOMIC_ABI and it require run on machine has TSO *AND* `Zacas`. |
| A6C.o + A6S.o + TSO.o | TSO_ATOMIC_ABI | 1 | TSO mapping is not conformance with A6C and A6S, but A6C and A6S are conformance/stronger than TSO, so result is TSO_ATOMIC_ABI, so set EF_RISCV_TSO=1 to require machine with TSO to run correctly. |
| A6C.o + A6S.o + A7.o | x | x | Report error, A6C and A7 mapping can't work together correctly.
| A6S.o + A7.o + TSO.o | TSO_ATOMIC_ABI |1 | A6S and A7 is compatible, but A6S and A7 are conformance/stronger than TSO, so it require run on machine has TSO and `Zacas` |


---

# Two bit scheme (EF_RISCV_ATOMIC_A6/EF_RISCV_ATOMIC_A7)

And let me try to using 2 bit scheme which proposed by https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/383, and we can use similar scheme on ELF attribute if we think it's right way.

ATOMIC_A6 = EF_RISCV_ATOMIC_A6
ATOMIC_A7 = EF_RISCV_ATOMIC_A7
TSO = EF_RISCV_TSO

|  Input Objects | ATOMIC_A6 | ATOMIC_A7  | TSO |  NOTE |
| --------------------- | -------------------- | ------------------- | --------------------- | ----- |
| A6C.o               | 1 | 0   | 0 | |
| A6S.o               | 1 | 1 | 0 | |
| TSO.o               | 0 | 0 | 1 |  |
| A7.o                 | 1 | 1  | 0 |  |

|  Input Objects | ATOMIC_A6 | ATOMIC_A7 | TSO |  NOTE |
| --------------------- | -------------------- | --------------------- | ----- |---- |
| A6C.o + A6S.o              | 1 | 0   | 0 | Not compatible with A7 |
| A6C.o + TSO.o               | 0 | 0  | 1 | Not compatibe with either A6 or A7 |
| A6C.o + A7.o               | x  | x | x | Report error, A6C and A7 mapping can't work together correctly. |
| A6S.o + TSO.o               | 0  | 0 | 1 | Not compatible with either A6 or A7, but require TSO |
| A6S.o + A7.o | 0 | 1 | 0 | Compatible with A7 but require `Zacas`. |
| TSO.o + A7.o |  0 | 0 | 1  | Not compatible with either A6 or A7, but require TSO and `Zacas` |
| A6C.o + A6S.o + TSO.o | 0 |0 | 1 | Not compatible with either A6 or A7, but require TSO|
| A6C.o + A6S.o + A7.o | x | x | x | Report error, A6C and A7 mapping can't work together correctly. |
| A6S.o + A7.o + TSO.o | 0 | 0 |1 | Not compatible with either A6 or A7, but require TSO and `Zacas` |

sorear:
It may be worth noting that the only thing EF_RISCV_TSO=1 allows you to do is delete fences (and aqrl bits) that are already no-ops on Ztso implementations, which means that it will only marginally help on Ztso implementations that can delete no-ops early in the frontend.

@kito-cheng

My argument is that TSO objects should be marked with the atomic ABI they are compatible with, so for the 2-bit version TSO.o should have all of ATOMIC_A6, ATOMIC_A7, and TSO = 1; for the #385 version TSO.o should have A6S_ATOMIC_ABI and EF_RISCV_TSO=1.

patrick-rivos:
I think the Two bit scheme (EF_RISCV_ATOMIC_A6/EF_RISCV_ATOMIC_A7) is the closest table you have to what I was thinking of.

The *_TSO mappings have been made with RVWMO compatibility in mind, so the TSO bit can be safely ignored when deciding compatibility (as long as the TSO bit is set if one of the inputs has the TSO bit set).

First, the compatibility between UNKNOWN, A6C, A6S, and A7.
Inputs ATOMIC_ABI | Output ATOMIC_ABI | NOTE
-|-|-
UNKNOWN + UNKNOWN | UNKNOWN |
A6C + A6C | A6C |
A6S + A6S | A6S |
A7 + A7 | A7 |
UNKNOWN + A6C | A6C | Must output the least-compatible ABI of the inputs. This doesn't guarantee the resulting binary is fully compatible (since one of the inputs ABI's is unknown), but it allows us to not error on existing binaries.
UNKNOWN + A6S | A6S | This doesn't guarantee the resulting binary is fully compatible (since one of the inputs ABI's is unknown), but it allows us to not error on existing binaries.
UNKNOWN + A7 | A7 | Must output the least-compatible ABI of the inputs. This doesn't guarantee the resulting binary is fully compatible (since one of the inputs ABI's is unknown), but it allows us to not error on existing binaries.
A6C + A6S | A6C | Must output the least-compatible ABI of the inputs
A6C + A7 | INVALID | Throw an error, A6C is incompatible with A7
A6S + A7 | A7 | Must output the least-compatible ABI of the inputs

Next, the behavior of TSO
Inputs EF_RISCV_TSO | Output EF_RISCV_TSO | NOTE
-|-|-
0 + 0 | 0 | Both RVWMO inputs means an RVWMO output
1 + 0 | 1 | One TSO input means TSO output
0 + 1 | 1 | One TSO input means TSO output
1 + 1 | 1 | Both TSO inputs means TSO output

Now, because the TSO mappings were created to have the same ABIs and be compatible with RVWMO mappings, we can just combine the two tables:

Inputs ATOMIC_ABI | Inputs EF_RISCV_TSO | Output ATOMIC_ABI | Output EF_RISCV_TSO | NOTE
-|-|-|-|-
UNKNOWN + UNKNOWN | 0 + 0  | UNKNOWN | 0 |
A6C + A6C |0 + 0 | A6C | 0 |
A6S + A6S|0 + 0  | A6S | 0 |
A7 + A7 |0 + 0 | A7 | 0 |
UNKNOWN + A6C |0 + 0 | A6C | 0 | Must output the least-compatible ABI of the inputs. This doesn't guarantee the resulting binary is fully compatible (since one of the inputs ABI's is unknown), but it allows us to not error on existing binaries.
UNKNOWN + A6S |0 + 0 | A6S | 0 | This doesn't guarantee the resulting binary is fully compatible (since one of the inputs ABI's is unknown), but it allows us to not error on existing binaries.
UNKNOWN + A7|0 + 0 | A7 | 0 | Must output the least-compatible ABI of the inputs. This doesn't guarantee the resulting binary is fully compatible (since one of the inputs ABI's is unknown), but it allows us to not error on existing binaries.
A6C + A6S |0 + 0 | A6C | 0 | Must output the least-compatible ABI of the inputs
A6C + A7 |0 + 0 | INVALID | 0 | Throw an error, A6C is incompatible with A7
A6S + A7 |0 + 0 | A7 | 0 | Must output the least-compatible ABI of the inputs
UNKNOWN + UNKNOWN | 0 + 1  | UNKNOWN | 1 | One TSO input means TSO output
A6C + A6C |0 + 1  | A6C | 1 |
A6S + A6S|0 + 1  | A6S | 1 |
A7 + A7 |0 + 1 | A7 | 1 |
UNKNOWN + A6C |0 + 1 | A6C | 1 | Must output the least-compatible ABI of the inputs. This doesn't guarantee the resulting binary is fully compatible (since one of the inputs ABI's is unknown), but it allows us to not error on existing binaries.
UNKNOWN + A6S |0 + 1 | A6S | 1 | This doesn't guarantee the resulting binary is fully compatible (since one of the inputs ABI's is unknown), but it allows us to not error on existing binaries.
UNKNOWN + A7|0 + 1 | A7 | 1 | Must output the least-compatible ABI of the inputs. This doesn't guarantee the resulting binary is fully compatible (since one of the inputs ABI's is unknown), but it allows us to not error on existing binaries.
A6C + A6S |0 + 1 | A6C | 1 | Must output the least-compatible ABI of the inputs
A6C + A7 |0 + 1 | INVALID | 1 | Throw an error, A6C is incompatible with A7
A6S + A7 |0 + 1 | A7 | 1 | Must output the least-compatible ABI of the inputs
UNKNOWN + UNKNOWN | 1 + 0  | UNKNOWN | 1 | One TSO input means TSO output
A6C + A6C |1 + 0 | A6C | 1 |
A6S + A6S|1 + 0 | A6S | 1 |
A7 + A7 |1 + 0 | A7 | 1 |
UNKNOWN + A6C |1 + 0 | A6C | 1 | Must output the least-compatible ABI of the inputs. This doesn't guarantee the resulting binary is fully compatible (since one of the inputs ABI's is unknown), but it allows us to not error on existing binaries.
UNKNOWN + A6S |1 + 0 | A6S | 1 | This doesn't guarantee the resulting binary is fully compatible (since one of the inputs ABI's is unknown), but it allows us to not error on existing binaries.
UNKNOWN + A7|1 + 0  | A7 | 1 | Must output the least-compatible ABI of the inputs. This doesn't guarantee the resulting binary is fully compatible (since one of the inputs ABI's is unknown), but it allows us to not error on existing binaries.
A6C + A6S |1 + 0 | A6C | 1 | Must output the least-compatible ABI of the inputs
A6C + A7 |1 + 0 | INVALID | 1 | Throw an error, A6C is incompatible with A7
A6S + A7 |1 + 0 | A7 | 1 | Must output the least-compatible ABI of the inputs
UNKNOWN + UNKNOWN | 1 + 1  | UNKNOWN | 1 | Both TSO input means TSO output
A6C + A6C |1 + 1 | A6C | 1 |
A6S + A6S|1 + 1  | A6S | 1 |
A7 + A7 |1 + 1 | A7 | 1 |
UNKNOWN + A6C |1 + 1 | A6C | 1 | Must output the least-compatible ABI of the inputs. This doesn't guarantee the resulting binary is fully compatible (since one of the inputs ABI's is unknown), but it allows us to not error on existing binaries.
UNKNOWN + A6S |1 + 1 | A6S | 1 | This doesn't guarantee the resulting binary is fully compatible (since one of the inputs ABI's is unknown), but it allows us to not error on existing binaries.
UNKNOWN + A7|1 + 1 | A7 | 1 | Must output the least-compatible ABI of the inputs. This doesn't guarantee the resulting binary is fully compatible (since one of the inputs ABI's is unknown), but it allows us to not error on existing binaries.
A6C + A6S |1 + 1 | A6C | 1 | Must output the least-compatible ABI of the inputs
A6C + A7 |1 + 1 | INVALID | 1 | Throw an error, A6C is incompatible with A7
A6S + A7 |1 + 1 | A7 | 1 | Must output the least-compatible ABI of the inputs

It's a bit unwieldy but shows all the 2-input states :)

Covering some 3 input cases:
Inputs ATOMIC_ABI | Inputs EF_RISCV_TSO | Output ATOMIC_ABI | Output EF_RISCV_TSO | NOTE
-|-|-|-|-
A6C + A6S + A6S | 0 + 0 + 1  | A6C | 1 | A TSO A6S object is compatible with an A6C RVWMO object, as long as the resulting object is run on a TSO machine.
A6C + A7 + A6S | 0 + 0 + 1  | INVALID | 1 | Report an error, A6C and A7 do not work together
A6C + A6S + A7 | 0 + 0 + 1  | INVALID | 1 | Report an error, A6C and A7 do not work together (even if one of them is TSO)
A6S + A7 + A7 | 0 + 0 + 1  | A7 | 1 | Atomic ABIs are compatible between TSO/RVWMO, as long as the resulting object is run on a TSO machine.
A6S + A6C + A6S | 0 + 0 + 1  | A6C | 1 | Atomic ABIs are compatible between TSO/RVWMO, as long as the resulting object is run on a TSO machine.
 
Regarding your tables, there are a few entries that I think are incorrect.

You can implement an A7 binary today using a note 3 mapping like:
`atomic_<op>(memory_order_seq_cst)`|`loop:lr.{w\|d}.aq; <op>; sc.{w\|d}.rl; bnez loop` |3, 4
A7 doesn't require Zacas.

Input Objects | ATOMIC_ABI
-|-
A6S.o + A7.o | A6S_ATOMIC_ABI 

This isn't right since A7 is the more restrictive of the two ABIs, so we should be outputting A7_ATOMIC_ABI.

hboehm:
I agree with Patrick.

The rules for UNKNOWN might be debatable, since UNKNOWN also applies to old gcc-generated code. But I'm OK with Patrick's argument.

If we decompose the ATOMIC_ABI value into an A6C-compatible bit and an A7-compatible bit, treating 00 as UNKNOWN,  with A6S = 11, then the ATOMIC_ABI combination rule becomes:

If either object file has tag 00 (UNKNOWN), then the result tag is the other one, otherwise
bitwise AND the two together. If the result is 0 report invalid, otherwise the AND result is the tag for the combination.

That description wins on brevity, I think, but perhaps not on clarity. Either description is fine with me.

kito-cheng:
@patrick-rivos

Thanks for expanding the table, that's really much clear to me, I think what I stuck is I am trying to define/understand ATOMIC_ABI is minimal conformance rather than compatible, so that's why I think A6S + A7 should be A6S_ATOMIC_ABI since I treat  A6S_ATOMIC_ABI is less stronger than A7_ATOMIC_ABI.

And one last thing I want to make sure is what ATOMIC_ABI should be used if compiler using this TSO mapping? does @sorear's description same as what you think?

@patrick-rivos @hboehm @sorear 

It's really nice discussion here, that fix my server wrong understanding around there :)
And I tend to keep using #385 scheme, but will updated with @patrick-rivos's merge rule, the reason is I want to make sure that able to extend for future soft-atomic stuffs.

patrick-rivos:
> @patrick-rivos
> 
> Thanks for expanding the table, that's really much clear to me, I think what I stuck is I am trying to define/understand ATOMIC_ABI is minimal conformance rather than compatible, so that's why I think A6S + A7 should be A6S_ATOMIC_ABI since I treat A6S_ATOMIC_ABI is less stronger than A7_ATOMIC_ABI.

Glad it was helpful. The way I think about it is that A6S is a compatibility mapping, so the general flow when merging ABIs is like this:
```
  UNKNOWN
    /|\ 
  /  ↓  \
 |  A6S  |
 ↓ ↙   ↘ ↓
A6C      A7
```
You can never go back up a level, so merging ABIs will always move you to the lowest level of your inputs.

Feel free to ignore this if it's confusing - it's the same as the table, just different way of displaying it. Also please forgive the bad ascii art ;)

> And one last thing I want to make sure is what ATOMIC_ABI should be used if compiler using this TSO mapping? does @sorear's description same as what you think?

Yes @sorear's description is the same as what I was thinking. The compiler should emit A6S_ATOMIC_ABI if no Note3 mappings are used. If Note3 mappings are used then we need to emit A7_ATOMIC_ABI.

riscv-atomic.adoc
```diff
@@ -148,11 +148,89 @@ mappings with code generated by a compiler using those older mappings.
...
+We present the mappings as a table in 3 sections. The first
+deals with translations for loads, stores, and fences. The next two sections
+address mappings for read-modify-write operations like `fetch_add`, and
+`exchange`. The second section deals with operations that have direct
+`amo` instruction equivalents in the RISC-V A extension. The final
+section deals with other read-modify-write operations that require
+the `lr` and `sc` instructions.
+
+
+[[tab:c11mappings]]
```
kito-cheng, line 173: `[[c11mappingstso]]`, prevent use same table name

riscv-atomic.adoc
```diff
@@ -148,11 +148,89 @@ mappings with code generated by a compiler using those older mappings.
...
+.Mappings from C/C++ primitives to RISC-V primitives
```
kito-cheng, line 174: Maybe `Mappings with Ztso from C/C++ primitives to RISC-V primitives`?, what I would like to suggest is don't use same title for the table.

patrick-rivos: Updated to `Mappings with Ztso extension from C/C++ primitives to RISC-V primitives`

riscv-atomic.adoc
```diff
@@ -148,11 +148,89 @@ mappings with code generated by a compiler using those older mappings.
...
-=== Other conventions
+== Ztso atomics mappings
 
-It is expected that the RVWMO AMO Mappings will be used for atomic read-modify-write
-operations that are directly supported by corresponding AMO instructions,
-and that LR/SC mappings will be used for the remainder, currently
+This specifies additional mappings of C and C\++ atomic operations to RISC-V
+machine instructions. Other languages, for example Java, provide similar
+facilities that should be implemented in a consistent manner, usually
+by applying the mapping for the corresponding C++ primitive.
```
hboehm, line 156: Do we want to duplicate the "Other languages ..." text here? It already appears above in the non-tso section. It reads better to me without it, but I don't have a strong opinion.

riscv-atomic.adoc
```diff
@@ -148,11 +148,89 @@ mappings with code generated by a compiler using those older mappings.
...
+All mappings interoperate correctly with the RVWMO mappings, and with the
+original "Table A.6" mappings.
+
+We present the mappings as a table in 3 sections. The first
+deals with translations for loads, stores, and fences. The next two sections
+address mappings for read-modify-write operations like `fetch_add`, and
+`exchange`. The second section deals with operations that have direct
+`amo` instruction equivalents in the RISC-V A extension. The final
+section deals with other read-modify-write operations that require
+the `lr` and `sc` instructions.
```
hboehm, line 170: Ditto. "We present the table in 3 sections, as above." would do, IMO. Again no strong opinion.

hboehm:
LGTM, whether or not you accept my two minor suggestions above.

patrick-rivos:
I think all the feedback has been addressed. LMK if I missed anything :)

jrtc27:
Please tidy up the commit history

riscv-atomic.adoc
```diff
@@ -148,11 +148,80 @@ mappings with code generated by a compiler using those older mappings.
 
 4) Currently only directly possible for 32- and 64-bit operands.
 
-=== Other conventions
+== Ztso atomics mappings
 
-It is expected that the RVWMO AMO Mappings will be used for atomic read-modify-write
-operations that are directly supported by corresponding AMO instructions,
-and that LR/SC mappings will be used for the remainder, currently
+This specifies additional mappings of C and C\++ atomic operations to RISC-V
```
jrtc27, line 153: {Cpp}

patrick-rivos:
Squashed the commits and replaced `C++`/`C\++` occurrences with `{Cpp}`.

kito-cheng (approved):
LGTM, thanks @patrick-rivos and @hboehm !

PR #392, opened 2023-07-18
Add R_RISCV_ALIGN_DOWN

palmer-dabbelt:
We're in the process of stumbling into another round of these bugs related to insufficient alignment bytes.  The idea of adding another alignment reloation that rounds down has come up a few times, but I don't remember if anyone's actually written the spec.  So here's one.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -452,7 +452,9 @@ Description:: Additional information about the relocation
                                             <| S + A
 .2+| 61      .2+| SUB_ULEB128   .2+| Static  | _ULEB128_         .2+| Local label subtraction <<uleb128-note,*note>>
                                             <| V - S - A
-.2+| 62-191  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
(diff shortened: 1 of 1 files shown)


jrtc27:
This needs justification within this forum

kito-cheng:
That remind me that we have similar issue for R_RISCV_RELAX: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/116, and @jrtc27 has proposed we can point a dummy ISA symbol by the `R_RISCV_RELAX`, and I think...that could also use similar mechanism here? and IIRC @ptomsich has mention RISC-V not implement max alignment for the `.align` directive, so I guess...we could resolve together instead of add two more relocation (one for R_RISCV_ALIGN_DOWN and one for R_RISCV_ALIGN_MAX) .

c.c. @Nelson1225  @MaskRay

MaskRay:
Thanks for tagging me. To address issues like https://github.com/riscv-collab/riscv-gnu-toolchain/issues/445 , I agree that discerning RVC vs non-RVC `R_RISCV_RELAX` is important, and we can use @jrtc27's idea to use the symbol index part of the `r_info` field (say, `symbol_index!=0` means non-RVC). This scheme is compatible with existing linkers.

I haven't thought much `R_RISCV_ALIGN`. Right now I just know that the two fields in the syntax `.align [abs-expr[, abs-expr[, abs-expr]]]` are not representable, but it seems that people are fine with the limitation. A new relocation type doesn't seem to be needed for the RVC vs non-RVC issue?

PR #394, opened 2023-08-12
Add GP-relative relocations

luismarques:
Following-up on @kito-cheng's suggestion to add the ePIC [proposal](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/343) content to the existing psABI documents, this initial PR adds the GP-relative relocations. These GP-relocations should be useful for purposes besides ePIC (e.g. see previous proposals of the compact/large code models).

For ease of review, I will make a separate PR to add `R_RISCV_GPREL_ADD`, used for relaxation purposes, unless you prefer otherwise.

I'm not sure these GP relocations merit a GP-relative addressing section, such as the existing "PC-Relative Symbol Addresses", as there aren't the same kind of complications to comment on. If you feel it's needed, I will add one to this PR.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -428,7 +428,13 @@ Description:: Additional information about the relocation
                                             <| S + A - P
 .2+| 46      .2+| RVC_LUI       .2+| Static  | _CI-Type_         .2+| High 6 bits of 18-bit absolute address
                                             <| S + A
-.2+| 47-50   .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
+.2+| 47      .2+| GPREL_LO12_I  .2+| Static  | _I-type_          .2+| Low 12 bits of a 32-bit GP-relative address, `%gprel_lo(symbol)`
+                                            <| S + A - GP
+.2+| 48      .2+| GPREL_LO12_S  .2+| Static  | _S-Type_          .2+| Low 12 bits of a 32-bit GP-relative address, `%gprel_lo(symbol)`
+                                            <| S + A - GP
+.2+| 49      .2+| GPREL_HI20    .2+| Static  | _U-Type_          .2+| High 20 bits of a 32-bit GP-relative address, `%gprel_hi(symbol)`
+                                            <| S + A - GP
+.2+| 50      .2+| *Reserved*                          .2+| -        |                  .2+| Reserved for future standard use
                                             <|
 .2+| 51      .2+| RELAX         .2+| Static  |                   .2+| Instruction can be relaxed, paired with a normal relocation at the same address
                                             <|


riscv-elf.adoc
```diff
@@ -452,7 +452,13 @@ Description:: Additional information about the relocation
                                             <| S + A
 .2+| 61      .2+| SUB_ULEB128   .2+| Static  | _ULEB128_         .2+| Local label subtraction <<uleb128-note,*note>>
                                             <| V - S - A
-.2+| 62-191  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
+.2+| 62      .2+| GPREL_HI20    .2+| Static  | _U-Type_          .2+| High 20 bits of a 32-bit GP-relative address, `%gprel_hi(symbol)`
+                                            <| S + A - GP
+.2+| 63      .2+| GPREL_LO12_I  .2+| Static  | _I-type_          .2+| Low 12 bits of a 32-bit GP-relative address, `%gprel_lo(symbol)`
+                                            <| S + A - GP
+.2+| 64      .2+| GPREL_LO12_S  .2+| Static  | _S-Type_          .2+| Low 12 bits of a 32-bit GP-relative address, `%gprel_lo(symbol)`
+                                            <| S + A - GP
```
kito-cheng, line 460: Let use 47 and 48 for GPREL_LO12_I and GPREL_LO12_S
https://github.com/riscv-non-isa/riscv-elf-psabi-doc/commit/44f98e0fd8104def00f2a5a8d94b23dd647d18fb

luismarques: Is there any particular reason why you didn't suggest moving the `GPREL_HI20` as well, to position 49?

(BTW, I'll try to be responsive but I'm currently away on vacation, so apologies for any additional reply latency)

jrtc27: Grouping them makes sense. GPREL_LO12_I isn't the same as GPREL_I anyway; the former is the lo12 of a 32-bit value, whilst the latter requires the value to be 12-bit.

kito-cheng: I asked that since I thought GPREL_LO12_I/S are same GPREL_I/S due to their calculation expression are same, I have no strong opinion on GPREL_HI20, it can stay 64 or moving to 49, either is fine to me.

kito-cheng:
Did you mind post the link of the lld or ld.bfd implementation for the process in the psABI - we require PoC for either linker :)

luismarques:
> Did you mind post the link of the lld or ld.bfd implementation for the process in the psABI - we require PoC for either linker :)

Here's a link to the first implementation I am aware of of these GPREL relocations, by Evandro and Nelson / SiFive, for the bfd linker: <https://github.com/ebahapo/riscv-binutils-gdb/commit/615efc0c960b15eebaf20dde6253283f40cc3d7e>.

kito-cheng:
cc @Nelson1225 @MaskRay linker experts

kito-cheng (approved):
LGTM from my side, just wait few more response from linker experts

Nelson1225:
LGTM, too.  It's good to see this series is still going on :)

luismarques:
> cc @Nelson1225 @MaskRay linker experts

ping @MaskRay and other linker experts

rui314:
This is looking good, but we are not going to use these relocations without relaxation because the size matters for the use case, no? If so, I think we want to define the relaxation scheme at the same time.

luismarques:
> This is looking good, but we are not going to use these relocations without relaxation because the size matters for the use case, no? If so, I think we want to define the relaxation scheme at the same time.

I have a use case where relaxation is not essential. I've decided to split relaxation into a separate PR for ease of review but if you really want I can extend this PR.

rui314:
I feel more comfortable with relaxation because we want to make sure that these relocations are actually relaxable. But I think that's not really a concern, as they can obviously be relaxable in the same way as `R_RISCV_TPREL_*` relocs. So I'm fine with this. What do you think @MaskRay?

MaskRay (approved):
Looks fine to me as well. These relocations can be used when the segment containing GP does not have a link-time constant distance from the program counter.

kito-cheng:
I am OK to add relaxation later, let moving forward :)

sorear:
I'd like to revert this before it gets into a tagged release.  Neither ePIC nor FDPIC will use it except possibly for linker-internal purposes after relaxation, and we don't assign numbers to linker-internal relocation types.  My apologies for missing this earlier.

MaskRay:
> I'd like to revert this before it gets into a tagged release. Neither ePIC nor FDPIC will use it except possibly for linker-internal purposes after relaxation, and we don't assign numbers to linker-internal relocation types. My apologies for missing this earlier.

I think `S + A - GP` can be used with FDPIC, which is the reason I think this PR is acceptable.

I am indeed concerned with the rest part of ePIC, so I cannot sign off on them.

sorear:
You can only use it for non-preemptible symbols if you know for a fact at the reference site that the referenced symbol is allocated to writable memory.  I don't think it's going to get anywhere near the amount of usage required to justify 1/48 of the RV32 relocation space.

MaskRay:
> You can only use it for non-preemptible symbols if you know for a fact at the reference site that the referenced symbol is allocated to writable memory. I don't think it's going to get anywhere near the amount of usage required to justify 1/48 of the RV32 relocation space.

I agree with you. I have also checked GCC arm and sh4's codegen for `-mfdpic`. I apologize for accepting this previously. Since no upstream toolchain/libc has ever used these relocation types, it should be fine to revert the change.

GOT-indirect addressing is required for accessing data symbols under two conditions:

* Preemptible symbols: Traditional GOT requirement.
* Non-preemptible symbols with potential data segment placement: This includes
  + Writable data symbols: This covers both locally declared (`int var;`) and externally declared (`extern int var;`) non-const variables.
  + Potential dynamic initialization: `const A a; extern const int var;`
  + Certain guaranteed constant initialization: `extern constinit const int *const extern_const;`. Constant initialization may also require a relocation, e.g. `constinit const int *const extern_const = &var;`

If the referenced data symbol is non-preemptible and guaranteed to be in the text segment, we can use PC-relative addressing.
However, this scenario is remarkably rare in practice.

```cpp
const int ro_array[] = {1,2}; // text segment
```

sorear:
Revert PR is #427.

PR #396, opened 2023-09-13
Fix TLSDESC -> Initial Exec example.

ishitatsuyuki:
I realized a mistake after trying to get IE relaxation into mold.

We need a PC-relative load here, hence `auipc` not `lui`.


rui314:
LGTM

kito-cheng (approved):

Issue #398, opened 2023-09-25
Deprecate R_RISCV_RVC_LUI?


rui314:
I was looking for a way to test the `R_RISCV_RVC_LUI` relocation for the mold linker, but it looks like there's no way to make the assembler to emit that relocation type. `lui a0, %hi20(foo)` produces a `R_RISCV_HI20` relocation, but `c.lui a0, %hi20(foo)` is rejected as an illegal combination of the instruction and the operand.

`R_RISCV_RVC_LUI` also seems to be an odd relocation, and I can't think of a valid use case for it. In order to use `c.lui`, you need to make sure that the operand address is representable as a signed 17 bit value. But if you know what the value is, you don't need to emit a relocation -- you can just emit that value directly. If you don't know what the value is, you can't safely emit that relocation because you don't know whether the value will fit in 17 bits or not.

The only valid use case would be a code model which limits the program size to 2^16. However, even for that code model, we don't need `R_RISCV_RVC_LUI` because the linker can already relax `lui` to `c.lui` if the operand is small enough.

lld supports that relocation, but it looks like there's no test for that relocation. mold does not support it at all, and we haven't received any user reports about it.

Does anyone know what this relocation was for? Is there any way to make the assembler/compiler to emit that relocation?

kito-cheng:
I guess that's more like an intermediate relocation used for the linker relaxation process, like `R_RISCV_GPREL_[IS]` relative relocation ( https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/205), because I also don't believe that's can be useful to let user use `c.lui` with this relocation directly, so we should consider deprecate and use this relocation as internal intermediate relocation only.

@Nelson1225 @MaskRay @jrtc27 any thought for this?

Nelson1225:
> I guess that's more like an intermediate relocation used for the linker relaxation process, like R_RISCV_GPREL_[IS] relative relocation ( https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/205), because I also don't believe that's can be useful to let user use c.lui with this relocation directly, so we should consider deprecate and use this relocation as internal intermediate relocation only.

Yap, user cannot use this relocation directly in GNU assembler.  We only accept immediate value in c.lui in gas, without any modifiers.  However, it's also really rare to see this relaxation in ld, since we should reserve the maxpagesize and commongaesize for data segment alignment.  The pagesize usually occupied most of the rvc_lui space...

[...]

PR #400, opened 2023-09-27
Remove R_RISCV_RVC_LUI

rui314:
R_RISCV_RVC_LUI is a binutil's internal relocation. We've never had a assembly syntax for the relocation, and there should be no object file that contains this relocation. This relocation has no valid use case.

Closes https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/398


kito-cheng (approved):
It's only used intermediate only, so remove that is good to me, thanks @rui314  :)

MaskRay (approved):

Issue #399, opened 2023-09-26
Define GOT-Relative data relocation


PiJoules:
We would like to introduce a new relocation type that works similarly to x86’s R_X86_64_GOTPCREL relocation.

In the “GOT-relative data relocations” section of [riscv-elf.adoc](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#relocations), we would add a new type “R_RISCV_GOT32_PCREL” that follows the existing wording to “R_RISCV_32_PCREL”, but instead evaluates to the 32-bit offset between a GOT entry for a given symbol and the current location where the relocation is applied, so its equation would be “G + GOT - P + A”.

The purpose of this relocation is to remove what we call “rtti_proxies” in the [relative vtables C++ ABI](https://youtu.be/9HGKlDiJy8E). This ABI saves memory by moving vtables into .rodata by replacing each of the relocations that would constitute the vtable (R_RISCV_64, R_AARCH64_ABS64, R_X86_64_64) with PC-relative offsets, requiring no dynamic relocations. Instead of placing a pointer to each virtual method into the vtable, we would instead take the offset between the vtable and the virtual function which can be statically computed if the compiler knows both symbols are DSO-local. The same is done with the pointer to the RTTI object, but an issue here is the RTTI is not guaranteed to be DSO-local, so we must emit a DSO-local “proxy” symbol which points to the actual RTTI object, and take the offset to the proxy instead. This proxy is equivalent to a GOT entry, but in order to use the GOT entry, we’d need a relocation that points to the entry itself rather than just the start of the GOT.

The proxies work in the meantime, but this new relocation would allow us to effectively “move” those proxies to the GOT which will provide these benefits:
1. It will reuse the same GOT slot across translation units without COMDAT machinery for each proxy, which reduces clutter and bloat in object files and streamlines linking by using simpler features with less bulk in the input files.
2. It will reuse the same GOT slot as a non-vtable reference to the same symbol would.
3. The GOT is tightly packed in the RELRO segment, whereas the proxy symbols will be placed in arbitrary parts of the .data.rel.ro section.

This could also be used for other PIC-friendly optimizations we’d like to implement that aren’t specific to vtables.

We have a [WIP](https://reviews.llvm.org/D132620) patch that implements this for x86_64, but does not yet work for RISC-V. What steps can we take to add this relocation?

rui314:
[...]

PR #402, opened 2023-09-28
Define GOT-Relative data relocation

PiJoules:
This introduces a new relocation `R_RISCV_GOT32_PCREL` that follows the existing wording to “R_RISCV_32_PCREL”, but instead evaluates to the 32-bit offset between a GOT entry for a given symbol and the current location where the relocation is applied, so its equation would be “G + GOT - P + A”.

See https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/399#issuecomment-1831014035 for a code example.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -420,7 +420,9 @@ Description:: Additional information about the relocation
                                             <| V - S - A
 .2+| 40      .2+| SUB64         .2+| Static  | _word64_          .2+| 64-bit label subtraction
                                             <| V - S - A
-.2+| 41-42   .2+| *Reserved*    .2+| -       |                   .2+| Reserved for future standard use
+.2+| 41      .2+| GOT32_PCREL   .2+| Static  | _word32_          .2+| 32-bit difference between the GOT entry for a symbol and the current location
+                                            <| G + GOT + A - P
+.2+| 42      .2+| *Reserved*    .2+| -       |                   .2+| Reserved for future standard use
                                             <|
 .2+| 43      .2+| ALIGN         .2+| Static  |                   .2+| Alignment statement. The addend indicates the number of bytes occupied by `nop` instructions at the relocation offset. The alignment boundary is specified by the addend rounded up to the next power of two.
                                             <|


riscv-elf.adoc
```diff
@@ -467,8 +467,10 @@ Description:: Additional information about the relocation
 .2+| 64      .2+| TLSDESC_ADD_LO12  .2+| Static  | _I-Type_          .2+| Low 12 bits of a 32-bit PC-relative offset into a TLS descriptor entry, `%tlsdesc_add_lo(address of %tlsdesc_hi)`, the addend must be 0
                                             <| S - P
 .2+| 65      .2+| TLSDESC_CALL      .2+| Static  |                   .2+| Annotate call to TLS descriptor resolver function, `%tlsdesc_call(address of %tlsdesc_hi)`, for relaxation purposes only
+.2+| 66      .2+| 32_PCREL      .2+| Static  | _word32_          .2+| 32-bit PC relative offset to the GOT entry for a symbol
+                                            <| G + GOT - P + A
```
rui314, line 471: The order should be `G + GOT + A - P` for the sake of consistency with the `GOT_HI20`'s expression.

riscv-elf.adoc
```diff
@@ -467,8 +467,10 @@ Description:: Additional information about the relocation
 .2+| 64      .2+| TLSDESC_ADD_LO12  .2+| Static  | _I-Type_          .2+| Low 12 bits of a 32-bit PC-relative offset into a TLS descriptor entry, `%tlsdesc_add_lo(address of %tlsdesc_hi)`, the addend must be 0
                                             <| S - P
 .2+| 65      .2+| TLSDESC_CALL      .2+| Static  |                   .2+| Annotate call to TLS descriptor resolver function, `%tlsdesc_call(address of %tlsdesc_hi)`, for relaxation purposes only
+.2+| 66      .2+| 32_PCREL      .2+| Static  | _word32_          .2+| 32-bit PC relative offset to the GOT entry for a symbol
```
rui314, line 470: Relocation number 41 is available. I'd use that number instead of 66.

Also, `32_PCREL` is a typo, right?

rui314:
The format of the table is broken. Check it yourself and please fix the table format. Screenshot 2023-09-29 at 11 14 28

riscv-elf.adoc
```diff
@@ -467,8 +467,10 @@ Description:: Additional information about the relocation
...
+                                            <| G + GOT - P + A
                                             <|
```
kito-cheng, line 472: Seems like need to remove this line.

rui314 (commented):
A question that naturally raises would be whether or not we want a 64-bit variant of this relocation, but I guess we don't need it at this moment because even the usual `GOT_HI20` and `LO12_I` combination can't address the entire 64-bit address space.

What is the assembly syntax for this new relocation?

riscv-elf.adoc
PiJoules: Yup. My b

PiJoules:
I think the formatting should be fixed now

rui314:
s/PC relative/PC-relative/ for the sake of consistency.

Other than that, LGTM

kito-cheng:
LGTM, but just require a PoC implementation either LD.BFD or LLD as convention :)

MaskRay:
I think this extension is fine.

We need a code example for my question on https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/399 . You can edit the description of the first comment to link to #399

riscv-elf.adoc
```diff
@@ -420,7 +420,9 @@ Description:: Additional information about the relocation
                                             <| V - S - A
 .2+| 40      .2+| SUB64         .2+| Static  | _word64_          .2+| 64-bit label subtraction
                                             <| V - S - A
-.2+| 41-42   .2+| *Reserved*    .2+| -       |                   .2+| Reserved for future standard use
+.2+| 41      .2+| GOT32_PCREL   .2+| Static  | _word32_          .2+| 32-bit PC-relative offset to the GOT entry for a symbol
```
MaskRay, line 423: 32-bit GOT entry offset relative to a symbol?

Nelson1225: 32-bit relative offset to GOT entry?  It seems like the two data relocations, 32_PCREL and GOT32_PCREL, have similar purpose, but the former is the offset between pc and symbol, and the later is the offset between pc and GOT entry?

PiJoules: > 32-bit GOT entry offset relative to a symbol?

32-bit GOT entry offset relative to the PC, just as @Nelson1225 mentions. GOT32_PCREL is the offset between PC and the GOT entry and 32_PCREL is the offset between PC and a symbol. Should the wording be updated?

PiJoules:
> I think this extension is fine.
> 
> We need a code example for my question on #399 . You can edit the description of the first comment to link to #399

Done

riscv-elf.adoc
MaskRay: I was actually confused by the AArch64 ABI wording as well.

Perhaps more verbose, 32-bit difference between XXX and YYY.

MaskRay (approved):

riscv-elf.adoc
PiJoules: Done

kito-cheng:
We got LGTM from two linker experts, also PoC for lld https://github.com/llvm/llvm-project/pull/72587

c.c. @Nelson1225 just remind you there are new relocation added.

PR #404, opened 2023-11-12
Specify the alignment and size for array

kito-cheng:
After I reviewing the calling convention part, I found we missed the alignment and size for array, fortunately it's both open source compilers are implement the same, so just document that down.


jrtc27:
> After I reviewing the calling convention part, I found we missed the alignment and size for array

That's because it's specified by C itself?

PR #405, opened 2023-11-12
Add ePIC relocations and rewrite rules

luismarques:
Continuing to follow up on the request to decompose the ePIC PR into multiple PRs, this PR adds the ePIC relocations and rewrite rules.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -503,6 +503,16 @@ Description:: Additional information about the relocation
                                             <| S - P
 .2+| 65      .2+| TLSDESC_CALL      .2+| Static  |                   .2+| Annotate call to TLS descriptor resolver function, `%tlsdesc_call(address of %tlsdesc_hi)`, for relaxation purposes only
                                             <|
+
+.2+| 66      .2+| EPIC_HI20    .2+| Static  | _U-Type_          .2+| <<Embedded PIC rewrite rules,Embedded PIC rewrite>> for the high 20 bits of a 32-bit PC- or GP-relative offset, `epic_hi(symbol)`
+                                            <| Rewrite
+.2+| 67      .2+| EPIC_LO12_I  .2+| Static  | _I-type_          .2+| <<Embedded PIC rewrite rules,Embedded PIC rewrite>> for the low 12 bits of a 32-bit PC- or GP-relative offset,  `epic_low(address of %epic_high)`
+                                            <| Rewrite
+.2+| 68      .2+| EPIC_LO12_S  .2+| Static  | _S-Type_          .2+| <<Embedded PIC rewrite rules,Embedded PIC rewrite>> for the low 12 bits of a 32-bit PC- or GP-relative offset,  `epic_low(address of %epic_high)`
+                                            <| Rewrite
+.2+| 69      .2+| EPIC_BASE_ADD .2+| Static  | _I-type_         .2+| <<Embedded PIC rewrite rules,Embedded PIC rewrite>> of `gp` addition, `%epic_base_add(symbol)`
+                                            <| Rewrite
+
 .2+| 66-191  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
                                             <|
 .2+| 192-255 .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for nonstandard ABI extensions
@@ -848,6 +858,48 @@ def align(addend):
     ALIGN *= 2
   return ALIGN
 ----
+
+==== Embedded PIC rewrite rules
+
+The Embedded PIC (ePIC) relocations allow addressing a symbol relative to either the PC or the GP, as appropriate for that symbol.
+
+ NOTE: Implementations are permitted to specify their own rules for choosing between GP-relative or PC-relative addressing. A possible rule is:
+
+* If the symbol resides in a writable output section, then GP-relative addressing is used;
+* If the symbol does not reside in an output section or the section is not writable, then PC-relative addressing is used instead.
+
+The ePIC relocations are applied to a sequence of instructions that initially address a symbol relative to the GP. When PC-relative addressing should be used instead, the ePIC relocations rewrite the instructions to perform PC-relative addressing and add the appropriate PC-relative relocations. When GP-relative addressing should be used, the instruction rewrites do not occur and GP-relative relocations are added. For correctness, the rewrites must occur even when linker relaxations are disabled. Paired `R_RISCV_RELAX` relocations are preserved during the rewrite process, and will pair with relocations added as part of that rewrite.
+
+===== `R_RISCV_EPIC_HI20` [[rels-epic-hi]]
+
+The `R_RISCV_EPIC_HI20` relocation must apply to an `lui` instruction. Its behavior depends on the residence of the referenced symbol, per the <<Embedded PIC rewrite rules,ePIC rewrite rules>>.
+
+* For PC-relative addressing:
+** Rewrites the `lui` instruction into an `auipc` instruction with the same operands, by overwriting the opcode field.
+** Adds a `R_RISCV_PCREL_HI20` relocation with the same symbol and addend, at the same offset.
+* For GP-relative addressing:
+** Adds a `R_RISCV_GPREL_HI20` relocation with the same symbol and addend, at the same offset.
+
+===== `R_RISCV_EPIC_BASE_ADD` [[rels-epic-base]]
+
+The `R_RISCV_EPIC_BASE_ADD` relocation must apply to an `add` or `c.add` instruction. Its behavior depends on the residence of the referenced symbol, per the <<Embedded PIC rewrite rules,ePIC rewrite rules>>.
+
+* For PC-relative addressing, it either:
+** Rewrites the addition instruction into a canonical `nop` (`addi x0, x0, 0`) or `c.nop` instruction (if the relocation is being applied to an `add` or `c.add` instruction, respectively), or
+** Deletes the `add` or `c.add` instruction.
+* For GP-relative addressing:
+** Nothing needs to be done.
+
+NOTE: If a `R_RISCV_GPREL_ADD` relaxation relocation is defined in the future, the GP-relative addressing rule could be updated to read: "Optionally adds a `R_RISCV_GPREL_ADD` relaxation relocation with the same symbol and addend, at the same offset."
+
+===== `R_RISCV_EPIC_LO12_*` [[rels-epic-lo]]
+
+The `R_RISCV_EPIC_LO12_I` and `R_RISCV_EPIC_LO12_S` relocations apply to instructions encoded using the `I` and `S` instruction formats, respectively. For both of them, the symbol points to an instruction with a `R_RISCV_EPIC_HI20` relocation. Their behavior depends on the residence of the symbol referenced by the respective `R_RISCV_EPIC_HI20` relocation, per the <<Embedded PIC rewrite rules,ePIC rewrite rules>>.
+
+* For PC-relative addressing, it either:
+** Adds a `R_RISCV_PCREL_LO12_I` or `R_RISCV_PCREL_LO12_S` relocation, as appropriate, with the same symbol and addend, at the same offset.
+* For GP-relative addressing:
+** Adds a `R_RISCV_GPREL_LO12_I` or `R_RISCV_GPREL_LO12_S` relocation, as appropriate, at the same offset. The symbol and addend of the new relocation are those of the corresponding `R_RISCV_EPIC_HI20` relocation.
 
 === Thread Local Storage
 


riscv-elf.adoc
```diff
@@ -799,6 +809,44 @@ def align(addend):
...
+
+===== `R_RISCV_EPIC_BASE_ADD` [[rels-epic-base]]
+
+The `R_RISCV_EPIC_BASE_ADD` relocation must apply to an `add` or `c.add` instruction. Its behavior depends on the residence of the referenced symbol, per the <<Embedded PIC rewrite rules,ePIC rewrite rules>>.
+
+* For PC-relative addressing, it either:
+** Rewrites the addition instruction into a canonical `nop` (`addi x0, x0, 0`) or `c.nop` instruction (if the relocation is being applied to an `add` or `c.add` instruction, respectively), or
+** Deletes the `add` or `c.add` instruction.
+* For GP-relative addressing:
+** Nothing needs to be done.
```
luismarques, line 839: In the future, I'm hoping we'll have a `R_RISCV_GPREL_ADD` relocation, for relaxation purposes. If that happens, we can replace this with: "Optionally adds a `R_RISCV_GPREL_ADD` relaxation relocation with the same symbol and addend, at the same offset."

riscv-elf.adoc
```diff
@@ -799,6 +809,44 @@ def align(addend):
   return ALIGN
 ----
 
+==== Embedded PIC rewrite rules
+
+The Embedded PIC (ePIC) relocations allow addressing a symbol relative to either the PC or the GP, as appropriate for that symbol:
```
kito-cheng, line 814: Could you mention rewrite process will perform even linker relaxation is disabled, also mention paired R_RISCV_RELAX will keep after rewrite?

riscv-elf.adoc
```diff
@@ -799,6 +809,44 @@ def align(addend):
...
+
+* If the symbol resides in a writable output section, then GP-relative addressing is used;
+* If the symbol does not reside in an output section or the section is not writable, then PC-relative addressing is used instead.
```
kito-cheng, line 817: Does it possible to use the GP-relative addressing reachable or PC-relative addressing reachable to determine which target relocation should use?

I think that could also address the scenario of RO data is close to RW data rather than close to text sections?

luismarques: > Does it possible to use the GP-relative addressing reachable or PC-relative addressing reachable to determine which target relocation should use?

How would that work when the (PC-relative) code/text/rodata segment can be relocated independently of the (GP-relative) data segment?

> I think that could also address the scenario of RO data close to RW data rather than close to text sections?

Those distances won't be static once the segments are relocated independently. Is your concern that you won't be able to address something at all or just that it will require a longer instruction sequence compared with some other approach? Don't forget that once the PC vs GP-relative addressing is resolved by the rewrite rules, you can still apply relaxations.

kito-cheng: > How would that work when the (PC-relative) code/text/rodata segment can be relocated independently of the (GP-relative) data segment?

Oh, I guess I didn't aware they may relocated independently, thanks for sharing this scenario.

> Those distances won't be static once the segments are relocated independently. Is your concern that you won't be able to address something at all or just that it will require a longer instruction sequence compared with some other approach? Don't forget that once the PC vs GP-relative addressing is resolved by the rewrite rules, you can still apply relaxations.

In my past experience there may have two case for the rodata layout, one is put close to data, and another one  is put close to the text section.

So if we determine by the writable attribute of section, that means the rodata can only place together with text section via PC-relative addressing.

My thought is make rewrite rule flexible - one of possible is list both rewrite rule here and mention toolchain may offer some option to select, the intention of this proposal is I would like to make it cover more case if possible, and for the PoC part, I think it's OK to implement one rule only.

kito-cheng: How about adding one more note? I am OK to moving this forward with this note if you also think it's appropriate.
`NOTE: Implementations are permitted to apply alternative rules for choosing between GP-relative or PC-relative addressing.`

luismarques: > How about adding one more note? I am OK to moving this forward with this note if you also think it's appropriate.
> `NOTE: Implementations are permitted to apply alternative rules for choosing between GP-relative or PC-relative addressing.`

I agree. I have rebased the PR and added the note you suggested. Thanks!

kito-cheng (approved):
LGTM :)

riscv-elf.adoc
```diff
@@ -849,6 +859,46 @@ def align(addend):
...
+
+NOTE: Implementations are permitted to apply alternative rules for choosing between GP-relative or PC-relative addressing.
```
jrtc27, line 869: Notes are non-normative, so can't override what's outside a note. If it's not required then the spec above needs to not dictate the rules.

riscv-elf.adoc
jrtc27: Put this as forward-looking non-normative next?

riscv-elf.adoc
luismarques: I've put the old description as part of the non-normative note. Please check if that addresses your concern.

riscv-elf.adoc
luismarques: Please see if the added note properly addresses your suggestion.

MaskRay:
I am happy with the GP-relative relocation part (also used by the FDPIC ABI proposal), but I am afraid I am concerned of `R_RISCV_EPIC_*` relocations. Commented here https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/343#issuecomment-1926279418

riscv-elf.adoc
```diff
@@ -849,6 +859,48 @@ def align(addend):
...
+==== Embedded PIC rewrite rules
+
+The Embedded PIC (ePIC) relocations allow addressing a symbol relative to either the PC or the GP, as appropriate for that symbol.
+
+ NOTE: Implementations are permitted to specify their own rules for choosing between GP-relative or PC-relative addressing. A possible rule is:
+
+* If the symbol resides in a writable output section, then GP-relative addressing is used;
+* If the symbol does not reside in an output section or the section is not writable, then PC-relative addressing is used instead.
+
+The ePIC relocations are applied to a sequence of instructions that initially address a symbol relative to the GP. When PC-relative addressing should be used instead, the ePIC relocations rewrite the instructions to perform PC-relative addressing and add the appropriate PC-relative relocations. When GP-relative addressing should be used, the instruction rewrites do not occur and GP-relative relocations are added. For correctness, the rewrites must occur even when linker relaxations are disabled. Paired `R_RISCV_RELAX` relocations are preserved during the rewrite process, and will pair with relocations added as part of that rewrite.
```
sorear, line 871:
```suggestion
The ePIC relocations are applied to a sequence of instructions that initially address a symbol relative to the GP. When PC-relative addressing should be used instead, the ePIC relocations rewrite the instructions to perform PC-relative addressing. When GP-relative addressing should be used, the instruction rewrites do not occur. For correctness, the rewrites must occur even when linker relaxations are disabled.  Relaxation may occur if paired `R_RISCV_RELAX` relocations are present.
```

The behavior of relocations needs to be specified in terms of **observable** linker behavior.  The output will not contain non-dynamic relocations (and ePIC does not support dynamic linking), so describe it purely in terms of instruction rewrites.

riscv-elf.adoc
```diff
@@ -849,6 +859,48 @@ def align(addend):
   return ALIGN
 ----
 
+==== Embedded PIC rewrite rules
+
+The Embedded PIC (ePIC) relocations allow addressing a symbol relative to either the PC or the GP, as appropriate for that symbol.
```
sorear, line 864: We need to allow absolute addressing as a third option (for weak symbols, and also usable for symbols not in an output section).

riscv-elf.adoc
```diff
@@ -849,6 +859,48 @@ def align(addend):
...
+
+===== `R_RISCV_EPIC_HI20` [[rels-epic-hi]]
+
+The `R_RISCV_EPIC_HI20` relocation must apply to an `lui` instruction. Its behavior depends on the residence of the referenced symbol, per the <<Embedded PIC rewrite rules,ePIC rewrite rules>>.
+
+* For PC-relative addressing:
+** Rewrites the `lui` instruction into an `auipc` instruction with the same operands, by overwriting the opcode field.
+** Adds a `R_RISCV_PCREL_HI20` relocation with the same symbol and addend, at the same offset.
+* For GP-relative addressing:
+** Adds a `R_RISCV_GPREL_HI20` relocation with the same symbol and addend, at the same offset.
```
sorear, line 881: Use observable behavior.

Relaxation may delete the `HI20` entirely, unless that is documented elsewhere.

riscv-elf.adoc
```diff
@@ -849,6 +859,48 @@ def align(addend):
...
+
+The `R_RISCV_EPIC_BASE_ADD` relocation must apply to an `add` or `c.add` instruction. Its behavior depends on the residence of the referenced symbol, per the <<Embedded PIC rewrite rules,ePIC rewrite rules>>.
+
+* For PC-relative addressing, it either:
+** Rewrites the addition instruction into a canonical `nop` (`addi x0, x0, 0`) or `c.nop` instruction (if the relocation is being applied to an `add` or `c.add` instruction, respectively), or
+** Deletes the `add` or `c.add` instruction.
+* For GP-relative addressing:
+** Nothing needs to be done.
+
+NOTE: If a `R_RISCV_GPREL_ADD` relaxation relocation is defined in the future, the GP-relative addressing rule could be updated to read: "Optionally adds a `R_RISCV_GPREL_ADD` relaxation relocation with the same symbol and addend, at the same offset."
```
sorear, line 893: Delete, not observable.

riscv-elf.adoc
```diff
@@ -849,6 +859,48 @@ def align(addend):
...
+
+===== `R_RISCV_EPIC_BASE_ADD` [[rels-epic-base]]
+
+The `R_RISCV_EPIC_BASE_ADD` relocation must apply to an `add` or `c.add` instruction. Its behavior depends on the residence of the referenced symbol, per the <<Embedded PIC rewrite rules,ePIC rewrite rules>>.
+
+* For PC-relative addressing, it either:
+** Rewrites the addition instruction into a canonical `nop` (`addi x0, x0, 0`) or `c.nop` instruction (if the relocation is being applied to an `add` or `c.add` instruction, respectively), or
+** Deletes the `add` or `c.add` instruction.
+* For GP-relative addressing:
+** Nothing needs to be done.
```
sorear, line 891: Only allow the instruction to be deleted if there is a paired `R_RISCV_RELAX`.

Allow deletion if there is a paired `R_RISCV_RELAX` and the `HI20` was deleted by relaxation.

riscv-elf.adoc
```diff
@@ -503,6 +503,16 @@ Description:: Additional information about the relocation
...
                                             <|
+
+.2+| 66      .2+| EPIC_HI20    .2+| Static  | _U-Type_          .2+| <<Embedded PIC rewrite rules,Embedded PIC rewrite>> for the high 20 bits of a 32-bit PC- or GP-relative offset, `epic_hi(symbol)`
+                                            <| Rewrite
+.2+| 67      .2+| EPIC_LO12_I  .2+| Static  | _I-type_          .2+| <<Embedded PIC rewrite rules,Embedded PIC rewrite>> for the low 12 bits of a 32-bit PC- or GP-relative offset,  `epic_low(address of %epic_high)`
+                                            <| Rewrite
+.2+| 68      .2+| EPIC_LO12_S  .2+| Static  | _S-Type_          .2+| <<Embedded PIC rewrite rules,Embedded PIC rewrite>> for the low 12 bits of a 32-bit PC- or GP-relative offset,  `epic_low(address of %epic_high)`
+                                            <| Rewrite
+.2+| 69      .2+| EPIC_BASE_ADD .2+| Static  | _I-type_         .2+| <<Embedded PIC rewrite rules,Embedded PIC rewrite>> of `gp` addition, `%epic_base_add(symbol)`
```
sorear, line 513: I would prefer to have this point at the `%epic_high` for consistency.

riscv-elf.adoc
```diff
@@ -849,6 +859,48 @@ def align(addend):
...
+NOTE: If a `R_RISCV_GPREL_ADD` relaxation relocation is defined in the future, the GP-relative addressing rule could be updated to read: "Optionally adds a `R_RISCV_GPREL_ADD` relaxation relocation with the same symbol and addend, at the same offset."
+
+===== `R_RISCV_EPIC_LO12_*` [[rels-epic-lo]]
+
+The `R_RISCV_EPIC_LO12_I` and `R_RISCV_EPIC_LO12_S` relocations apply to instructions encoded using the `I` and `S` instruction formats, respectively. For both of them, the symbol points to an instruction with a `R_RISCV_EPIC_HI20` relocation. Their behavior depends on the residence of the symbol referenced by the respective `R_RISCV_EPIC_HI20` relocation, per the <<Embedded PIC rewrite rules,ePIC rewrite rules>>.
+
+* For PC-relative addressing, it either:
+** Adds a `R_RISCV_PCREL_LO12_I` or `R_RISCV_PCREL_LO12_S` relocation, as appropriate, with the same symbol and addend, at the same offset.
+* For GP-relative addressing:
+** Adds a `R_RISCV_GPREL_LO12_I` or `R_RISCV_GPREL_LO12_S` relocation, as appropriate, at the same offset. The symbol and addend of the new relocation are those of the corresponding `R_RISCV_EPIC_HI20` relocation.
```
sorear, line 902: Describe observable behavior - it relocates and relaxes as if there were an XXX relocation, it does not create an XXX relocation.

sorear (changes requested):
I don't have a good understanding of the use cases for embedded systems that need PIC but not shared libraries.  Is there any freely source available example of something that would use this?  If it's needed for a commercial RTOS, is there at least a public manual or datasheet?

@MaskRay Was your comment questioning the existence of the use case for ePIC without shared libraries, suggesting that ePIC should be a special case of FDPIC, or something else?

Assuming that the use case is valid, the general approach looks good.  Relocation behavior needs to be described in the language of externally observable linker behavior, not implementation details like synthesized non-dynamic relocations.  I think we also need an absolute option, at least for weak symbols but it may be useful more generally.

MaskRay:
> I don't have a good understanding of the use cases for embedded systems that need PIC but not shared libraries. Is there any freely source available example of something that would use this? If it's needed for a commercial RTOS, is there at least a public manual or datasheet?
> 
> @MaskRay Was your comment questioning the existence of the use case for ePIC without shared libraries, suggesting that ePIC should be a special case of FDPIC, or something else?
> 
> Assuming that the use case is valid, the general approach looks good. Relocation behavior needs to be described in the language of externally observable linker behavior, not implementation details like synthesized non-dynamic relocations. I think we also need an absolute option, at least for weak symbols but it may be useful more generally.

My comment is questioning the existence of the use case for ePIC without shared libraries. I think we should consider FDPIC, which works with MMU-less and MMU systems, supports dynamic libraries. After FDPIC support, we can figure out whether there is still ePIC need. (I kinda suspect no.)

riscv-elf.adoc
kito-cheng: +1 for this change

PR #408, opened 2023-11-22
Clarify that multiple R_RISCV_LO12_I/R_RISCV_LO12_S can share one single R_RISCV_HI20

MaskRay:
This is most common on rv32 when loading a 8-byte aligned global variable, e.g.
```
double NaturallyAlignedScalar = 5.;
double accessNaturallyAlignedScalar() { return NaturallyAlignedScalar; }
//      lui     a1, %hi(NaturallyAlignedScalar)
//      lw      a0, %lo(NaturallyAlignedScalar)(a1)
//      lw      a1, %lo(NaturallyAlignedScalar+4)(a1)

double NaturallyAlignedArray[4] = { 3., 4., 5., 6. };
double accessNaturallyAlignedArray() {
  return NaturallyAlignedArray[0] + NaturallyAlignedArray[3];
}
//      lui     a2, %hi(NaturallyAlignedArray)
//      lw      a0, %lo(NaturallyAlignedArray)(a2)
//      lw      a1, %lo(NaturallyAlignedArray+4)(a2)
//      addi    a3, a2, %lo(NaturallyAlignedArray)
//      lw      a2, 24(a3)
//      lw      a3, 28(a3)
```

The HI20 values for the multiple fragments must be identical and all the relaxed global-pointer offsets must be in range.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -581,6 +581,20 @@ The following assembly and relocations show loading an absolute address:
     addi a0, a0, %lo(symbol) # R_RISCV_LO12_I (symbol)
 ----
 
+A symbol can be loaded in multiple fragments using different addends, where
+multiple instructions associated with `R_RISCV_LO12_I`/`R_RISCV_LO12_S` share a
+single `R_RISCV_HI20`. The HI20 values for the multiple fragments must be
+identical.
+
+[,asm]
+----
+    lui a0, 0       # R_RISCV_HI20 (symbol)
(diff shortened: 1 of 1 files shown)


rui314:
The examples in this change raise the question of whether the example code is actually valid, and to address that question, we need to mention the alignment issue and its resolution, as pointed out by @kito-cheng. This discussion seems a bit derailed from the primary purpose of the psABI. Maybe just saying that there may be more than one R_RISCV_LO12_I/R_RISCV_LO12_S for a R_RISCV_HI20 is enough?

PR #414, opened 2023-12-18
Add a tag to indicate whether to work around this binutils ADD/SUB/SE…

palmer-dabbelt:
…T bug

See <https://sourceware.org/bugzilla/show_bug.cgi?id=31179> for slightly more info.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1155,6 +1155,7 @@ non-standard ABI extensions.
 | Tag_RISCV_priv_spec_revision        |       12 | uleb128        | *Deprecated*, indicates the revision version of the privileged specification.
 | Tag_RISCV_atomic_abi                |       14 | uleb128        | Indicates which version of the atomics ABI is being used.
 | Tag_RISCV_x3_reg_usage              |       16 | uleb128        | Indicates the usage definition of the X3 register.
+| TAG_RISCV_addsubset_addend_bug      |       17 | byte           | Indicates whether the object suffers from the ADD/SUB/SET addend duplication bug (see binutils PR31179).
 | Reserved for non-standard attribute | >= 32768 | -              | -
 |===
 


MaskRay:
This `R_RISCV_SUB*` addend bug (fixed now) is certainly unfortunate but it is not severe. It affects `.reloc` directives using `R_RISCV_SUB*`, which are extremely uncommon (never encountered if I did not write assembler/linker tests).

If an object is subject to this gas bug, what can the toolchain do?
I believe it's a non-starter for other toolchain components (LLVM) to detect the bug.
We cannot retrospectively add a tag to the relocatable file with the bug.
It's also not feasible to force new relocatable files without the bug to add a tag.

The `R_RISCV_SUB_ULEB128` bug "RISC-V: Clarify the behaviors of SET/ADD/SUB relocations." fixed is worse as it affects `.uleb128 (A + 3) - (B + 2)` (not just `.reloc`).
However, such subtraction expressions are relatively uncommon as well and ULEB128 relocations are very new.

Personally I think we should just treat the gas bug as a defect. Usually we'd not run into a problem.
If we use advanced assembly tricks, we should be aware of the binutils<2.41 problem.

Nelson1225:
> This R_RISCV_SUB* addend bug (fixed now) is certainly unfortunate but it is not severe. It affects .reloc directives using R_RISCV_SUB*, which are extremely uncommon (never encountered if I did not write assembler/linker tests).

If this is rare, then I think I can have a workaround/hack in GNU assembler to make .reloc ADD/SET/SUB/ULEB128 friendly with the 2.41 linker behavior.

rui314:
I'm not sure if making a modification to the spec is the best way to handle the situation, but this proposal, as it stands, is incomplete because it does not specify anything other than defining a new tag.

What should the linker do if the tag is present in an object file? I think the spec needs to address that question.

kito-cheng:
Has short chart with Nelson about the bug and this tag, and we both don't think the tag resolve the problem unless all other toolchain also always emit that tag as well.

Otherwise object file produced by LLVM (and all other toolchain) will handle incorrectly with newer binutils, because that object has no tag, and then it will use legacy way to handle, which is wrong...

kito-cheng:
More detail here:

Use this uleb128 directive as example:
```
.uleb128 bar - foo + 1
``` 

The assembler should emit `R_RISCV_SET_ULEB128 bar + 1` and `R_RISCV_SUB_ULEB128 foo + 0`, but binutils has emit `R_RISCV_SET_ULEB128 bar + 1` and `R_RISCV_SUB_ULEB128 foo + 1`, it will become `(bar + 1) - (foo + 1)` which is `bar - foo`, but why it still work? because ld.bfd will ignore the `Addend` of R_RISCV_SUB_ULEB128.

So it will become problem once new binutils (2.42, not release yet) link the object files which produced by binutils (2.41), because that will evaluate the `(bar + 1) - (foo + 1)` right in above example.

And `lld` is implement that right, so it will got problem when it link object came from bintuils 2.41.

Nelson1225:
> because ld.bfd will ignore the Addend of R_RISCV_SUB_ULEB128
Should be ignore the addend of SET_ULEB128, so need this patch to commit in binutils 2.42, https://sourceware.org/pipermail/binutils/2023-December/131356.html.

jrtc27:
We can’t go adding tags to document historic bugs. If binutils is broken then that’s pretty sad and I have questions about the testing methodology in use, but that’s for the binutils maintainers to deal with like any other bug.

kito-cheng:
Discussed in last psABI meeting, and we reach consensus to close this PR since we believe this tag isn't help that issue too much.

PR #415, opened 2023-12-19
Clarify ULEB128 note

rui314:
Previously, it is not very clear what the linker does for `R_RISCV_{SET,SUB}_ULEB128` relocations. Based on the discussion at https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/413, the intended specification appears to be as follows:

 - the assembler allocates enough space for the final computed value for the `R_RISCV_{SET,SUB}_ULEB128` relocation pair and fills the space with a ULEB128 value, which may be redundantly encoded
 - the linker reads an existing ULEB128 value and overwrites it, without changing the length of the existing ULEB128 value

This change serves to clarify the above points.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -484,10 +484,11 @@ assembler syntax, including that of the relocations, is described in the
 _RISC-V Assembly Programmer's Manual_ <<rv-asm>>.
 
 [[uleb128-note]]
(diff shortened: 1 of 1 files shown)


kito-cheng (approved):
Thanks, it's much clear for how to padding :)

PR #416, opened 2023-12-22
Specify the Calling Convention for Fixed-Length Vectors

kito-cheng:
Previously, there was no mention of fixed-length vectors in the psABI, however fortunately, GCC and Clang are implemented same ABI rule for fixed-length vectors, so this PR is document the de factor behavior for that.

- Fixed-length vector are treat as aggregates.
- The aligment requirement is equivalent to the alignment requirement of its elemental type.

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -201,6 +201,8 @@ Aggregates larger than 2×XLEN bits are passed by reference and are replaced in
 the argument list with the address, as are {Cpp} aggregates with nontrivial copy
 constructors, destructors, or vtables.
 
+Fixed-length vectors are treated as aggregates.
+
 Empty structs or union arguments or return values are ignored by C compilers
 which support them as a non-standard extension.  This is not the case for {Cpp},
 which requires them to be sized types.
@@ -594,6 +596,18 @@ The alignment of `max_align_t` is 16.
 
 Structs and unions are aligned to the alignment of their most strictly aligned
 member. The size of any object is a multiple of its alignment.
+
+=== Fixed-length vector
+
+Various compilers have support for fixed-length vector types, for example GCC
+and Clang both support declaring a type with `__attribute__((vector_size(N))`,
+where N is a positive number larger than zero.
+
+The alignment requirement for the fixed length vector shall be equivalent to the
+alignment requirement of its elemental type.
+
+The size of the fixed length vector is determined by multiplying the size of its
+elemental type by the total number of elements within the vector.
 
 === C/{Cpp} type representations
 


kito-cheng:
cc. @palmer-dabbelt @JeffreyALaw @preames @topperc @rofirrim @lhtin

lhtin:
> - Fixed-length vector are treat as aggregates.

This proposal is great and avoids complicating the passing of fixed-length vectors. But the problem is that it doesn't take advantage of vector registers. Are there any more details about this issue?

> GCC and Clang are implemented same ABI rule for fixed-length vectors

Also I tried, GCC currently passes through reference even if the vector size is less than 2*XLEN, which is not according to this proposal. Here is the example: https://godbolt.org/z/3ooovcz7c

kito-cheng:
> Also I tried, GCC currently passes through reference even if the vector size is less than 2*XLEN, which is not according to this proposal. Here is the example: https://godbolt.org/z/3ooovcz7c

That's bug, try to drop `v` from the `-march` and you will see different way to pass the arguments

kito-cheng:
> This proposal is great and avoids complicating the passing of fixed-length vectors. But the problem is that it doesn't take advantage of vector registers. Are there any more details about this issue?

We need separated calling convention for that case, because not every system has `v`, and here is the working draft for that one:

https://github.com/riscv-non-isa/riscv-elf-psabi-doc/tree/fixed-length-vector-cc

riscv-cc.adoc
```diff
@@ -493,6 +495,18 @@ The alignment of `max_align_t` is 16.
 Structs and unions are aligned to the alignment of their most strictly aligned
 member. The size of any object is a multiple of its alignment.
 
+=== Fixed-length vector
+
+Various compilers has support fixed-length vector, one of example is GCC and
```
asb, line 500: "Various compilers have support for fixed-length vector types, for example GCC and Clang both support declaring a type with `__attribute__((vector_size(N))`, where `N` is a positive number larger than zero."

riscv-cc.adoc
```diff
@@ -493,6 +495,18 @@ The alignment of `max_align_t` is 16.
...
+clang both support declare type with `__attribute__((vector_size(N))`, where `N`
+is a postive number larger than zero.
+
+The alignment requirement for the fixed length vector shall be equivalent to the
+alignment requirement of its elemental type.
+
+The size of the is determined by multiplying the size of its
```
asb, line 507: The size of the is => The size of the fixed length vector

asb (approved):
LGTM, modulo my suggested wording tweaks/fixes. Thanks for filling in this gap Kito.

lhtin (approved):
LGTM

sorear (approved):
No ABI break for systems with mixed V and non-V packages, which is my minimum requirement.  Passing small vectors on integer registers is (a) consistent with floats in the non-float ABI, (b) probably will be the optimal approach for the P extension when it exists, (c) benefits in simplicity, so I think I'm fine with it.

PR #417, opened 2024-01-04
Add program property and PLT for CFI extension

kito-cheng:
We introduce .note.gnu.property section to store infomations that linker or runtime system may use, and we define two bit for landing pad and shadow stack.

Those two bit will checked by loader - and use that info to enable the landing pad checking and/or shadow stack feature.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -762,7 +762,7 @@ The first entry in the PLT occupies two 16 byte entries for the default PLT styl
 [,asm]
 ----
 1:  auipc  t2, %pcrel_hi(.got.plt)
-    sub    t1, t1, t3               # shifted .got.plt offset + hdr size + 12 
+    sub    t1, t1, t3               # shifted .got.plt offset + hdr size + 12
     l[w|d] t3, %pcrel_lo(1b)(t2)    # _dl_runtime_resolve
     addi   t1, t1, -(hdr size + 12) # shifted .got.plt offset
     addi   t0, t2, %pcrel_lo(1b)    # &.got.plt
@@ -771,7 +771,7 @@ The first entry in the PLT occupies two 16 byte entries for the default PLT styl
     jr     t3
 ----
 
-And occupies two 16 byte entries for the unlabeled landing pad PLT style:
+And occupies two 16 byte entries for the landing pad PLT style:
 
 [,asm]
 ----
@@ -1525,7 +1525,7 @@ Description:: Additional information about the program property type.
 each bit indicating a feature that the object file is compatible with. The
 linker should perform a bitwise AND operation when merging different objects.
 
-The value of `GNU_PROPERTY_RISCV_FEATURE_1_AND` is defined in <<rv-prog-prop-type>>.
+The `pr_type` value of `GNU_PROPERTY_RISCV_FEATURE_1_AND` is `0xc0000000`.
 
 [%autowidth]
 |===
@@ -1538,7 +1538,7 @@ The value of `GNU_PROPERTY_RISCV_FEATURE_1_AND` is defined in <<rv-prog-prop-typ
 executable sections are built to be compatible with the landing pad mechanism
 provided by the Zicfilp extension in the unlabeled scheme: Executables and
 shared libraries with this bit set are required to generate PLTs in the
-unlabeled landing pad PLT style, and all of the labels of lpad instructions are
+landing pad PLT style, and all of the labels of lpad instructions are
 set to 0, i.e. unlabeled.
 See <<Tool Requirements for Generating Landing Pad Instruction>> for more details.
 
@@ -1561,8 +1561,7 @@ instruction with a compatible label value. Indirect branches may originate from:
 
 * Static linkers when generating PLT entries.
 
-* Other executables and shared libraries via calls through PLT entries or
-  function pointers.
+* Other executables and shared libraries via calls through function pointers.
 
 It is desirable to minimize the number of landing pad instructions to limit the
 number of indirect branch destinations within the program. The following tool
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -777,6 +777,7 @@ And occupies three 16 byte entries for the simple landing pad PLT style:
     jr     t3
     nop
     nop
+    nop
 ----
 
 Subsequent function entry stubs in the PLT take up 16 bytes.
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1484,7 +1484,7 @@ a different features.
 sections are built to be compatible with the landing pad mechanism provided by
 the `Zicfilp` extension. An executable or shared library with this bit set is
 required to generate PLTs with the landing pad (`lpad`) instruction, and all
-label are set to `1`.
+label are set to `0`.
 
 `GNU_PROPERTY_RISCV_FEATURE_1_CFI_SS`: This bit indicate that all executable
 sections are built to be compatible with the shadow stack mechanism provided by
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1531,6 +1531,7 @@ provided by the Zicfilp extension in the unlabeled scheme: Executables and
 shared libraries with this bit set are required to generate PLTs in the
 unlabeled landing pad PLT style, and all of the labels of lpad instructions are
 set to 0, i.e. unlabeled.
+See <<Tool Requirements for Generating Landing Pad Instruction>> for more details.
 
 `GNU_PROPERTY_RISCV_FEATURE_1_CFI_SS`: This bit indicates that all executable
 sections are built to be compatible with the shadow stack mechanism provided by
@@ -1539,6 +1540,39 @@ set requires the execution environment to provide either the `Zicfiss` extension
 or the `Zimop` extension. When the executable or shared library is compiled with
 compressed instructions then loading it with this bit set requires the execution
 environment to provide the `Zicfiss` extension or the `Zcmop` extensions.
+
+=== Tool Requirements for Generating Landing Pad Instructions
+
+For an executable or shared library to set the GNU property
+`GNU_PROPERTY_RISCV_FEATURE_1_CFI_LP_UNLABELED`, every indirect branch targeting
+a location guarded by landing pads must point to a landing pad (`lpad`)
+instruction with a compatible label value. Indirect branches may originate from:
+
+* Relocatable object producers, such as a compiler or assembler.
+
+* Static linkers when generating PLT entries.
+
+* Other executables and shared libraries via calls through PLT entries or
+  function pointers.
+
+It is desirable to minimize the number of landing pad instructions to limit the
+number of indirect branch destinations within the program. The following tool
+requirements determine which tool is responsible for inserting landing pad
+instructions, allowing tools to omit landing pad instructions if it can be
+proven that there are no indirect branches targeting the location.
+
+A relocatable object producer is required to insert a landing pad instruction at
+the target of an indirect branch originating from within the same relocatable
+object.
+
+A relocatable object producer must insert a landing pad instruction at any
+location whose address escapes the relocatable object. This includes all symbols
+that can be exported into the dynamic symbol table by a static linker.
+
+A static linker is required to generate PLT entries with appropriate landing pad
+instructions. During linker relaxation, a static linker may remove unnecessary
+landing pad instructions if it can prove that there are no indirect branches
+targeting the corresponding location.
 
 === Mapping Symbol
 
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -727,9 +727,9 @@ The PLT (Procedure Linkage Table) exists to allow function calls between
 dynamically linked shared objects. Each dynamic object has its own
 GOT (Global Offset Table) and PLT (Procedure Linkage Table).
 
-RISC-V has defined several PLT styles, which used for different situation,
-the default PLT sytle should be used if the program is not met the condition for
-using all other PLT sytle.
+RISC-V defines several PLT styles, which are used in different situations.
+The default PLT style should be used if the program does not meet the conditions
+for using all other PLT sytles.
 
 [[plt-style]]
 .PLT styles
@@ -763,11 +763,11 @@ The first entry in the PLT occupies two 16 byte entries for the default PLT styl
     jr     t3
 ----
 
-And occupies three 16 byte entries for the simple landing pad PLT style:
+And occupies three 16 byte entries for the unlabeled landing pad PLT style:
 [,asm]
 ----
-1:  lpad 0
-    auipc  t2, %pcrel_hi(.got.plt)
+    lpad 0
+1:  auipc  t2, %pcrel_hi(.got.plt)
     sub    t1, t1, t3               # shifted .got.plt offset + hdr size + 16
     l[w|d] t3, %pcrel_lo(1b)(t2)    # _dl_runtime_resolve
     addi   t1, t1, -(hdr size + 16) # shifted .got.plt offset
@@ -794,11 +794,11 @@ The code sequences of the PLT entry for the default PLT style:
     nop
 ----
 
-The code sequences of the PLT entry for the the simple landing pad PLT style:
+The code sequences of the PLT entry for the unlabeled landing pad PLT style:
 [,asm]
 ----
-1:  lpad 0
-    auipc   t3, %pcrel_hi(function@.got.plt)
+    lpad 0
+1:  auipc   t3, %pcrel_hi(function@.got.plt)
     l[w|d]  t3, %pcrel_lo(1b)(t3)
     jalr    t1, t3
 ----
@@ -1492,9 +1492,10 @@ meaning of each column is given below:
 
 Name:: The name of the program property type, omitting the prefix of `GNU_PROPERTY_RISCV_`.
 
-Value:: The type value for the program property type.
+Value:: The `pr_type` value for the program property type.
 
-Size:: The data type size hold within this program property type.
+Size:: The size (`pr_datasz`) of data type held within this program property
+       type.
 
 Description:: Additional information about the program property type.
 
@@ -1511,7 +1512,6 @@ Description:: Additional information about the program property type.
 
 ==== GNU_PROPERTY_RISCV_FEATURE_1_AND
 
-
 `GNU_PROPERTY_RISCV_FEATURE_1_AND` describes a set of features, where each bit
 represents a different feature. The linker should perform a bitwise AND
 operation when merging different objects.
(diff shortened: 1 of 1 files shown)


kito-cheng:
cc @ved-rivos

riscv-elf.adoc
```diff
@@ -1322,6 +1325,43 @@ value. After the merge, the resulting value will be the non-zero one.
...
+[%autowidth]
+|===
+| Bit | Bit Name
+|   0 | GNU_PROPERTY_RISCV_FEATURE_1_ZICFILP
+|   1 | GNU_PROPERTY_RISCV_FEATURE_1_ZICFISS
+|===
+
+`GNU_PROPERTY_RISCV_FEATURE_1_ZICFILP` This bit indicate the program has built
+with landing pad (`Zicfilp` extension), executable or shared libary tag with
+this bit should generate PLT with landing pad.
```
ved-rivos, line 1358:
```suggestion
`GNU_PROPERTY_RISCV_FEATURE_1_ZICFILP` This bit indicate that all executable sections are built with the landing pad mechanism provided by the `Zicfilp` extension. An executable or shared library with this bit set is required to generate PLTs with the landing pad (`lpad`) instruction.
```

riscv-elf.adoc
```diff
@@ -1322,6 +1325,43 @@ value. After the merge, the resulting value will be the non-zero one.
...
+
+`GNU_PROPERTY_RISCV_FEATURE_1_ZICFISS`: This bit indicate the program has built
+with shadow stack (`Zicfiss` extension), executable or shared libary tag with
+this bit will require execution environment at least has implement the
+instructions that defined in `Zimop` extension.
```
ved-rivos, line 1363:
```suggestion
`GNU_PROPERTY_RISCV_FEATURE_1_ZICFISS`: This bit indicate that all executable sections are built to be compatible with the shadow stack mechanism provided by the `Zicfiss` extension. Loading an executable or shared library with this bit set requires the execution environment to provide either the `Zicfiss` extension or the `Zimop` extension. When the executable or shared library is compiled with compressed instructions then loading an executable with this bit set requires the execution environment to provide the `Zicfiss` extension or to provide both the `Zcmop` and `Zimop` extensions.
```

riscv-elf.adoc
ved-rivos: Should we define two flags - one for single-label and one for multi-label. Binaries compiled in single-label mode cannot be linked to binaries compiled with multi-label mode and vice-versa.

riscv-elf.adoc
```diff
@@ -1322,6 +1325,43 @@ value. After the merge, the resulting value will be the non-zero one.
...
+[[rv-seg-type]]
+.RISC-V-specific segment types
+[cols="3,2,3"]
+[width=80%]
+|===
+| Name                              | Value      | Size    | Meaning
+
+| GNU_PROPERTY_RISCV_FEATURE_1_AND  | 0xc0000000 | 4-bytes | RISC-V processor-specific features used in program.
+|===
+
```
ved-rivos, line 1342: Do we need to add a note that "AND" implies that when merging executables, the AND of the feature flags from each of the executable is used to produce the property for the merged executable? Or is that already specified elsewhere?

rui314:
My concern about this proposal is that, with the 32-bit bitmap, we can't have more than 32 extensions. That might be OK for x86-64 that is controlled only by Intel and AMD, but it is likely that we would have way more extensions than x86-64 for RISC-V. I'm not sure even 64 bits is enough. Maybe we should avoid the bitmap completely?

kito-cheng:
@rui314 
GNU property is kind of key-value structure, key is 32 bit integer, and we have 536,870,912 can use for the key per the definition of GNU_PROPERTY_LOPROC and GNU_PROPERTY_HIPROC.

```c
/* Processor-specific semantics, lo */
#define GNU_PROPERTY_LOPROC                     0xc0000000
/* Processor-specific semantics, hi */
#define GNU_PROPERTY_HIPROC                     0xdfffffff

```

So we have  536,870,912  * 32 rather than only 32-bit bitmap :)

rui314:
@kito-cheng Ah, you are right. Even though I've implemented the feature to the linker, I forgot the actual on-file format. Thank you for pointing that out!

MaskRay:
Thanks for the proposal. This looks good for x86/Arm parity.
You may want to mention that `n_type` is `NT_GNU_PROPERTY_TYPE_0`.

If all relocatable files contain `.note.gnu.property` (which contains `NT_GNU_PROPERTY_TYPE_0` notes) with the `GNU_PROPERTY_XXX` bit set, or `-z xxx` is specified, the linker will mark the output file as compatible with the feature

> Add property for CFI extension

"property" is a heavily overloaded term. Consider "program property" or just use `.note.gnu.property` to be more specific.

All assembly files need `.note.gnu.property` marking to be diagnostic free in the absence of a linker option expressing the "force" intention. This makes the CFI features on Arm/x86 quite difficult to deploy...

sorear:
@kito-cheng Isn't the usefulness of this degraded by the fact that we only allocate 1 32-bit property instead of the ~ 3 * 32768 allocated on x86?  When we start to use the next bitmap, an old linker won't be able to propagate it, and so on.

kito-cheng:
Split base program property part into https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/428

riscv-elf.adoc
kito-cheng: Addressed in separated PR, this PR will based on that: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/428

riscv-elf.adoc
```diff
@@ -1395,6 +1395,43 @@ handle program property types even if they are not recognized.
...
+| Bit | Bit Name
+|   0 | GNU_PROPERTY_RISCV_FEATURE_1_CFI_LP_SIMPLE
+|   1 | GNU_PROPERTY_RISCV_FEATURE_1_CFI_SS
+|===
+
+`GNU_PROPERTY_RISCV_FEATURE_1_CFI_LP_SIMPLE` This bit indicate that all executable
+sections are built to be compatible with the landing pad mechanism provided by
+the `Zicfilp` extension. An executable or shared library with this bit set is
+required to generate PLTs with the landing pad (`lpad`) instruction, and all
+label are set to `1`.
```
ved-rivos, line 1437: For a single label scheme a label value of 0 would be more appropriate than 1. Why is the choice to use value of 0. Also we should add a bit for the multi-label scheme.

kito-cheng: Oh, think few more second and found some memory why I try to doing that: Only use `lpad 0` on those function may come from anywhere (e.g. PLT resolver), and use non-0 value to prevent indirect jump without setting up landing pad, 0 is universal value which allow source doing indirect jump without setup x7/t2.

ved-rivos: For a single label case, its okay to use label of 0. Setting label of 1 before every indirect jump adds no additional value. A single label scheme allows an indirect jump to land on any landing pad.

ved-rivos: And is also why multi-label should be default. While single label is available and can be selected - user should make that choice judiciously to not use multi-label..

kito-cheng:
Changes:
- Explicitly specify the labeling scheme in `GNU_PROPERTY_RISCV_FEATURE_1_CFI_LP_SIMPLE`, label will always set to 0.
- Add simple landing pad PLT.
- Add DT_RISCV_SIMPLE_LP_PLT to mark this object use simple landing pad PLT.

riscv-elf.adoc
kito-cheng: > For a single label case, its okay to use label of 0. Setting label of 1 before every indirect jump adds no additional value. A single label scheme allows an indirect jump to land on any landing pad.

I guess the value to toolchain PoC is make sure we have insert right landing pad label value into `t2` :P but anyway I agree with you there is no additional value when we are not talking about toolchain PoC.

sorear:
Why do we need both the dynamic tag and the GNU property?  Surely the type of the PLT entries should be part of what the GNU property encodes.

riscv-elf.adoc
```diff
@@ -673,11 +687,28 @@ in the PLT occupies two 16 byte entries:
...
+    auipc  t2, %pcrel_hi(.got.plt)
+    sub    t1, t1, t3               # shifted .got.plt offset + hdr size + 12
+    l[w|d] t3, %pcrel_lo(1b)(t2)    # _dl_runtime_resolve
+    addi   t1, t1, -(hdr size + 12) # shifted .got.plt offset
+    addi   t0, t2, %pcrel_lo(1b)    # &.got.plt
+    srli   t1, t1, log2(16/PTRSIZE) # .got.plt offset
+    l[w|d] t0, PTRSIZE(t0)          # link map
+    jr     t3
+    nop
+    nop
```
mylai-mtk, line 703: Missing one more `nop` to fill up the added 16 byte?

kito-cheng: Thanks, fixed :)

kito-cheng:
@sorear 

> Why do we need both the dynamic tag and the GNU property? Surely the type of the PLT entries should be part of what the GNU property encodes.

Good point...I was try to learn from AArch64, they have define both GNU prop and dynamic tag, but I realized they didn't use that dynamic tag at all, so...let drop that :P

> I'm not sure it's a good idea to merge this without simultaneously having a design for the multi-label case. There are a few things that I think are needed for multi-label, like software-guarded branches to and from the PLT and avoiding landing pads for symbols that cannot legally be called indirectly, that won't be well exercised by single labels and aren't even in this proposal.

Yeah, my plan is send out a separated PR for that, we have implement simple scheme with slight variant, so simple scheme version should be fine so far, and we gonna to design and implement complex scheme in next few month.

And here is the source tree for who is interesting on that : https://github.com/sifive/riscv-gnu-toolchain/tree/cfi-dev included qemu (Contribute by Rivos) and GNU toolchain.

riscv-elf.adoc
```diff
@@ -673,11 +687,29 @@ in the PLT occupies two 16 byte entries:
...
-entry redirects to the first PLT entry which calls `_dl_runtime_resolve`
-and fills in the GOT entry for subsequent calls to the function:
+And occupies three 16 byte entries for the simple landing pad PLT style:
+[,asm]
+----
+1:  lpad 0
+    auipc  t2, %pcrel_hi(.got.plt)
+    sub    t1, t1, t3               # shifted .got.plt offset + hdr size + 12
+    l[w|d] t3, %pcrel_lo(1b)(t2)    # _dl_runtime_resolve
+    addi   t1, t1, -(hdr size + 12) # shifted .got.plt offset
```
mylai-mtk, line 697: I guess the additional offset `12` is actually `16` here? Since the `jalr` in PLT entries are now the 4th instruction.

kito-cheng:
Changes:
- Rename `GNU_PROPERTY_RISCV_FEATURE_1_CFI_LP_SIMPLE` to `GNU_PROPERTY_RISCV_FEATURE_1_CFI_LP_UNLABELED`
- Fix wrong offset in the first PLT stubs for the simple landing pad PLT.

kito-cheng:
We have implementation on both toolchain, so I think we are ready to land for this PR :)
- LLVM: https://github.com/llvm/llvm-project/pull/127036
- GCC: https://github.com/gcc-mirror/gcc/commit/2b3efe793e9ecf0f98da1daa85a091e931b4ca04

Let me know if further comment on this

mylai-mtk (commented):
@kito-cheng It looks like the content of this PR is stable and ready to be merged, so I guess it the time to review it once again to reconfirm my knowledge and fine tune for wording, garmmar, etc. I've found quite a few typing mistakes/inappropriates in the process, but the content is good. Thanks for the effort!

riscv-elf.adoc
```diff
@@ -1442,6 +1483,58 @@ that a linker or runtime loader needs to check for compatibility.
...
+unlabeled landing pad PLT style, and all of the labels of lpad instructions are
+set to 0, i.e. unlabeled.
+
+`GNU_PROPERTY_RISCV_FEATURE_1_CFI_SS`: This bit indicates that all executable
+sections are built to be compatible with the shadow stack mechanism provided by
+the `Zicfiss` extension. Loading an executable or shared library with this bit
+set requires the execution environment to provide either the `Zicfiss` extension
+or the `Zimop` extension. When the executable or shared library is compiled with
+compressed instructions then loading it with this bit set requires the execution
+environment to provide the `Zicfiss` extension or the `Zimop` extensions.
```
mylai-mtk, line 1536: Typo: Zcmop extension?

riscv-elf.adoc
```diff
@@ -746,11 +760,29 @@ in the PLT occupies two 16 byte entries:
...
+    sub    t1, t1, t3               # shifted .got.plt offset + hdr size + 16
+    l[w|d] t3, %pcrel_lo(1b)(t2)    # _dl_runtime_resolve
+    addi   t1, t1, -(hdr size + 16) # shifted .got.plt offset
+    addi   t0, t2, %pcrel_lo(1b)    # &.got.plt
+    srli   t1, t1, log2(16/PTRSIZE) # .got.plt offset
+    l[w|d] t0, PTRSIZE(t0)          # link map
+    jr     t3
+    nop
+    nop
+    nop
```
mylai-mtk, line 777: @kito-cheng I had not tested this PLT header, since I work with musl libc, which does not support lazy binding. Do you happen to have tested it?

riscv-elf.adoc
kito-cheng: Oh, yeah, thanks for catching this

riscv-elf.adoc
kito-cheng: Yes, we tested that with glibc and passed gcc testsuite

kito-cheng:
LLVM has merge most of related changes, and some PR are review in progress:
- https://github.com/llvm/llvm-project/pull/127036
- https://github.com/llvm/llvm-project/pull/125642
- https://github.com/llvm/llvm-project/pull/91855
- https://github.com/llvm/llvm-project/pull/127193

GCC has been merged code gen for unlabeled landing pad and shadows stack:
- https://github.com/gcc-mirror/gcc/commit/dc76aa0e4d5398104b6b26f08b46524b97de5100
- https://github.com/gcc-mirror/gcc/commit/805a052d0083e8e76c720fd7ac52cb3ccc36c310
- https://github.com/gcc-mirror/gcc/commit/2b3efe793e9ecf0f98da1daa85a091e931b4ca04
- https://github.com/gcc-mirror/gcc/commit/59a869d7196132ae5257fcb068508561d3526c7d

binutils has send out the patch for review:
- https://sourceware.org/pipermail/binutils/2025-May/141120.html
- https://sourceware.org/pipermail/binutils/2025-May/141121.html
- https://sourceware.org/pipermail/binutils/2025-May/141122.html

kito-cheng:
I think this PR is ready to go, does any one can give some blessing for this PR? :)

@Nelson1225 @rui314 @MaskRay @mylai-mtk

riscv-elf.adoc
```diff
@@ -724,15 +724,29 @@ The PLT (Procedure Linkage Table) exists to allow function calls between
 dynamically linked shared objects. Each dynamic object has its own
 GOT (Global Offset Table) and PLT (Procedure Linkage Table).
 
+RISC-V defines several PLT styles, which are used in different situations.
+The default PLT style should be used if the program does not meet the conditions
+for using all other PLT sytles.
```
mylai-mtk, line 729: > for using any of the other PLT sytles.

Small grammar mistake.

mylai-mtk (approved):
Found one small grammar mistake, otherwise LGTM.

riscv-elf.adoc
```diff
@@ -746,11 +760,29 @@ in the PLT occupies two 16 byte entries:
...
 ----
 
-Subsequent function entry stubs in the PLT take up 16 bytes and load a
-function pointer from the GOT. On the first call to a function, the
-entry redirects to the first PLT entry which calls `_dl_runtime_resolve`
-and fills in the GOT entry for subsequent calls to the function:
+And occupies three 16 byte entries for the unlabeled landing pad PLT style:
+[,asm]
+----
+    lpad 0
```
MaskRay, line 766: For a PLT entry, the landing instruction for a function can be elided in many cases as it cannot be an indirect branch target. Potential indirect branch target cases include: canonical PLT entries, non-preemptible ifunc, range extension thunks, and PLT32 data relocations.

`https://github.com/ARM-software/abi-aa/issues/196 `

riscv-elf.adoc
```diff
@@ -1442,6 +1483,94 @@ that a linker or runtime loader needs to check for compatibility.
...
+A relocatable object producer is required to insert a landing pad instruction at
+the target of an indirect branch originating from within the same relocatable
+object.
+
+A relocatable object producer must insert a landing pad instruction at any
+location whose address escapes the relocatable object. This includes all symbols
+that can be exported into the dynamic symbol table by a static linker.
+
+A static linker is required to generate PLT entries with appropriate landing pad
+instructions. During linker relaxation, a static linker may remove unnecessary
```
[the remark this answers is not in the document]
kito-cheng: Here is a follow up PR for adding relocation for lpad: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/452
Also I am working on bintuils side.

riscv-elf.adoc
```diff
@@ -746,11 +760,29 @@ in the PLT occupies two 16 byte entries:
...
-function pointer from the GOT. On the first call to a function, the
-entry redirects to the first PLT entry which calls `_dl_runtime_resolve`
-and fills in the GOT entry for subsequent calls to the function:
+And occupies three 16 byte entries for the unlabeled landing pad PLT style:
+[,asm]
+----
+    lpad 0
+1:  auipc  t2, %pcrel_hi(.got.plt)
+    sub    t1, t1, t3               # shifted .got.plt offset + hdr size + 16
+    l[w|d] t3, %pcrel_lo(1b)(t2)    # _dl_runtime_resolve
```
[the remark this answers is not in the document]
kito-cheng: Current PLT entry/header has use t3, which already incompatible with RVE, so I think it's fine to use t3 here?
https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#procedure-linkage-table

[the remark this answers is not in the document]
jrtc27: Hm, so it does...

kito-cheng:
ping, any further comment for this?

kito-cheng:
Update on binutils part: Nelson is approved, but waiting glibc patch send out before merge.

riscv-elf.adoc
```diff
@@ -1445,6 +1495,94 @@ that a linker or runtime loader needs to check for compatibility.
...
+each bit indicating a feature that the object file is compatible with. The
+linker should perform a bitwise AND operation when merging different objects.
+
+The value of `GNU_PROPERTY_RISCV_FEATURE_1_AND` is defined in <<rv-prog-prop-type>>.
+
+[%autowidth]
+|===
+| Bit | Bit Name
+|   0 | GNU_PROPERTY_RISCV_FEATURE_1_CFI_LP_UNLABELED
+|   1 | GNU_PROPERTY_RISCV_FEATURE_1_CFI_SS
```
xypron, line 1534: Why do we need a bit for unlabeled landing pads?
I would have expected a bit for "have landing pad" and a property for the labeling scheme which may be unlabeled or labeled according to a specific algorithm.

xypron: For easy migration of distro we will probably need at least:
* landing pads are unlabeled, caller does not set x7
* landing pads are unlabeled, but caller set x7 according to labeling scheme
* both landing pads and callers abide to labeling scheme

PR #418, opened 2024-01-04
Standard Fixed-length Vector Calling Convention Variant

kito-cheng:
This proposal outlines a new variant of the calling convention specifically designed for fixed-length vectors. The primary aim of this variant is to facilitate the passing of fixed-length vectors through vector registers. This approach is derived from the standard vector calling convention, it uses the same register conventions and argument passing and return value rules.

A key aspect of this variant is the introduction of ABI_VLEN, which denotes the width of a vector register within this convention. The ABI_VLEN is constrained to be no wider than the ISA's VLEN (Vector Length), ensuring compatibility while allowing for flexibility in different implementations. This parameter can be configured via compiler command line options or through function attributes in source code.

The document recommends setting the default ABI_VLEN to 128 bits, acknowledging it as a common minimal requirement while allowing the flexibility for lower VLEN (32 or 64 bits) as permitted by the ISA. This flexibility is crucial for optimizing the utilization of longer VLENs in various cores.

The proposal specifies how fixed-length vector arguments are passed based on their size relative to ABI_VLEN. Vectors smaller than ABI_VLEN are passed in a single vector argument register, while larger vectors are passed in multiple registers, following the LMUL (Length Multiplier) pattern of 2, 4, or 8, depending on their size.

Additionally, the proposal addresses the handling of structs and unions containing fixed-length vectors. Structs with members that are all fixed-length vectors follow the vector tuple type rules if they conform to size constraints. In contrast, unions with fixed-length vectors adhere to the integer calling convention.


kito-cheng:
cc. @palmer-dabbelt @JeffreyALaw @preames @topperc @rofirrim @lhtin

riscv-cc.adoc
```diff
@@ -329,6 +329,81 @@ type would be passed.
...
+to 128, as it's the most common minimal requirement. However, it is not fixed
+to 128, since the ISA allows the VLEN to be only 32 bits or 64 bits. This
+also enables the utilization of the capacity of longer VLEN. Users can build
+with an optimized library with larger ABI_VLEN for better utilization of those
+cores with longer VLEN.
+
+A fixed-length vector argument is passed in a vector argument register if the
+size of the vector is less than ABI_VLEN bit.
+
+A fixed-length vector argument is passed in two vector argument registers,
```
topperc, line 369: I assume if the hardware VLEN is larger than ABI_VLEN, the elements past ABI_VLEN will continue filling the first register. Since VLEN and ABI_VLEN are both powers of 2, VLEN must be >= 2*ABI_VLEN so no elements are in the second register.

riscv-cc.adoc
```diff
@@ -329,6 +329,81 @@ type would be passed.
...
+
+A struct containing just one fixed-length vector array is passed as though it
+were a vector tuple type if the size of the base element for the array is less
+or equal to 8×ABI_VLEN bit, and the size of the array is less than 8×ABI_VLEN
+bit. Otherwise, it will use the rule defined in the hardware floating-point
+calling convention.
+
+A fixed-length vector argument is passed by reference and is replaced in the
+argument list with the address if it is larger than 8×ABI_VLEN bit or if
+there is a shortage of vector argument registers.
```
lhtin, line 395: This paragraph looks a little more coherent when it's moved below line 385. Because they are both rules describing a single fixed-length vector argument.

riscv-cc.adoc
```diff
@@ -329,6 +329,81 @@ type would be passed.
...
+the register convention and the rules for passing arguments and return values
+are the same.
+
+NOTE: The reason we define a separate calling convention variant is that we
+would like to define a flexible convention to utilize the variable length
+feature in the vector extension, also considering embedded vector extensions,
+such as zve32x.
+
+ABI_VLEN refers to the width of a vector register in the calling convention
+variant.
```
lhtin, line 349: Will ABI_VLEN be used as a condition to determine whether two object files can be linked? For example, if the ABI_VLEN of one object file is 128 and the other is 256, should the link between them be successful?

riscv-cc.adoc
lhtin: Indeed, so it's better if the `continue filling the first register` rule is described in this proposal.

riscv-cc.adoc
```diff
@@ -329,6 +329,81 @@ type would be passed.
 Floating-point registers fs0-fs11 shall be preserved across procedure calls,
 provided they hold values no more than ABI_FLEN bits wide.
 
+=== Standard Fixed-length Vector Calling Convention Variant
```
sorear, line 332: The variant itself seems fine, modulo nits, but how are we planning to enable it?

If it's automatically used by `-march=rva23 -mabi=ilp32d` that will create major compatibility issues for binary distributions that use a fixed ABI and allow mixing packages at different architecture levels (either as an explicit user action, or as an implementation detail when rebuilding the distribution to change the architecture requirement).

If a new `-mabi=` value is required to enable use of the variant, it will be usable on closed systems where all packages are built at once, but not on binary distributions, since there is no expectation that binary code built with different `-mabi=` options is interoperable at all.  This will include Debian and Alpine and might include Android and Fedora if their ABIs are finalized prior to the acceptance of this PR.

If it's enabled on a per-function basis using an attribute, or automatically for functions not visible across DSO boundaries, then it's effectively part of the definition of the attribute or a compiler implementation detail and may belong in riscv-c-api-doc or gccint, not here.

riscv-cc.adoc
sorear, line 395: Given ABI_VLEN=128 and XLEN=64, if all vector argument registers are used, a fixed-length vector argument of length 128 will be passed by reference, but a single-member struct with the same will be passed in two integer registers.  Is this a deliberate break from the base ABIs which try to always pass single-member structs the same as their content?

riscv-cc.adoc
```diff
@@ -329,6 +329,81 @@ type would be passed.
...
+
+A fixed-length vector argument is passed in eight vector argument registers,
+similar to vector data arguments with LMUL=8, if the size of the vector is
+greater than 4×ABI_VLEN bit and less than or equal to 8×ABI_VLEN bit.
+
+A struct containing members with all fixed-length vectors will be passed in
+vector argument registers like a vector tuple type if all members have the
+same length, the length is less or equal to 8×ABI_VLEN bit, and the size of
+the whole struct is less than 8×ABI_VLEN bit. Otherwise, it will use the rule
+defined in the hardware floating-point calling convention.
```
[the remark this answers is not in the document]
kito-cheng: Good catch, I tend to pass that by reference like tuple-type in vector calling convention, will added more word to clarify that.

kito-cheng:
ChangeLog:
 - Reorder rule.
  - Pass struct as tuple-type in register only when vector arg reg is
      enough, otherwise passed in reference.
  - Add NOTE for describe what if ABI_VLEN is smaller than VLEN, also
      come with an example.
  - Add NOTE for describe different functions may use different
      ABI_VLEN values.

riscv-cc.adoc
kito-cheng: I was thinking let it fallback as base ABI, but I realized their may has some scenario is that may put within a POD class for easier operation, so I think we should consider that case as well.

kito-cheng:
ChangeLog:
  - Add rule for single fixed-length vector or fixed-length vector array with size 1.
  - Add rule for zero-length fixed-length arrays.
  - Add explicitly rule for fixed-length vector struct as vector tuple type: pass by ref if no enough arg register.

kito-cheng:
Proposal for function attribute syntax: https://github.com/riscv-non-isa/riscv-c-api-doc/pull/68

riscv-cc.adoc
```diff
@@ -433,6 +433,125 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+similar to vector data arguments with LMUL=8, if the size of the vector is
+greater than 4×ABI_VLEN bit and less than or equal to 8×ABI_VLEN bit.
+
+A fixed-length vector argument is passed by reference and is replaced in the
+argument list with the address if it is larger than 8×ABI_VLEN bit or if
+there is a shortage of vector argument registers.
+
+A struct containing members with all fixed-length vectors will be passed in
+vector argument registers like a vector tuple type if all members have the
+same length, the length is less or equal to 8×ABI_VLEN bit, and the size of
```
4vtomat, line 491: Is a struct with single fixed-length vector considered as tuple type? e.g. { <2 x i32> }.

riscv-cc.adoc
```diff
@@ -433,6 +433,125 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+the whole struct is less than 8×ABI_VLEN bit.
```
4vtomat, line 492:
```suggestion
the whole struct is less than or equal to 8×ABI_VLEN bit.
```

riscv-cc.adoc
4vtomat, line 491:
```suggestion
same length, the length is less than or equal to 8×ABI_VLEN bit, and the size of
```

riscv-cc.adoc
```diff
@@ -433,6 +433,125 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+If there are not enough vector argument registers to pass the entire struct,
+it will pass by reference and is replaced in the argument list with the address.
+Otherwise, it will use the rule defined in the hardware floating-point calling
+convention.
+
+A struct containing just one fixed-length vector or on fixed-length vector
```
4vtomat, line 498:
```suggestion
A struct containing just one fixed-length vector or a fixed-length vector
```

riscv-cc.adoc
```diff
@@ -433,6 +433,125 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+array of length one, it will flattening as a single fixed-length vector argument
```
4vtomat, line 499:
```suggestion
array of length one, it will be flattened as a single fixed-length vector argument
```

riscv-cc.adoc
```diff
@@ -433,6 +433,125 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+if the size of the vector is less than or equal to 8×ABI_VLEN bit.
+
+Struct with zero-length fixed-length arrays use the rule defined in the hardware
+floating-point calling convention, which means it won't consume vector argument
+register eitehr in C or {Cpp}.
+
+A struct containing just one fixed-length vector array is passed as though it
+were a vector tuple type if the size of the base element for the array is less
```
4vtomat, line 507:
```suggestion
were a vector tuple type if the size of the base element for the array is less than
```

riscv-cc.adoc
4vtomat: Since the struct would have >=2 elements here, so I think the length can't be `8xABI_VLEN`.

riscv-cc.adoc
```diff
@@ -433,6 +433,125 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+or equal to 8×ABI_VLEN bit, and the size of the array is less than 8×ABI_VLEN
```
4vtomat, line 508:
```suggestion
or equal to 4×ABI_VLEN bit, and the size of the array is less than or equal to 8×ABI_VLEN
```

4vtomat: Since the array would have >=2 elements here, so I think the length can't be 8xABI_VLEN.

riscv-cc.adoc
```diff
@@ -433,6 +433,125 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+
+NOTE: We suggest the toolchain implementation set the default value of ABI_VLEN
+to 128, as it's the most common minimal requirement. However, it is not fixed
+to 128, since the ISA allows the VLEN to be only 32 bits or 64 bits. This
+also enables the utilization of the capacity of longer VLEN. Users can build
+with an optimized library with larger ABI_VLEN for better utilization of those
+cores with longer VLEN.
+
+A fixed-length vector argument is passed in a vector argument register if the
+size of the vector is less than ABI_VLEN bit.
```
lukel97, line 471: Should this be "no more than ABI_VLEN bits"?

riscv-cc.adoc
kito-cheng: Array with length 1 is legal :P

riscv-cc.adoc
```diff
@@ -467,7 +467,7 @@ with an optimized library with larger ABI_VLEN for better utilization of those
 cores with longer VLEN.
 
 A fixed-length vector argument is passed in a vector argument register if the
-size of the vector is less than ABI_VLEN bit.
+size of the vector is no more than ABI_VLEN bits.
```
lukel97, line 476: I just want to double check, I wasn't concerned about the wording but it seemed to me like it should be `vec_size <= ABI_VLEN` and not `vec_size < ABI_VLEN`. Unless that was intentional?

kito-cheng: Yeah, and I guess I should say something like that to prevent ambiguous:
```suggestion
size of the vector is less than or equal to ABI_VLEN bit.
```

riscv-cc.adoc
```diff
@@ -428,6 +428,179 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+calling convention for fixed-length vectors. ABI_VLEN can vary from 32 bits
+(as in `Zve32x`) up to the maximum supported by the ISA. The flexibility of
+ABI_VLEN enables the convention to adapt to both low-end embedded systems and
+high-performance processors that utilize wider vector registers.
+
+The ABI_VLEN is a parameter of this calling convention variant. It could be set
+by the command line option for the compiler or specified by the function
+attribute in the source code.
+
+NOTE: We suggest the toolchain implementation set the default value of ABI_VLEN
```
topperc, line 468: This isn't possible unless V or Zvl128b is in the ISA string since ABI_VLEN must be less than or equal to the ISA VLEN.

riscv-cc.adoc
```diff
@@ -428,6 +428,179 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+similar to vector data arguments with LMUL=2, if the size of the vector is
+greater than ABI_VLEN bit and less than or equal to 2×ABI_VLEN bit.
+
+A fixed-length vector argument is passed in four vector argument registers,
+similar to vector data arguments with LMUL=4, if the size of the vector is
+greater than 2×ABI_VLEN bit and less than or equal to 4×ABI_VLEN bit.
+
+A fixed-length vector argument is passed in eight vector argument registers,
+similar to vector data arguments with LMUL=8, if the size of the vector is
+greater than 4×ABI_VLEN bit and less than or equal to 8×ABI_VLEN bit.
```
topperc, line 509: `bit` -> `bits`    in two places on this line

riscv-cc.adoc
```diff
@@ -428,6 +428,179 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+Otherwise, it will use the rule defined in the hardware floating-point calling
+convention.
+
+A struct containing just one fixed-length vector or a fixed-length vector
+array of length one, it will be flattened as a single fixed-length vector argument
+if the size of the vector is less than or equal to 8×ABI_VLEN bit.
+
+Struct with zero-length fixed-length arrays use the rule defined in the hardware
+floating-point calling convention, which means it won't consume vector argument
+register eitehr in C or {Cpp}.
```
topperc, line 549: either*

riscv-cc.adoc
```diff
@@ -428,6 +428,179 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+
+A struct containing just one fixed-length vector array is passed as though it
+were a vector tuple type if the size of the base element for the array is less than
+or equal to 8×ABI_VLEN bit, and the size of the array is less than 8×ABI_VLEN
+bit.
```
topperc, line 554: `bit` -> `bits`

riscv-cc.adoc
kito-cheng: Drop those non-mandatory note to prevent confusing, and let this part leave to toolchian convention  or c-api to document that

kito-cheng:
Changes:
- Rebase
- Fix typos
- Drop non-mandatory note about default ABI_VLEN setting

kito-cheng:
LLVM has supported VLS CC https://github.com/llvm/llvm-project/commit/c804e86f558a42f328946331af391d700747fa90

kito-cheng:
Ping, LLVM part has landed a while, does any one from LLVM community is interested to review this again?

riscv-cc.adoc
```diff
@@ -428,6 +428,179 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
 
+NOTE: Functions that use the standard vector calling convention
+variant follow an additional name mangling rule for {Cpp}.
+For more details, see <<Name Mangling for Standard Calling Convention Variant>>.
+
+=== Standard Fixed-length Vector Calling Convention Variant
+
+This section defines the calling convention variant for fixed-length vectors.
+The intention of this variant is to pass fixed-length vectors via the vector
+register. For the definition of a fixed-length vector, see
```
topperc, line 439:
```suggestion
registers. For the definition of a fixed-length vector, see
```

riscv-cc.adoc
```diff
@@ -428,6 +428,179 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+such as `Zve32x`.
+
+ABI_VLEN refers to the width of a vector register in the calling convention
+variant.
+
+The ABI_VLEN must be no wider than the ISA's VLEN, meaning that the ISA may
+support wider vector registers than the ABI, but the ABI's VLEN cannot exceed
+the ISA's VLEN.
+
+ABI_VLEN represents the width (in bits) of the vector register available in the
```
topperc, line 458:
```suggestion
ABI_VLEN represents the width (in bits) of the vector registers available in the
```

riscv-cc.adoc
```diff
@@ -428,6 +428,179 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+calling convention for fixed-length vectors. ABI_VLEN can vary from 32 bits
+(as in `Zve32x`) up to the maximum supported by the ISA. The flexibility of
+ABI_VLEN enables the convention to adapt to both low-end embedded systems and
+high-performance processors that utilize wider vector registers.
+
+The ABI_VLEN is a parameter of this calling convention variant. It could be set
+by the command line option for the compiler or specified by the function
```
topperc, line 465:
```suggestion
by a command line option for the compiler or specified by a function
```

riscv-cc.adoc
```diff
@@ -428,6 +428,179 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+A fixed-length vector argument is passed in eight vector argument registers,
+similar to vector data arguments with LMUL=8, if the size of the vector is
+greater than 4×ABI_VLEN bit and less than or equal to 8×ABI_VLEN bit.
+
+[NOTE]
+===
+Fixed-length vectors that are not a power-of-2 in size will be rounded up to
+the next power-of-2 length for the purpose of register allocation and handling.
+For instance, a vector type like `int32x3_t` (which contains three 32-bit
+integers) will be treated as an `int32x4_t` (a 128-bit vector, as LMUL=1) in
```
topperc, line 516: The LMUL=1 reference here is dependent on ABI_VLEN right?

riscv-cc.adoc
```diff
@@ -428,6 +428,179 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+vector argument registers like a vector tuple type if all members have the
+same length, the length is less than or equal to 4×ABI_VLEN bit, and the size of
+the whole struct is less than or equal to 8×ABI_VLEN bit.
+If there are not enough vector argument registers to pass the entire struct,
+it will pass by reference and is replaced in the argument list with the address.
+Otherwise, it will use the rule defined in the hardware floating-point calling
+convention.
+
+A struct containing just one fixed-length vector or a fixed-length vector
+array of length one, it will be flattened as a single fixed-length vector argument
```
topperc, line 544:
```suggestion
array of length one, will be flattened as a single fixed-length vector argument
```

The comma should maybe be removed as well.

riscv-cc.adoc
```diff
@@ -428,6 +428,179 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+if the size of the vector is less than or equal to 8×ABI_VLEN bit.
+
+Struct with zero-length fixed-length arrays use the rule defined in the hardware
```
topperc, line 547:
```suggestion
Structs with zero-length fixed-length arrays use the rule defined in the hardware
```

Or "A struct"

riscv-cc.adoc
topperc, line 549:
```suggestion
register either in C or {Cpp}.
```

riscv-cc.adoc
topperc, line 554:
```suggestion
bits.
```

riscv-cc.adoc
```diff
@@ -428,6 +428,179 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+attribute in the source code.
+
+NOTE: We suggest the toolchain implementation set the default value of ABI_VLEN
+to 128, as it's the most common minimal requirement. However, it is not fixed
+to 128, since the ISA allows the VLEN to be only 32 bits or 64 bits. This
+also enables the utilization of the capacity of longer VLEN. Users can build
+with an optimized library with larger ABI_VLEN for better utilization of those
+cores with longer VLEN.
+
+A fixed-length vector argument is passed in a vector argument register if the
```
topperc, line 475:
```suggestion
A fixed-length vector argument is passed in one vector argument register if the
```

To increase the similarity with "two", "four", and "eight" below.

riscv-cc.adoc
```diff
@@ -428,6 +428,179 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
...
+minimizing the need for continuous adjustments as new extensions and data types
+are introduced in the future.
+
+The consistency in applying these rules to unsupported element types guarantees
+a smooth transition when future vector extensions become available, allowing for
+seamless integration of new features without requiring significant changes to
+the calling convention.
+===
+
+A fixed-length vector argument is passed in two vector argument registers,
```
topperc, line 499: Are these registers aligned to be a valid LMUL=2 register group?

riscv-elf.adoc
```diff
@@ -202,6 +202,34 @@ See the "Type encodings" section in _Itanium {Cpp} ABI_
 for more detail on how to mangle types. Note that `__bf16` is mangled in the
 same way as `std::bfloat16_t`.
 
+=== Name Mangling for Standard Calling Convention Variant
+
+Function use standard calling convention variant have to append extra ABI tag to
```
topperc, line 207:
```suggestion
Functions using the standard calling convention variant have to append extra ABI tag to
```

riscv-elf.adoc
```diff
@@ -202,6 +202,34 @@ See the "Type encodings" section in _Itanium {Cpp} ABI_
...
+the function name mangling, the rule are same as the "ABI tags" section in
```
topperc, line 208: "rules are the same" or "rule is the same"

riscv-cc.adoc
kito-cheng: Add few sentence to say it follow the same register constraint, also for LMUL 4 and LMUL 8

topperc:
> If you change anything regarding structs then how can it be specified per function? Perhaps leave anything related to structs out of the proposal.

It only changes how structs are passed and returned for that function. The attribute must be on the prototype so the caller will know the ABI for the call. What's your concern?

kito-cheng:
GCC PoC is here, gonna to summit to upstream for review in next few day:
https://github.com/kito-cheng/gcc/tree/kitoc/vls-cc

kito-cheng:
Gonna to merge this since we have this on both LLVM and GCC, also reviewed for a while, although there is few room to improve on wording, but that relative minor and can be update later.

PR #420, opened 2024-01-25
Fix TLSDESC description

MaskRay:
The table specifies R_RISCV_TLSDESC_LOAD_LO12 and
R_RISCV_TLSDESC_ADD_LO12, while some text uses the `_I` suffix. Remove the `_I` suffix to be compatible with
LLVM and the proposed binutils patch.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1857,8 +1857,8 @@ Relaxation result:
 
 [,asm]
 ----
-	auipc   tX, <pcrel-got-offset-for-symbol-hi>
-	{ld,lw} a0, <pcrel-got-offset-for-symbol-lo>(tX)
+	auipc   a0, <pcrel-got-offset-for-symbol-hi>
+	{ld,lw} a0, <pcrel-got-offset-for-symbol-lo>(a0)
 ----
 --
 
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1041,8 +1041,8 @@ assembler functions. The emitted relocations are in the comments.
 ----
 label:
 	auipc tX, %tlsdesc_hi(symbol)         // R_RISCV_TLSDESC_HI20 (symbol)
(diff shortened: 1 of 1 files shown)


MaskRay:
@ishitatsuyuki

ishitatsuyuki:
The _I change looks good to me.

I’m not sure if the tX change is necessary. Both a0 and tX are legal choices here (although tY is not guaranteed to be alive because LOAD is allowed to be reordered after ADD). I think a0 is better overall because the resulting instruction sequence is macro-fusable if both instructions are adjacent.

ilovepi:
I think it makes sense that things are kept consistent with the initial template,  so the use of `tX` seems appropriate to me.

If there's a big concern about that, then maybe  we can add aa comment/note that using `a0` would maximize the benefit in most cases, but is up to the implementation.

ishitatsuyuki:
> I think it makes sense that things are kept consistent with the initial template, so the use of tX seems appropriate to me.

I’m not sure if I see why this is more consistent. The destination register for the ADD instruction is a0, so I think it makes more sense for the relaxed instruction to target a0 too.

It’s fine to clarify that both a0/tX can be used, but if we want to pick one, I don’t see any reason so far why tX should be used instead.

MaskRay:
> > I think it makes sense that things are kept consistent with the initial template, so the use of tX seems appropriate to me.
> 
> I’m not sure if I see why this is more consistent. The destination register for the ADD instruction is a0, so I think it makes more sense for the relaxed instruction to target a0 too.
> 
> It’s fine to clarify that both a0/tX can be used, but if we want to pick one, I don’t see any reason so far why tX should be used instead.

I was thinking whether we needed the flexibility but I guess the answer is probably no.
From the perspective of implementations, a fixed register is easier, so I reverted the auipc wording change :)

ishitatsuyuki (approved):
Thanks!

kito-cheng (approved):
Thanks :)

Issue #401, opened 2023-09-28
Embedding R_RISCV_RELAX to another relocations


rui314:
Object files for RISC-V contain lots of `R_RISCV_RELAX` relocations, and the number seems astonishing. If my calculation is correct, if I build the mold linker as a release binary for RISC-V, `R_RISCV_RELAX` relocation records occupy more than 10% of the total input object files (the input object files are in total 132MiB, and they contains 615,542 `R_RISCV_RELAX` relocations, so the relocations occupy 615,542 * 24 = 14,773,008 bytes.) That's way more than what I expected.

(Can someone verify it with your binary? I believe my calculation is correct but still can't completely believe it because it's too big.)

We do care about object file size. Large object files don't only take up more disk space but also slows down build system because of slow IO.

`R_RISCV_RELAX` essentially conveys a single bit information that the relocation pointing to the same place is relaxable. But on disk, each `R_RISCV_RELAX` occupies 24 _bytes_ of space on RV64 because it's represented as an independent relocation record. That's 192x data bloat. It feels we should do something to improve the situation. Do you all have any ideas on how to improve it?

As a starter, I'd like to propose the following scheme to embed the `R_RISCV_RELAX` bit to relocation record.

**Proposal**

RV64's relocation record looks like this:

```
struct {
  u64 r_offset;
  u32 r_type;
  u32 r_sym;
  i64 r_addend;
};
```

where `r_type` represents the relocation type. The maximum relocation type is currently limited to 256, and with all possible future extensions, it is hard to imagine that we'd need more than 2^16 distinctive relocation types (I believe we'll never have more than 1024 relocation types, but I'm erring on the side of caution.) That gives us the opportunity to redefine the upper half of the `r_type` record as follows:

```
struct {
  u64 r_offset;
  u16 r_type;
  u16 reserved : 15;
  u16 relaxable : 1;
  u32 r_sym;
  i64 r_addend;
};
```

`relaxable` bit is turned on if a relocation is relaxable. In this scheme, we can embed all `R_RISCV_RELAX` relocations to adjacent relocations.

If an object file containing the new relocation records is consumed by a tool that doesn't understand the new format, the tool would report an "unknown relocation type" error because the `relaxable` bit would be interpreted as part of `r_type`. So we need to implement it first to the linker and other ELF-consuming tools, wait for a few years and then turn it on for the assembler/compiler.

[...]

PR #421, opened 2024-01-25
Relaxable TLSDESC only requires that R_RISCV_TLSDESC_HI20 to be paired with R_RISCV_RELAX

MaskRay:
Requiring 4 R_RISCV_RELAX relocations impose a large size increase to the relocatable file. To mitigate this size increase, we can say that the whole TLSDESC code sequence is relaxable if the first instruction (R_RISCV_TLSDESC_HI20) is paired with R_RISCV_RELAX.

A statically linked executable typically has a simple relocation resolver that handles just RELATIVE/IFUNC. For a `-fpic -mtls-dialect=desc` relocatable file, the linker is required to perform TLS optimization to local-exec for a statically linked executable. Therefore, instruction rewriting is essentially inevitable if the relocation resolver is kept simple. Ths general-dynamic TLS model, without a defined optimization scheme, actually has the same issue.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1848,9 +1848,9 @@ Relaxation candidate (`tX` and `tY` can be any combination of two general purpos
 ----
 label:
 	auipc tX, <hi>      // R_RISCV_TLSDESC_HI20 (symbol), R_RISCV_RELAX
-	lw    tY, tX, <lo>  // R_RISCV_TLSDESC_LOAD_LO12 (label), R_RISCV_RELAX
-	addi  a0, tX, <lo>  // R_RISCV_TLSDESC_ADD_LO12 (label), R_RISCV_RELAX
-	jalr  t0, tY        // R_RISCV_TLSDESC_CALL (label), R_RISCV_RELAX
+	lw    tY, tX, <lo>  // R_RISCV_TLSDESC_LOAD_LO12 (label)
+	addi  a0, tX, <lo>  // R_RISCV_TLSDESC_ADD_LO12 (label)
+	jalr  t0, tY        // R_RISCV_TLSDESC_CALL (label)
 ----
 
 Relaxation result:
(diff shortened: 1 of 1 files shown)


MaskRay:
@ilovepi @ishitatsuyuki 

My lld/ELF implementation checks whether the first instruction (R_RISCV_TLSDESC_HI20) has R_RISCV_RELAX
https://github.com/llvm/llvm-project/pull/79239

On the LLVM side, I want to push for emitting one R_RISCV_RELAX as well.
If people say "this is not conforming", I think I'd like a mode to remove the redundant R_RISCV_RELAX.

We don't need to do the same as `R_RISCV_TPREL_HI20`.

ishitatsuyuki:
Looks sensible. I think mold works without the RELAX annotations on paired relocs too, but I'll need to double check.

rui314:
I think that this issue is red herring because the number of RELAX relocations for TLSDESC is probably fewer than 0.1% of those for `R_RISCV_HI20` and `R_RISCV_LO12_I`. To solve it, I believe we should really consider something like https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/401. I'm not opposed to this proposal, but I'm skeptical about whether this change would make any meaningful difference.

MaskRay:
> I think that this issue is red herring because the number of RELAX relocations for TLSDESC is probably fewer than 0.1% of those for `R_RISCV_HI20` and `R_RISCV_LO12_I`. To solve it, I believe we should really consider something like #401. I'm not opposed to this proposal, but I'm skeptical about whether this change would make any meaningful difference.

I made a reply: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/401#issuecomment-1912548898

I think we can ensure that `R_RISCV_RELAX` causes less harm (size increase) for newer relaxable code sequences like TLSDESC.

kito-cheng (approved):
For those kind of chained relocation, only the first relocation is enough to represent it's relaxable.

`R_RISCV_PCREL_HI20`/`R_RISCV_PCREL_LO12_[I|S]` may be the candidate, however I realized we really need a ABI version tag before doing those not fully compatible changes.

And another candidate for this idea is the long instruction sequence version of large code model I think :)

MaskRay:
Can this be merged? :)

rui314:
LGTM

PR #422, opened 2024-02-02
Name mangling for standard vector calling convention

kito-cheng:
Add extra name mangling rule for standard vector calling convention on C++ ABI, fortunately the Itanium C++ ABI already defined mangling rule for ABI tag, so we just need to define the ABI tag name.

Example for the mangling rule:
```c
__attribute__((riscv_vector_cc)) void foo();
// _Z3fooB15riscv_vector_ccv with this mangling rule
// _Z3foov without this mangling rule
```

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -433,6 +433,10 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
 all vector registers. Hence, the standard vector calling convention variant
 won't disrupt the `jmp_buf` ABI.
 
+NOTE: Functions that use the standard vector calling convention
+variant follow an additional name mangling rule for {Cpp}.
+For more details, see <<Name Mangling for Standard Calling Convention Variant>>.
+
 === ILP32E Calling Convention
 
 IMPORTANT: RV32E is not a ratified base ISA and so we cannot guarantee the
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -149,6 +149,34 @@ See the "Type encodings" section in _Itanium {Cpp} ABI_
 for more detail on how to mangle types. Note that `__bf16` is mangled in the
 same way as `std::bfloat16_t`.
 
+=== Name Mangling for Standard Calling Convention Variant
+
+Function use standard calling convention variant have to append extra ABI tag to
+the function name mangling, the rule are same as the "ABI tags" section in
+_Itanium {Cpp} ABI_.
+
+.ABI Tag name for calling convention variants
+[cols="5,2"]
+[width=80%]
+|===
+| Name                                       | ABI tag name
+
+| Standard vector calling convention variant | riscv_vector_cc
+|===
+
+
+For example:
+[,c]
+----
+    __attribute__((riscv_vector_cc)) void foo();
+----
+
+is mangled as
+[,c]
+----
+    _Z3fooB15riscv_vector_ccv
+----
+
 === Name Mangling for Vector Data Types, Vector Mask Types and Vector Tuple Types.
 
 The vector data types and vector mask types, as defined in the section


riscv-cc.adoc
```diff
@@ -433,6 +433,10 @@ NOTE: `setjmp`/`longjmp` follow the standard calling convention, which clobbers
 all vector registers. Hence, the standard vector calling convention variant
 won't disrupt the `jmp_buf` ABI.
 
+NOTE: Functions that use the standard vector calling convention
```
sorear, line 436: I don't think we need to specially mangle the names of _functions_ that use the vector calling convention.  While arguably useful for function multi-versioning and the `[[gnu::target_version]]` attribute in particular, this is orthogonal to C++ overload resolution.

We do need to mangle _function pointer types_ that use the vector calling convention, whether or not the function taking the parameter itself uses the vector calling convention, to allow overloading a function that takes arguments with different ABIs.

riscv-elf.adoc
```diff
@@ -149,6 +149,34 @@ See the "Type encodings" section in _Itanium {Cpp} ABI_
 for more detail on how to mangle types. Note that `__bf16` is mangled in the
 same way as `std::bfloat16_t`.
 
+=== Name Mangling for Standard Calling Convention Variant
+
+Function use standard calling convention variant have to append extra ABI tag to
+the function name mangling, the rule are same as the "ABI tags" section in
```
sorear, line 155: ABI tags (B) are used for the GNU abi_tag extension, which uses a string that is entirely under the control of a library author.  It is probably wrong to use for architecture-specific features, and it does not fix my original example since it cannot be used on unnamed entities.

I was originally thinking of using a vendor-defined pointer qualifier as defined in [5.1.5.1 Qualified types](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.qualified-type), but the closest existing thing to what I'm asking for is the definition of Dx in [5.1.5.3 Function types](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.function-type).  We may need upstream guidance.

kito-cheng: Wait, I thought it work? https://gcc.godbolt.org/z/95shE9a13

kito-cheng: (let me read the Itanium C++ ABI more carefully again)

sorear: `f` and `g` in your example have the same calling convention, so there's no need for the compiler to generate two versions.  If you add __attribute__((aarch64_sve_pcs)) to g, it breaks.

kito-cheng: Oh, yeah, I got your main point: it should mangling function (pointer) type, not function name, I thought abi_tag works for function type but it only decorate/mangling function name.

sorear:
I filed itanium-cxx-abi/cxx-abi#178 to maybe get an expert opinion.  Hope that isn't overstepping.

kito-cheng:
> I filed https://github.com/itanium-cxx-abi/cxx-abi/issues/178 to maybe get an expert opinion. Hope that isn't overstepping.

Thanks for doing this, I think you are better person to create that since you have more context than me on this, will support you on that PR if needed :)

sorear:
swiftcall in clang appears to already have the behavior we need.

kito-cheng:
Change to draft since this PR is not intend to merge so far, but we may keep this PR for keep tracking and/or discuss this issue.

kito-cheng:
Oh, I should reply after read the PR, I didn't notice there is some discussion within the PR, anyway let me find some time to update and update this PR later.

(@sorear you can create another PR in case you are interested since I may deal with other stuffs around the psABI repo first)

kito-cheng:
Not really spec related, but kinda bookmark for myself:

clang implementation for Swift name mangling:
[CXXNameMangler::mangleExtParameterInfo@ItaniumMangle.cpp](https://github.com/llvm/llvm-project/blob/main/clang/lib/AST/ItaniumMangle.cpp#L3492)

pz9115:
LGTM, and do we need a version number in the doc to deal possible future version changes.

PR #424, opened 2024-02-13
Clarify Vector Register Calling Convention

ewlu:
The current wording for the vector register calling convention is unclear since both conventions are labelled as standard. This may cause confusion as to which calling convention is used by default. To avoid future confusion, explicitly separate the two conventions into different subsections to highlight the differences and remove "standard" label to the variant


kito-cheng (approved):

kito-cheng:
LGTM, Thanks :)

Issue #412, opened 2023-12-05
Calling convention description of va_list et al. are unclear


PerMildner:
https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc#va_list-va_start-and-va_arg says:

> A callee with variadic arguments is responsible for copying the contents of registers used to pass variadic arguments to the vararg save area, which must be contiguous with arguments passed on the stack. The va_start macro initializes its va_list argument to point to the start of the vararg save area. The va_arg macro will increment its va_list argument according to the size of the given type, taking into account the rules about 2×XLEN aligned arguments being passed in "aligned" register pairs.

The "vararg save area" does not seem to be defined or mentioned elsewhere.

It is not clear to me, from the quote, how the layout of the vararg save area is affected by the the use of '"aligned" register pairs'. I would expect `va_arg` to operate solely against an in-memory "vararg save area" and thus not need to know or care about register placement details.

So, perhaps some reference or details would be good?

jrtc27:
It is defined there. You save all the remaining GPRs to the region of memory that is in your stack frame and contiguous with the arguments passed on the stack, i.e. the top of your stack frame. That region  of memory is the vararg save area.

The concept of aligned register pairs comes after that definition and does not affect va_start. It is in the part that defines how va_arg operates on that region of memory. That rule exists to allow you to _not_ have to special-case whether a value was passed in a register or on the stack, since it makes the vararg save area look identical to how values are passed on the stack. Therefore the approach here is "spill it all to memory and treat it like it was all passed on the stack". Compilers are free to optimise and extract directly from registers if they wish, this just governs the ABI-visible behaviour of what a va_list is (a pointer) and points to (a single block of argument memory).

PR #426, opened 2024-02-24
va_list & bit-field improvements

### LINKED ISSUE #410: Bitfield integer calling convention garbled
https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc#integer-calling-convention says:

> Bitfields may larger than its integer type, bits excess than its integer type will treat as padding bits, then padding to begin at the next alignment boundary.

This makes no sense, I think.

sorear:
Documentation improvements intended to fix #410 and #412 with no changes to requirements.  Longer descriptions in commit messages.

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -216,21 +216,6 @@ Empty structs or union arguments or return values are ignored by C compilers
 which support them as a non-standard extension.  This is not the case for {Cpp},
 which requires them to be sized types.
 
-Bitfields are packed in little-endian fashion. A bitfield that would span the
-alignment boundary of its integer type is padded to begin at the next
-alignment boundary. For example, `struct { int x : 10; int y : 12; }` is
-a 32-bit type with `x` in bits 9-0, `y` in bits 21-10, and bits 31-22
-undefined.  By contrast, `struct { short x : 10; short y : 12; }` is a 32-bit
-type with `x` in bits 9-0, `y` in bits 27-16, and bits 31-28 and bits 15-10
-undefined.
-
-Bitfields may larger than its integer type, bits excess than its integer
-type will treat as padding bits, then padding to begin at the next alignment
-boundary. For example `struct { char x : 9; char y; }` is a 24 byte type with
-`x` in bits 7-0, `y` in bit 23-16, and bits 15-8 undefined,
-`struct { char x : 9; char y : 2 }` is a 16-bit type with `x` in bits 7-0, `y`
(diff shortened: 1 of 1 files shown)


riscv-cc.adoc
```diff
@@ -639,6 +624,39 @@ The type `size_t` is defined as `unsigned int` for RV32 and `unsigned long` for
...
+
+Zero length bit-fields are aligned relative to the start of the containing
+struct according to their declared type and, since they must be unnamed, do not
+affect the struct alignment.  Access to a bit-field before or after a
+zero-length bit-field will not access any byte on the other side of the
+zero-length bit-field.  This can require bit-field accesses to be misaligned or
+split.  For example `struct { int x : 8; char : 0; int y : 16; int z : 8; }` is
+a 32-bit type in which `y` occupies bits 23-8, but accesses to `y` are not
+allowed to access byte 0 so they must either perform a halfword access starting
+at byte 1 or two byte accesses at byte 1 and byte 2.
```
kito-cheng, line 659: The constraint is too strong, and checked with RISC-V GCC/LLVM code gen, RISC-V GCC seems not doing that, also do you mind point out related x86-64 ABI part, I only found zero-width bit field will classified as e NO_CLASS[1], no so strong constraint on the zero-length bit-field, but I maybe missing something.

- RISC-V GCC (13 and trunk): Emit a double word load and then shift [2]
```
struct X{ int x : 8; char : 0; int y : 16; int z : 8; };

struct X x;

int foo(){
    return x.y;
}
```
- RISC-V clang (17 and trunk): will emit two byte load, and then combine together.

[1]https://gitlab.com/x86-psABIs/x86-64-ABI/-/blob/master/x86-64-ABI/low-level-sys-info.tex?ref_type=heads#L662
[2] https://godbolt.org/z/ErsKjdvs9

kito-cheng (commented):
Thanks for the PR, but I guess I have few questions around zero-length bit-fields.

riscv-cc.adoc
sorear: As @ptomsich says this requirement is only for writes and it's a C requirement, not something novel in the psABI. I dropped the specific wording about byte and word accesses because we don't have any similar language for non-bit-field struct members, but left something to point readers at the C requirement.

kito-cheng (approved):
LGTM :)

kito-cheng:
Gonna merge this, thanks for the improvement!

PR #427, opened 2024-02-24
Revert "Add GP-relative relocations"

sorear:
This reverts commit d49e48097bcad410e7e6d96ebf9d94d6d41dc9c0.

Both the ePIC and FDPIC ABIs are expected to define relocation types which can be used for both PC-relative and GP-relative data, and are expected to define relaxations to optimize access to both types of data. Explicit, compiler-generated relocations using GP are only useful in cases where an object is known at compile time to be local and present in the GP-relative segment.  The relaxations will result in GP-relative accesses, but the relocations involved will be linker internal and need not have allocated standard numbers.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -463,15 +463,7 @@ Description:: Additional information about the relocation
                                             <| S + A - P
 .2+| 45      .2+| RVC_JUMP      .2+| Static  | _CJ-Type_         .2+| 11-bit PC-relative jump offset
                                             <| S + A - P
-.2+| 46      .2+| *Reserved*    .2+| -       |                   .2+| Reserved for future standard use
-                                            <|
-.2+| 47      .2+| GPREL_LO12_I  .2+| Static  | _I-type_          .2+| Low 12 bits of a 32-bit GP-relative address, `%gprel_lo(symbol)`
-                                            <| S + A - GP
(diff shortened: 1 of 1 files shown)


MaskRay (approved):
I have made extensive investigations of older solutions (-msep-data, -mid-shared-library, -mno-pic-data-is-text-relative, -fropi and -frwpi, -mfdpic; m68k/sh4/arm) and will keep adding notes to https://maskray.me/blog/2024-02-20-mmu-less-systems-and-fdpic

I agree that these GP-relative relocation types will not be useful.

With suitable optimizations and relaxations, the ePIC relocations can be completely replaced.

AFAIK, these relocation types haven't been used in any upstream projects.

kito-cheng:
LGTM but I would like to merge this until https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/405 update :)

cc. @luismarques

luismarques:
Sounds reasonable.

MaskRay:
I am now working on ARM FDPIC for LLVM+Clang to get some experience. Yes, I am more convinced we should remove relocation types for direct GP-relative addressing:)

sorear:
Should I merge this, then?

kito-cheng (approved):
Let moving forward :)

PR #429, opened 2024-02-27
FDPIC/ePIC draft specification and work plan

sorear:
This needs extensive editing to match the existing document, and I also haven't even started on prototype coding, but it's become clear the specification needs to exist somewhere other than my answers to questions on IRC so let's start a draft here.

I intend to create a second PR when I have something ready for formal review. Feel free to use this as a discussion thread for the high-level design.

### DIFF
--- a/riscv-fdpic-epic.adoc
+++ b/riscv-fdpic-epic.adoc
@@ -1,0 +1,549 @@
+[[riscv-fdpic-epic]]
+= RISC-V FDPIC and ePIC ABI supplement
+ifeval::["{docname}" == "riscv-cc"]
+include::prelude.adoc[]
+endif::[]
+
+== Purpose and need
+
+The RISC-V ELF psABI defines PIC code models which can be used to change the
+load address of an object or combine several independently linked objects
+without modifying executable memory. This supports loading of shared libraries,
+as well as dynamic memory management in systems with a single address space.
+However, the existing PIC mechanisms assume a constant displacement between the
+code and data of a single object, which precludes sharing code between multiple
+instances of a single object in an environment without address translation, as
+well as using code located in read-only memory if the location of the data is
+variable.
+
+The FDPIC and ePIC supplement permits dynamic memory management in such cases
+by defining new code models where the code and data have independently varying
+addresses for each object in a process image. These models may also be useful
+as an alternative to the non-PIC large model in cases where the code and data
+occupy fixed addresses, but at a larger separation than is accomodated by the
+range limits of the exsiting code models.
+
+== High level alternatives
+
+Not providing code models supporting independent relocation of code and data
+requires RISC-V systems without address translation to provide a separate
+instance of the code for each instance of the data, and in writable memory if
+the number or location of the data instances can change.
+
+This document proposes an implementation strategy called **FDPIC**.  Under
+FDPIC the `gp` register holds a per-function base address, called the **GOT
+address**, for the data of the object containing the currently executing
+function. All control transfers that potentially cross between FDPIC objects
+must load `gp` for the new function to establish the ABI environment. All data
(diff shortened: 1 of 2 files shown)


riscv-fdpic-epic.adoc
```diff
@@ -0,0 +1,549 @@
...
+== Data representation
+
+FDPIC redefines a function pointer as a pointer to a struct containing two
+address-sized values, called a **function descriptor** and the ABI's namesake.
+The first value in the descriptor is the address of the first instruction of
+the function's code. The second value is the `gp` register for the function.
+
+NOTE: This style of function descriptor is used in specialized FDPIC ABIs for
+Blackfin, FR-V, SuperH, and Arm, and is part of the default ABI for PA-RISC,
+POWER (v1 only, has a third "static chain" field), and IA-64.
```
MaskRay, line 71: In PPC64 ELFv1, "static chain" is for dynamic linker's lazy binding. This is "environment pointer"?

riscv-fdpic-epic.adoc
```diff
@@ -0,0 +1,549 @@
...
+`gp` from the descriptor; all other calls use the value of `gp` the caller was
+entered with.
+
+FDPIC `gp` rules apply orthogonally to all standard calling convention
+variants and do not affect the setting of `STO_RISCV_VARIANT_CC`.
+
+In ePIC, `gp` is invariant and holds the GOT address for the process image at
+all instruction boundries.
+
+== Range extension thunks
```
MaskRay, line 103: This seems to belong to another proposal?

riscv-fdpic-epic.adoc
```diff
@@ -0,0 +1,549 @@
...
+                                       a     R_RISCV_TLSDESC_ADD_LO12 0
+                                       e  jalr t0, t2
+                                       e     R_RISCV_TLSDESC_CALL 0
+```
+
+== Changes to other specifications
+
+=== riscv-toolchain-conventions
+
+Add -mfdpic option for FDPIC and -mepic option for ePIC.
```
MaskRay, line 481: I think -mepic in the previous proposal is in spirit identical to -msep-data (m68k, bfin) and -mno-pic-data-is-text-relative (arm).

`-fPIE -mfdpic` with hidden visibility declarations can replace `-fPIE -msep-data`, though setting the call-clobbered FDPIC register has slight overhead.

Do we want to introduce another option to avoid the minor FDPIC register overhead?

riscv-fdpic-epic.adoc
sorear: I think the two terms are interchangeable; they both go into r11 anyway. x86-64 also calls the register static chain.

riscv-fdpic-epic.adoc
sorear: You're saving two memory accesses on a generic function call. It's probably close to `-fomit-frame-pointer` in total.

kito-cheng:
Ping, @sorear do you have further update on this, or do you need some help on PoC implementation?

Ruinland:
Hi @sorear and @kito-cheng  , I am wondering how's the current status of this psABI ?
If my memories serve me correctly, there were once a proposal to add toolchain and musl-libc support with RISC-V FDPIC.
Yet judged from what I see in this PR, the spec seemed have never landed ?

kito-cheng:
> Hi @sorear and @kito-cheng , I am wondering how's the current status of this psABI ? If my memories serve me correctly, there were once a proposal to add toolchain and musl-libc support with RISC-V FDPIC. Yet judged from what I see in this PR, the spec seemed have never landed ?

Yeah, this PR is kinda lack of resource to moving forward, I am happy to push this forward, however this is not on my priority list yet, and seems @sorear are busy with other stuffs, so...volunteer is welcome :P

PR #433, opened 2024-04-09
Fix AsciiDoc syntax error

nick-knight:
To prevent AsciiDoc formatting substitutions in cases like these, single and double backslashes suffice:
https://docs.asciidoctor.org/asciidoc/latest/subs/prevent/

A more powerful tool is the "passthrough":
https://docs.asciidoctor.org/asciidoc/latest/pass/
This is a heavier hammer than is needed here. I do see passthroughs used elsewhere in this document, in cases where backslashes would suffice. I despise AsciiDoc too much to write a longer PR descriptio


kito-cheng:
Thanks :)

PR #434, opened 2024-04-23
Add complex labeling scheme for landing pad

kito-cheng:
NOTE: it's chained PR, which based on #417 
TODO: We don't add mechanism for generating right PLT yet, we may need a specialized section to record the function signature or hash, so that static linker can generate right label at PLT entries.

---

Function signature based labeling scheme, follow the "Function types" mangling rule defined in Itanium C++ ABI.

With few specific rules for member functions:

- Member functions  use the "Pointer-to-member types" mangling rule
- Virtual functions use the member function type of the base class that first defined the virtual function.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -662,6 +662,7 @@ using all other PLT sytle.
 |===
 | Default PLT                  | -
 | Simple landing pad PLT | Must use this PLT style when `GNU_PROPERTY_RISCV_FEATURE_1_CFI_LP_SIMPLE` is set.
+| Complex landing pad PLT | Must use this PLT style when `GNU_PROPERTY_RISCV_FEATURE_1_CFI_LP_COMPLEX` is set.
 |===
 
 The first entry of a shared object PLT is a special entry that calls
@@ -672,6 +673,9 @@ GOT offsets lazily on the first call to any function, except when
 dynamic linker before the executable is started. Lazy resolution of GOT
 entries is intended to speed up program loading by deferring symbol
 resolution to the first time the function is called.
+
+The PLT entry is 16 bytes for the default PLT style and the simple landing pad
+PLT style, and 32 bytes for the complex landing pad PLT style.
 
 The first entry in the PLT occupies two 16 byte entries for the default PLT style:
 
@@ -704,7 +708,25 @@ And occupies three 16 byte entries for the simple landing pad PLT style:
     nop
 ----
 
-Subsequent function entry stubs in the PLT take up 16 bytes.
+The complex landing pad PLT style occupies two 32 byte entries:
+
+[,asm]
+----
+1:  lpad 0
+    auipc  t2, %pcrel_hi(.got.plt)
+    sub    t1, t1, t3               # shifted .got.plt offset + hdr size + 24
+    l[w|d] t3, %pcrel_lo(1b)(t2)    # _dl_runtime_resolve
+    addi   t1, t1, -(hdr size + 24) # shifted .got.plt offset
+    addi   t0, t2, %pcrel_lo(1b)    # &.got.plt
+    srli   t1, t1, log2(32/PTRSIZE) # .got.plt offset
+    l[w|d] t0, PTRSIZE(t0)          # link map
+    jr     t3
+    nop
+    nop
+----
+
+Subsequent function entry stubs in the PLT take up 16 bytes or 32 bytes depends
+on the style.
 On the first call to a function, the entry redirects to the first PLT entry
 which calls `_dl_runtime_resolve` and fills in the GOT entry for subsequent
 calls to the function.
@@ -725,6 +747,19 @@ The code sequences of the PLT entry for the the simple landing pad PLT style:
     auipc   t3, %pcrel_hi(function@.got.plt)
     l[w|d]  t3, %pcrel_lo(1b)(t3)
     jalr    t1, t3
+----
+
+The code sequences of the PLT entry for the the complex landing pad PLT style:
+[,asm]
+----
+1:  lpad    <hash-value-for-function>
+    auipc   t3, %pcrel_hi(function@.got.plt)
+    l[w|d]  t3, %pcrel_lo(1b)(t3)
+    lui     t2, <hash-value-for-function>
+    jalr    t1, t3
+    nop
+    nop
+    nop
 ----
 
 ==== Procedure Calls
@@ -1469,6 +1504,7 @@ a different features.
 | Bit | Bit Name
 |   0 | GNU_PROPERTY_RISCV_FEATURE_1_CFI_LP_SIMPLE
 |   1 | GNU_PROPERTY_RISCV_FEATURE_1_CFI_SS
+|   2 | GNU_PROPERTY_RISCV_FEATURE_1_CFI_LP_COMPLEX
 |===
 
 `GNU_PROPERTY_RISCV_FEATURE_1_CFI_LP_SIMPLE` This bit indicate that all executable
@@ -1485,6 +1521,12 @@ or the `Zimop` extension. When the executable or shared library is compiled with
 compressed instructions then loading an executable with this bit set requires
 the execution environment to provide the `Zicfiss` extension or to provide both
 the `Zcmop` and `Zimop` extensions.
+
+`GNU_PROPERTY_RISCV_FEATURE_1_CFI_LP_COMPLEX` This bit indicate that all executable
+sections are built to be compatible with the landing pad mechanism provided by
+the `Zicfilp` extension. An executable or shared library with this bit set is
+required to generate PLTs with the landing pad (`lpad`) instruction, and all
+label are set to a value which hashed from its function signature.
 
 === Mapping Symbol
 
@@ -1528,6 +1570,61 @@ like memcpy provides two versions, one built with `rv64gc`, and one built with
 attribute is recording for minimal execution environment requirements, so the
 ISA information from arch attribute is not enough for the disassembler to
 disassemble the `rv64gcv` version correctly.
+
+== Label Value Compuatation for Complex Labeling Scheme Landing Pad
+
+The label value for the complex labeling scheme landing pad is computed from the
+hash of the function signature string, which uses the same scheme as the
+"Function types" mangling rule defined in the _Itanium {Cpp} ABI_
+<<itanium-cxx-abi>>, the value is taken from the lower 20 bits of the MD5
+hash result of the function signature string.
+
+Additionally, here are a few specific rules for {Cpp} member functions:
+
+- {Cpp} member functions should use the "Pointer-to-member types" mangling rule
+  defined in the _Itanium {Cpp} ABI_ <<itanium-cxx-abi>>.
+- Virtual functions in {Cpp} should use the member function type of the base
+  class that first defined the virtual function.
+
+
+Example:
+
+[,cxx]
+----
+
+double foo(int, float *);
+
+class Base
+{
+public:
+  virtual void memfunc1();
+  virtual void memfunc2(int);
+};
+
+class Derived : public Base
+{
+public:
+  virtual void memfunc1();
+  virtual void memfunc3(double);
+  void memfunc4();
+};
+
+class DerivedDerived : public Derived
+{
+public:
+  virtual void memfunc2(int);
+  virtual void memfunc3(double);
+};
+
+----
+
+The function signatures for the above functions are described below:
+
+- `foo` is encoded as `FdiPfE`.
+- `Base::memfunc1` and `Derived::memfunc1` are both encoded as `M4BaseFvvE`.
+- `Base::memfunc2` and `DerivedDerived::memfunc2` are both encoded as `M4BaseFviE`.
+- `Derived::memfunc3` and `DerivedDerived::memfunc3` are both encoded as `M7DerivedFvdE`.
+- `Derived::memfunc4` is encoded as `M7DerivedFvvE`.
 
 == Linker Relaxation
 


mylai-mtk:
Regarding the mechanism to pass the calculated lpad labels to static linker for PLT generation, maybe we can just add a symbol for each called function? Currently, the LLVM KCFI mechanism adds similar symbols to allow assembly codes to reference labels computed from C function declarations.

This approach has the benefit of being easily human readable when examining the assembly text and object dumps (through symbol table), and it does not require us to invent a new format for this purpose, which means it can already be accepted by existing assemblers and compiling pipelines that utilizes independent assemblers.

The downside of this may be that it adds quite a lot of additional entries to the symbol table and the data structure of symbol table entries are a bit too bloated for this purpose, but if we decide to pass the symbols along in the relocatables instead of fetching from shared objects at static link time, maybe we can just advise programmers to strip away the symbol tables after linking so the program size can be reduced.

mylai-mtk:
Hi @kito-cheng, thanks for the reply.

> [Build attribute](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#attributes): build attribute has reserve a space to associate attribute to a symbol.

Regarding this option, I don't see any structure that looks like the "build attribute" you mentioned that works with symbols in the "Attributes" section of the linked document. Can you be more specific on what existing format we have that you're referring to?

> *1: IIRC KCFI require full LTO, every symbol should resolve at that stage, so the undefined weak symbol may not a problme in that situation, but I am expert on that, so plz correct me if I am wrong :)

I don't think we need LTO with KCFI. KCFI works much like Zicfilp, except that it's software emulated. In KCFI, the label used at the caller site comes from the called target's apparent signature as seen from caller, which should be available in C programs for most of the time. The only exception I know is K&R-style functions, which technically do not have signatures and accept a plethora of argument type combinations. However, intended K&R-style functions are rare IMHO, and probably should use a special label rule if we are to handle them.

> However option 1 and option 2 will has some problem when dealing with undefined weak symbol*1, there won't have an address associate with that kind of symbol.

I wasn't talking about generating symbol at the same address as the function symbol. In KCFI, the label symbols are associated to the called functions by name, e.g. for a `int foo();` C declaration, it would introduce a symbol with the assembly code `.weak __kcfi_typeid_foo; .set __kcfi_typeid_foo, <label>;`, so I don't think symbol addresses matter here.

What I proposed was to add something like `.weak __zicfilp_label_foo; .set __zicfilp_label_foo, <label>;` for every called function in the relocatables. This way, the relocatables would always contain the required labels for PLT generation, and these labels would be calculated from the function signatures that the callers think they're calling. Given that in a relocatable, we can't have two different global symbols of the same name (can we?), using name to associate the called function and its label should be strong enough. Please correct me if I skip something important here 🙇‍♀️

riscv-elf.adoc
```diff
@@ -1529,6 +1571,61 @@ attribute is recording for minimal execution environment requirements, so the
...
+hash of the function signature string, which uses the same scheme as the
+"Function types" mangling rule defined in the _Itanium {Cpp} ABI_
+<<itanium-cxx-abi>>, the value is taken from the lower 20 bits of the MD5
+hash result of the function signature string.
+
+Additionally, here are a few specific rules for {Cpp} member functions:
+
+- {Cpp} member functions should use the "Pointer-to-member types" mangling rule
+  defined in the _Itanium {Cpp} ABI_ <<itanium-cxx-abi>>.
+- Virtual functions in {Cpp} should use the member function type of the base
```
mylai-mtk, line 1586: I'm not a C++ expert, but I have tried a program where the virtual method override relationship is ambiguous (at least to me): 

```c++
#include <iostream>

using namespace std;

class C1 {
public:
    virtual int f1() {
        cout << "C1::f1\n";
        return 0;
    }
};

class C2 {
public:
    virtual int f1() {
        cout << "C2::f1\n";
        return 0;
    }
};

class C3 : public C1, public C2 {
public:
    int f1() override {
        cout << "C3::f1\n";
        return 0;
    }
};

int main() {
    C3 c3;
    C1 *c1p = &c3;
    C2 *c2p = &c3;
    (c1p->*(&C1::f1))();
    (c2p->*(&C2::f1))();
    return 0;
}
```

On [godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIM6SuADJ4DJgAcj4ARpjEIADMAGykAA6oCoRODB7evv6p6ZkCIWGRLDFxSbaY9o4CQgRMxAQ5Pn5c1bVZDU0EJRHRsXoKjc2teR0jvf1lFRIAlLaoXsTI7Bzm8aHI3lgA1Cbxbk4jxJish9gmGgCC1zdeGUZ7zGwKKUxreyPoh1a39x2TAUCj2bi4BwA7P8bikvFFaHhkCB7ns0XsAG54ZpeMR7UIEPZULgQeZQmHoyl7NBeQmHNz0g5mMzgkAgYkmACsbgY5jMf1RVLRZwIKwYew0AtulJMkIAIvdZQr4jDAQYQWCzOT7nCEUiUdL0VicXiCUSSWTZRShdTlnSjoy%2BW4zGyOdzecypTcbXsRWKJV6ZfLFcGVYrbkCNW54nsQHtdYjkGCOvH4YnNdrbgn9YK0WbiaS9qgMbFiHh9lbc1SafaGUcmSz4q6uFyeXzAza/cRxZKw4a0UqQ8rVbczSwmKFC5X%2B2CY8h4h20eC9gAqZBcFIHeJypmJeeLjNrsybw478x7hd973oiDrlIAWkuK4g59Z7K483mpIPt%2BPj/i2DPq%2BLrvp%2B35XpSXY9l6SocIstCcJyvB%2BBwWikKgnAMpY1jfMsqyYEy8Q8KQBCaHBiwANYJPEAB08T0QxjGMckCEcJIyFkehnC8AoIAaCRZGLHAsBIGgLApHQsTkJQYkSfQcTAFw8QBDQtAELEvEQFEnDEVEoRNAAnjpvB6cwxAGQA8lE2iYA4xmkGJbCCBZDC0EZ3C8FgUReMAbhiLQvEeaQWDjkY4hBfgZwOHgJaBcRmCqLZtLrGhBI1JxiJRMQhkeFg9kEGWLD2SWxBROkmBypgoXAIiRiCXwBjAAoABqeCYAA7hZKSMPZ/CCCIYjsFIMiCIoKjqKhOh6AYdWmNhlj6HgUS8ZAiyoCkdQMIFvDFqW5aYCtpKdLZm0uAw7ieG0ejBKEAzlEMHRpBkm3jO0BTPVkMyDHEkw1Cd3SjC0l0TMd0X1IDX33T9tiA69wwQ7dswPYsCh4WsCz6IhHGTVxHB7KoAAciT3okkh7MAyBJkpNFahAuCECQhEfrwpGTZ%2BpBUfRdFMTz9EsZw7GkChaEYRwPF8QJbOkMJiAgHacIENJt6oOJknEOErDrITxOk%2BTlN7NTZi8Jg%2BBEGW6B6H1wiiOIw1W2NaicboHTtdlKTGfBWNC5xosWbSCtFlQ%2BNEyTZMU1TtG0x4qvyUz8ws4JlHUdzvNMZjbHYyL3G2BLrNaOzrFG97OOiwnbOLCVGTOJIQA%3D), this program prints:

```
C3::f1
C3::f1
```

, which implies that `C3::f1` overrides both `C1::f1` and `C2::f1`, so the problem arises: Which base (`C1` or `C2`) do we use for label calculation?

kito-cheng: Just read "Inside the C++ Object Model" again, seems like the thunk able to let us distinguish between `C3::C1::f1` and `C3::C2::f1`.

Add one more parent to your case:
```cxx
#include <iostream>

using namespace std;

class C1 {
public:
    virtual int f1() {
        cout << "C1::f1\n";
        return 0;
    }
};

class C2 {
public:
    virtual int f1() {
        cout << "C2::f1\n";
        return 0;
    }
};

class CX {
public:
    virtual int f1() {
        cout << "C2::f1\n";
        return 0;
    }
};


class C3 : public C1, public C2, public CX {
public:
    int f1() override {
        cout << "C3::f1\n";
        return 0;
    }
};

int main() {
    C3 c3;
    c3.f1(); // Invoke _ZN2C32f1Ev directly
    C1 *c1p = &c3;
    C2 *c2p = &c3;
    CX *cxp = &c3;
    c1p->f1(); // Invoke *(_ZTV2C3+16) which is _ZN2C32f1Ev
    c2p->f1(); // Invoke *(_ZTV2C3+40) which is _ZThn8_N2C32f1Ev
    cxp->f1(); // Invoke *(_ZTV2C3+64) which is _ZThn16_N2C32f1Ev
    return 0;
}
```

The thunk function is used for adjust this pointer, `_ZThn8_N2C32f1Ev` will adjust `this` pointer to `this + 8` which used for `C2`, and `_ZThn16_N2C32f1Ev` will adjust `this` pointer to `this + 16` which used for `CX`.

So signature for `_ZN2C32f1Ev` is `M2C1FvvE`, `_ZThn8_N2C32f1Ev` is `M2C2FvvE` and `_ZThn16_N2C32f1Ev` is `M2CXFvvE`.

However I need your double check with this...I also not really family with those C++ stuffs...:P

[1] https://stackoverflow.com/questions/18398409/c-inherit-from-multiple-base-classes-with-the-same-virtual-function-name
[2] https://reverseengineering.stackexchange.com/questions/4543/what-is-a-non-virtual-thunk

mylai-mtk: Thanks for pointing out that there may be a thunk to adjust the object pointer before the actual call, and I assume that you want to add lpad insns to thunks to fix the current problem. The question I have now is whether we should rely on this thunk mechanism to adjust our landing pad encoding scheme?

In my example, the core problem is that the existing term ("use the member function type of the base class that first defined the virtual function") may run into a problem where "the base class that first defined the virtual function" is ambiguous, since a virtual method in the child class may override methods coming from different inheritance paths at the same time, making it hard to determine which base class first defined (declared?) the virtual function.

Knowing that there are thunks and other details on how Itanium C++ ABI implements virtual calls may enable us to encode landing pads properly by carefully crafting our rules according to the object layout/thunk rules under all inheritance relationship combinations (virtual/normal inheritance, primary base or not), but coding these into this ABI might go a little too far for the purpose of Zicfilp IMHO, as it would be complex and hard to understand, and effectively mandates the use of the Itanium C++ ABI on object layout and virtual calls.

riscv-elf.adoc
```diff
@@ -1529,6 +1571,61 @@ attribute is recording for minimal execution environment requirements, so the
 ISA information from arch attribute is not enough for the disassembler to
 disassemble the `rv64gcv` version correctly.
 
+== Label Value Compuatation for Complex Labeling Scheme Landing Pad
+
+The label value for the complex labeling scheme landing pad is computed from the
+hash of the function signature string, which uses the same scheme as the
+"Function types" mangling rule defined in the _Itanium {Cpp} ABI_
```
mylai-mtk, line 1578: Do we adopt the [substitution](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangling-compression) rules of the Itanium C++ ABI mangling rules?

riscv-elf.adoc
```diff
@@ -1529,6 +1571,61 @@ attribute is recording for minimal execution environment requirements, so the
 ISA information from arch attribute is not enough for the disassembler to
 disassemble the `rv64gcv` version correctly.
 
+== Label Value Compuatation for Complex Labeling Scheme Landing Pad
```
mylai-mtk, line 1574: How do we encode labels for the main function? The main function is special, since it is a K&R-style function in many programs (in the form of `int main()`), but the C library may assume a particular signature for it, e.g. musl assumes `main` to be `int (*)(int, char **, char **)`. This inconsistency in signatures breaks the start `main` process in libc when complex labels are used, since the main function is usually jumped to with an indirect jump.

deepak0414: Not a solution for other K&R style functions.

But for `main`, it's expected to be called once.
I think `entry point` of program can be without any landing pad.

Since C library is reaching there via a static property, it can use sw guarded jmp/call (using t2)

mylai-mtk: After looking into libc (glibc, musl), I'm not sure if sw guarded jump is suitable here.

For dynamically linked binaries, libc routes the start main procedure back to the static binary entry point, so it's the same case for `main`.

For statically linked binaries, libc obtains the main function pointer by static link time symbol resolution. This pointer is then passed through a few libc initialization functions, and finally indirectly jumped to. My concern here is that the place where the main function pointer is obtained is a bit too far from where it's used, and this leaves space (instructions in between pointer obtainment and usage) for malicious actors to manipulate the pointer.

---

Also, when tracing the libc start main procedure of dynamically linked programs, it came up to me that the `_start` symbol also need a CFI label, since it's indirectly jumped to from the dynamic loader, with its address obtained from the auxiliary vector (`aux[AT_ENTRY]`). For an assembly function like `_start`, I think the only way we can have a label for it is to manually assign one?

riscv-elf.adoc
```diff
@@ -704,7 +708,25 @@ And occupies three 16 byte entries for the simple landing pad PLT style:
...
+----
+1:  lpad 0
+    auipc  t2, %pcrel_hi(.got.plt)
+    sub    t1, t1, t3               # shifted .got.plt offset + hdr size + 24
+    l[w|d] t3, %pcrel_lo(1b)(t2)    # _dl_runtime_resolve
+    addi   t1, t1, -(hdr size + 24) # shifted .got.plt offset
+    addi   t0, t2, %pcrel_lo(1b)    # &.got.plt
+    srli   t1, t1, log2(32/PTRSIZE) # .got.plt offset
+    l[w|d] t0, PTRSIZE(t0)          # link map
+    jr     t3
```
mylai-mtk, line 723: This place does not set X7 to a label. Does this imply that the official label of `_dl_runtime_resolve()` is `0`?

riscv-elf.adoc
kito-cheng: I incline define a special rule for main - fix the signature to `int main(int argc, char *argv[])` which defined in C23 spec.

riscv-elf.adoc
kito-cheng: Yes, and I think it worth to mention in the spec :)

riscv-elf.adoc
kito-cheng: We need to discard the class name from the signature _**IF**_ we don't handle this, so I would like to add rule to the spec, it's complicated for human, but it's not too hard to compiler :P we already have those thunk stuffs in compiler.

mylai-mtk:
We need to have a way (define macro?) to know which labeling scheme (simple/complex) is in use for the current compilation, so assembly files can know which label to use. This is needed for libc implementation.

mylai-mtk:
(Nitpicking...) I propose that we move away from the name of "complex" when using function signatures as the label content. Though we have a "simple" labeling scheme and it's natural to use its opposite term "complex" to name our new scheme, which is not simple, I think the term "complex" covers too many possibilities and thus reveals too little about what it actually is. Also, avoiding the "complex" term allows us to define more label schemes in the future without the embarrassment of feeling like defining another new "complex" scheme. Based on these points, I propose to name this current new label scheme using the name "func_sig" (the underscore may be removed), which is precise and tells what it really does.

kito-cheng:
> (Nitpicking...) I propose that we move away from the name of "complex" when using function signatures as the label content. Though we have a "simple" labeling scheme and it's natural to use its opposite term "complex" to name our new scheme, which is not simple, I think the term "complex" covers too many possibilities and thus reveals too little about what it actually is. Also, avoiding the "complex" term allows us to define more label schemes in the future without the embarrassment of feeling like defining another new "complex" scheme. Based on these points, I propose to name this current new label scheme using the name "func_sig" or "mangled_sig" (the underscore may be removed), which is precise and tells what it really does.

Good suggestion,  let me rename it to function signature / func_sig :)

kito-cheng:
Create a PR for adding macro https://github.com/riscv-non-isa/riscv-c-api-doc/pull/76

kito-cheng:
Changes:
- Rename complex labeling scheme to function signature based labeling scheme
- Fix the PLT stubs
- Add labeling rule for `main` and `_dl_runtime_resolve`.
- Clarify the rule for those virtual function from more than one base class.
- Add @mylai-mtk  as coauthor since he has contributes many useful feedback.

riscv-elf.adoc
kito-cheng: Address :)

riscv-elf.adoc
kito-cheng: Added rule for main

riscv-elf.adoc
kito-cheng: Added

riscv-elf.adoc
```diff
@@ -727,6 +750,19 @@ The code sequences of the PLT entry for the the simple landing pad PLT style:
     jalr    t1, t3
 ----
 
+The code sequences of the PLT entry for the the function signature based landing pad PLT style:
+[,asm]
+----
+1:  lpad    <hash-value-for-function>
```
mylai-mtk, line 756: It suddenly came up to me: Do we really need this `lpad`?

+ For direct calls, this `lpad` does not serve any purpose, as this program point is reached with direct jumps
+ For indirect calls, the function's taken address would refer to the actual function instead of this PLT entry, so all indirect calls through that address would not go through this `lpad`
+ For `dlsym` stubs, it looks like the actual function's addresses instead of PLT entries are returned by `dlsym()`, so this `lpad` would not be reached if a function is called through `dlsym()`

So... it looks like in all compiler-generated cases, this `lpad` is useless and redundant? Am I missing something? 🤔

aswaterman: > For indirect calls, the function's taken address would refer to the actual function instead of this PLT entry, so all indirect calls through that address would not go through this lpad

Alas, this is not always true.  If everything were compiled as PIC, then I believe you'd be right.

But if a non-PIC part of the program takes the address of a function that it calls through the PLT, it'll get the address of the PLT entry, rather than the runtime address of the function.  (And then, to ensure function pointer equality, even PIC parts of the program must get the address of the PLT entry if they take the address of the same function... but only if the non-PIC part took the address at some point.)

I think we have enough information at PLT-generation time to omit the landing pad, since at that point we know if anyone might indirectly call via the PLT.

riscv-elf.adoc
```diff
@@ -727,6 +750,19 @@ The code sequences of the PLT entry for the the simple landing pad PLT style:
...
+    auipc   t3, %pcrel_hi(function@.got.plt)
+    l[w|d]  t3, %pcrel_lo(1b)(t3)
+    lui     t2, <hash-value-for-function>
```
aswaterman, line 759: Doesn't `t2` already contain the hash value at this point?  (I'm assuming the PLT and the real function can have the same hash value.)  If that's the case, we can shrink the PLT entries back down to 16 bytes, right?

ved-rivos: Its needed for direct call to PLT followed by indirect tail call from PLT to target...

aswaterman: Oops. Right.

riscv-elf.adoc
```diff
@@ -662,6 +662,7 @@ using all other PLT sytle.
 |===
 | Default PLT                  | -
 | Simple landing pad PLT | Must use this PLT style when `GNU_PROPERTY_RISCV_FEATURE_1_CFI_LP_SIMPLE` is set.
```
ved-rivos, line 664: Maybe we could call it "LP_SINGLE" or "LP_NONE" instead of "LP_SIMPLE".. to indicate these are unlabeled landing pads?

riscv-elf.adoc
aswaterman: Furthermore, if it's known at static link time that none of the PLTs need the landing pad because the address of a PLT entry is never taken, then all of the AUIPCs could be removed, allowing the PLTs to shrink back to 16 bytes in this common case.  Right @ved-rivos?

ved-rivos: Yes, that should be possible. We also have a [proposal](https://github.com/riscv/riscv-cfi/issues/157) for linker relaxations for the ld to prune other landing pads as well.

mylai-mtk: Thanks for pointing these out! 🙇‍♀️

riscv-elf.adoc
```diff
@@ -1529,6 +1572,77 @@ attribute is recording for minimal execution environment requirements, so the
 ISA information from arch attribute is not enough for the disassembler to
 disassemble the `rv64gcv` version correctly.
 
+== Label Value Compuatation for Function Signature based Scheme Landing Pad
+
+The label value for the function signature based labeling scheme landing pad is computed from the
+hash of the function signature string, which uses the same scheme as the
+"Function types" mangling rule defined in the _Itanium {Cpp} ABI_
```
mylai-mtk, line 1579: I've run into a problem when testing my prototype toolchain that implements this rule. The problem is that some data types have different definitions in C vs C++, and this makes labels incompatible when a function is defined in one language, but called from the other.

The problematic type I encountered is the `wchar_t` type, which is defined by C to be

> an integer type defined in the [<stddef.h>](http://port70.net/%7Ensz/c/c11/n1570.html#7.19) header

, i.e. a `typedef` that is usually defined to be `int`, while in C++: 

> Type wchar_t is a distinct type whose values can represent distinct codes for ...

, which means a built-in type and keyword. This means when a source code is interpreted as a C language file, `wchar_t` is mangled to `i` (since it's usually typedef-ed to be `int`), while when in C++ it's mangled to `w` (the built-in `wchar_t` type).

ved-rivos: Should we state that for purpose of generating the hash the mangling follows C definition?

mylai-mtk: > Should we state that for purpose of generating the landing pad hash the mangling scheme follows the C definition?

I don't think this is (easily) feasible for compilers due to these reasons:
1. The C standard does not define `wchar_t` to be a specific type, so compilers would need to parse the system header to know what it really is. This almost means instantiating a new C context, and compile the header itself.
2. I'm afraid parsing the system header won't be enough:
2.1. Libc headers are commonly subject to user-defined macros like `_POSIX_SOURCE`, which changes behaviors of headers. This means the headers can't be parsed independently, but have to be parsed in the context of the user source file, which may be C++
2.2. The libc headers from glibc and musl both masks the definition of `wchar_t` if `__cplusplus` is defined, which implies that we have to mask `__cplusplus` macro
2.3. Furthermore, changing commonly relied on macros like `__cplusplus` may change the contents of the user source file. Users may define more or less types/functions depending on `__cplusplus`, which would bring hazards when we map the labels from the C version AST back to C++
2.4 Moreover, in C++, user do not need to include <stddef.h> to use `wchar_t`, so when re-parsing the user source code as C, compilers may need to include the necessary headers automatically. I think if a user defines `_POSIX_SOURCE`, preprocessors would have no way to know if parsing the additional headers with or without user's `_POSIX_SOURCE` definition would be appropriate
2.5 Note: I don't have examples of macros changing the definition of `wchar_t` so far, so I'm just raising a hypothetic problem with this whole (2.)

In a nut shell, considering what's achievable by the preprocessor, to support this proposal would mean a huge mess for compiler implementation.

Note: Forcing C to be C++ may be easier though, with these problems in my mind:
1. I'm not sure if getting a C++ `wchar_t` type reference from a C AST context would break any compiler assertion or bring any problem
2. I suppose forcing a C typedef-ed `wchar_t` to be a C++ built-in `wchar_t` should be implemented by type name matching, but I don't think this is a good thing to do. Users may have:

```c
// a.c
typedef int wchar_t;
void foo(wchar_t);

// b.c
typedef unsigned int wchar_t;
void bar(wchar_t);
```

Using type name matching means that `foo()` and `bar()`, despite taking different parameters, would have the same label.

mylai-mtk: I think Kito's more limited version of treating some types as C types may be more favorable, but we can simplify things a bit.

> so we could make it more explicit to use int in the psABI spec

I don't think it's necessary to fix `wchar_t` to `int` for compiled code, given that the status quo is acceptable.

> state wchar_t should use same mangling scheme as C

I think we don't need to first fix `wchar_t` to `int` in C, then say that all `wchar_t` are treated like a C type. We can just say all `wchar_t` are treated as `int` for the purpose of label generation. This should suffice.

jrtc27: > both RISC-V GCC[2] and RISC-V LLVM[3] are implement it as int

FreeBSD's ABI is different, whether deliberate or a historical accident.

kito-cheng: > FreeBSD's ABI is different, whether deliberate or a historical accident.

I saw FreeBSD in RISC-V GCC is still using `int`? Does it still the case for RISC-V FreeBSD? 

Anyway I am not intend to force it as `int`, I would try to say it should use same type as C in psABI, explicitly document Linux target use `int`, and mention different target may have different definition.

https://github.com/gcc-mirror/gcc/blob/master/gcc/config/riscv/freebsd.h#L21

kito-cheng: >> state wchar_t should use same mangling scheme as C
>
> I think we don't need to first fix wchar_t to int in C, then say that all wchar_t are treated like a C type. We can just say all wchar_t are treated as int for the purpose of label generation. This should suffice.

It's part of platform ABI, it will cause problem if user using different type (w/ different size) and then invoke function provided from `wchar_t`.

jrtc27: Hm, I could have sworn one of those types differed between FreeBSD and Linux... not sure what I'm remembering then.

riscv-elf.adoc
```diff
@@ -1529,6 +1572,77 @@ attribute is recording for minimal execution environment requirements, so the
...
+the MD5 hash result of the function signature string.
+
+Additionally, here are a few specific rules:
+
+- `main` funciton is using signature of
+   `(int, pointer to pointer to char) returning int` (`FiiPPcE`).
+- `_dl_runtime_resolve` use zero for the landing pad.
+- {Cpp} member functions should use the "Pointer-to-member types" mangling rule
+  defined in the _Itanium {Cpp} ABI_ <<itanium-cxx-abi>>.
+- Virtual functions in {Cpp} should use the member function type of the base
```
mylai-mtk, line 1591: TLDR; At the core of this problem is that we have no idea whether a pointer to member function would contain a concrete member method, or a thunk, and this hinders label generation at indirect call sites given the current label generation rule mandates different classes to be encoded in these 2 cases.

Consider the following case:

```c++
class C1 { virtual void f1(); };
class C2 { virtual void f2(); };
class C3: public C1, public C2 {}; // C1 is the primary base
```

Then we have 2 cases:

1. `void (C3::*p)() = C3::f1;`: In this case, we need to make the indirect call with label signature `void (C1::*)()`. The virtual function pointer stored in `C3`'s virtual table is `C1::f1` without `this` pointer adjustment, since `f1` is last defined in `C1`, which is the primary base of `C3`. For a non-adjusting entry point of a virtual function, we encode the label with the class that first declared the function, which is `C1`.
2. `void (C3::*p)() = C3::f2;`: In this case, we need to make the indirect call with label signature `void (C3::*)()`, since `f2` is last defined in `C2`, which is a secondary base, so a thunk adjusting the `this` pointer from `C3` to `C2` is needed, and by current draft spec, we encode `void (C3::*)()` in the label for a thunk like this.

These cases bring up the problem: The class needed for label encoding at a single indirect call site cannot be determined.

(Note: I just finished the Itanium C++ ABI on object layout and virtual table generation for the first time, so there may be misunderstandings in my arguments. Please correct me if the Itanium ABI does not mean what I said. 🙇‍♀️)

mylai-mtk: I think the same problem also exist for non-virtual construct:

```c++
class C1 {
public:
    void f1() {}
};

class C2 {
public:
    void f2() {}
};

class C3: public C1, public C2 {};

int main() {
    void (C3::*p)() = &C3::f1;
    C3 c3;
    (c3.*p)();
}
```

mylai-mtk: One way I can think of to workaround this is to change how pointer to function member is represented, and carry the correct label along with the pointer when it's assigned a concrete method.

kito-cheng: Thanks for reporting that issue, I suspect we may...consider drop the class name from the signature, and put a prefix into the signature string e.g. we encode the class name into the string before like `M4BaseFvvE`, and it may become `MFvvE` to indicate that's member function which come with a this pointer.

riscv-elf.adoc
```diff
@@ -1529,6 +1572,77 @@ attribute is recording for minimal execution environment requirements, so the
...
+== Label Value Compuatation for Function Signature based Scheme Landing Pad
+
+The label value for the function signature based labeling scheme landing pad is computed from the
+hash of the function signature string, which uses the same scheme as the
+"Function types" mangling rule defined in the _Itanium {Cpp} ABI_
+<<itanium-cxx-abi>>, and the function signature will use the "Compression" rule
+defined in  _Itanium {Cpp} ABI_, the value is taken from the lower 20 bits of
+the MD5 hash result of the function signature string.
+
+Additionally, here are a few specific rules:
```
mylai-mtk, line 1584: We may need a special rule for destructors of global/static objects that get registered with `__cxa_atexit()`.

Destructors of global/static objects would be registered with `__cxa_atexit(<destructor_func_ptr>, <obj_ptr>, <dso_handle>)` after construction, and are expected to be called at program exit/dso unload with indirect call `(void (*)(void*))<destructor_func_ptr>((void*)<obj_ptr>)`.

Maybe we can force all destructors to use the label `void (*)(void*)` to solve this problem? Destructors are rarely called indirectly, except the virtual ones.

riscv-elf.adoc
mylai-mtk, line 1579: > <[function-type](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.function-type)> ::= [<[CV-qualifiers](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.CV-qualifiers)>] [<[exception-spec](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.exception-spec)>] [Dx] F [Y] <[bare-function-type](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.bare-function-type)> [<[ref-qualifier](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.ref-qualifier)>] E

In the \<function-type\> mangling rule, there's an optional component for \<exception-sepc\>, which exists under the condition

> When an exception-specification (i.e., noexcept, noexcept(expression), or throw(type(s))) is part of the function type, it is mangled according to <[exception-spec](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.exception-spec)> as described below.  -- Itanium C++ ABI

However, according to [noexcept specifier (since C++11) (en.cppreference.com)](https://en.cppreference.com/w/cpp/language/noexcept_spec), the noexcept-specification is not a part of the function type until C++17, and it is a part of the function type since C++17. This means that whether the \<exception-spec\> component appears depends on the C++ version. Is this what we want?

I'm asking because I suppose that this would affect the case where a post-C++17 project needs to use another library compiled with pre-C++17 standard.

kito-cheng: Thanks for catching this, I tried following program, which build-able with -std=c++17 and above, but failed with c++11...so yes, I think we should omit `<exception-spec>`.
```
void g(void pfa() noexcept){};
void g(void pfa()){};
```

mylai-mtk: I think I have run into this problem with a pretty common toolchain configuration during my test:

I have the LLVM `libc++.so.1.0` built using post-C++17 standard, which I suppose is the default configuration since I did not specify any C++ standard in my build script, and my test app code is built using C++14, and linked to the just-built `libc++.so.1.0`. 
The problematic function is the `void operator delete(void *) noexcept`, which has the mangled name of `_ZdlPv`, but the labels contained in `libc++.so.1.0` comes from `DoFvPvE`, while the one in the test app code has `FvPvE`. The difference `Do` is indeed, the \<exception-spec\> part.

riscv-elf.adoc
kito-cheng: Sounds reasonable suggestion, let me update that next work day :)

And that remind me we have function with destructor attribute, but in that case we may just use sw-guarded jump.

riscv-elf.adoc
kito-cheng: Maybe LP_ZERO or LP_ZERO_PAD? LP_NONE sounds like no landing pad, LP_SINGLE sounds like could be any value other than 0 but just fixed to a value.

ved-rivos: LP_UNLABELED?

kito-cheng: LP_UNLABELED sounds great :)

riscv-elf.adoc
```diff
@@ -1529,6 +1572,77 @@ attribute is recording for minimal execution environment requirements, so the
...
+
+- `main` funciton is using signature of
+   `(int, pointer to pointer to char) returning int` (`FiiPPcE`).
+- `_dl_runtime_resolve` use zero for the landing pad.
+- {Cpp} member functions should use the "Pointer-to-member types" mangling rule
```
mylai-mtk, line 1589: I guess we need to mention that static member methods should be treated like normal standalone functions?

riscv-elf.adoc
mylai-mtk, line 1591: Virtual functions allow the return type to be a compatible type of that of the overridden function, and this return type is mangled into the label according to the \<function-type\> mangle rule. This means overriders with a different return type would fail on virtual function calls through base class pointers, and we need to fix this.

To fix this, I believe we need to drop the return type when mangling virtual function types, since it would have a similar problem of indeterminable indirect call site signature mentioned in [another comment](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/434#discussion_r1609172934). To be more specific, consider the following case:

```c++
// Note that A is the primary base of B, so the thunk to adjust the returned pointer is not needed.
struct A {
  virtual A* f();
};

struct B : public A {
   virtual B* f() override;
}
```

+ Case 1: To support `B b; A *pa = &b; pa->f();`, we need `B::f` to mangle its return type as `A*`
+ Case 2: To support `B* (B::*pbf)() = &B::f; (B().*pbf)();`, we need `B::f` to mangle its return type as `B*`

These two cases require contradicting return types to be mangled, so I think we need to drop the return type in the mangled label.

ved-rivos: Could this exclusion of return type be restricted to the cases where such override is allowed. That would avoid loosening the check for virtual functions which do not need that relaxation.

mylai-mtk: > Could this exclusion of return type be restricted to the cases where such override is allowed

I believe we have to compromises all virtual functions, since in case 2, at compile time, we could have no idea whether the return type `X*` is an overriding compatible type due to the possibility that `X` can be forward declared.

mylai-mtk:
Regarding functions having no prototype, which are still widely but probably unintendedly used in new C codes, should we impose some restrictions on the usage of them so that we can properly label them? To be specific, I propose that we treat types like `float foo()`, which seemingly take no parameter, as `float foo(void)` when calculating labels, and prohibit calling pointers like `float (*)()` with arguments to avoid accepting existing ancient code bases. I believe this is what modern programmers mean when declaring a function with nothing in the parameter list.

kito-cheng:
Changes:
- Special rule for return type of member function.
- Special rule for class destructors
- `<exception-spec>` should be ignored.
- Static functions should follow the same rules as normal functions.
- `wchar_t` is platform dependent.
- Functions with an empty parameter list are treated as `void` (`v`).

riscv-elf.adoc
```diff
@@ -1529,6 +1572,96 @@ attribute is recording for minimal execution environment requirements, so the
...
+- `_dl_runtime_resolve` uses zero for the landing pad.
+- {Cpp} member functions should use the "Pointer-to-member types" mangling rule
+  defined in the Itanium C++ ABI <<itanium-cxx-abi>> with the following
+  additional rules:
+  - Virtual functions in {Cpp} should use the member function type of the base
+    class that first defined the virtual function.
+  - If a virtual function is inherited from more than one base class, it should
+    use the type of the first base class. Thunk functions will use the type of
+    the corresponding base class.
+  - The return type of a class member function should mangle to `void *` if it
```
mylai-mtk, line 1597: I wish we can use the term "covariant return type" to describe the case. The term is commonly used to describe the overriding relationship between the return types of base and derived functions according to my Google search results. Also, it's the term used in [virtual function specifier - cppreference.com](https://en.cppreference.com/w/cpp/language/virtual#Covariant_return_types), which should make it formal enough to be used in a specification.

However, using the term here is a bit too restrictive, as compilers do not need to know (and cannot know in the forward declaration case) if the return type is a covariant return type when calculating the label due to fact that we're degrading all possible covariant return type cases to `void *`, regardless of whether it's really a method with a covariant return type as this fact may not be determinable when compiling a single translation unit. Considering this approach, I would recommend just incorporating the part of covariant return type definition that we truly need:

> both types are pointers or references (lvalue or rvalue) to classes. Multi-level pointers or references are not allowed.

to this specification, in order to be precise and aligned to other C++ documents when describing the desired case that we want to apply this rule to, and add a note to [virtual function specifier - cppreference.com](https://en.cppreference.com/w/cpp/language/virtual#Covariant_return_types) to introduce a more complete story to readers.

mylai-mtk, line 1597: This rule regarding covariant return types should only affect virtual functions?

kito-cheng: Also apply for non-virtual function since member function pointer could take from both virtual function and non-virtual function

mylai-mtk: Yes, member function pointer could take from both virtual function and non-virtual function. But considering that calls through member function pointers actually have a conditional branch that determines whether the contained function "descriptor" is a virtual table index or a normal function pointer, and act differently according to the result, we can in fact contain this rule to virtual methods.

kito-cheng: Added one note to describe that.

kito-cheng:
Changes:
- Tweak the rule for the return type to virtual function only.
- Tweak the rule for virtual functions

riscv-elf.adoc
kito-cheng: Yeah, I was struggle on we should not have diverge rule between virtual function and non-virtual function, but we already did, so yeah, taken your suggestion :)

For reg pressure: The live range is short so I think it's OK.

riscv-elf.adoc
kito-cheng: New rule for the virtual function should address this
```
+  - Virtual functions should use `v` for `<class type>` rather than the actual
+    class name.
```

mylai-mtk: > New rule for the virtual function should address this
> 
> ```
> +  - Virtual functions should use `v` for `<class type>` rather than the actual
> +    class name.
> ```

What about the non-virtual case?

riscv-elf.adoc
```diff
@@ -1529,6 +1572,95 @@ attribute is recording for minimal execution environment requirements, so the
 ISA information from arch attribute is not enough for the disassembler to
 disassemble the `rv64gcv` version correctly.
 
+== Label Value Compuatation for Function Signature based Scheme Landing Pad
```
mylai-mtk, line 1575: > \<function-type\> ::= [<[CV-qualifiers](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.CV-qualifiers)>] [<[exception-spec](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.exception-spec)>] [Dx] F [Y] <[bare-function-type](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.bare-function-type)> [<[ref-qualifier](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.ref-qualifier)>] E

Do we mangle the 'Y' component? It looks like the existence of this component depends on compiler implementation:

> A "Y" prefix for the bare function type encodes extern "C" in implementations which distinguish between function types with "C" and "C++" language linkage. -- Itanium C++ ABI

, and it looks like currently [GCC](https://github.com/gcc-mirror/gcc/blob/96fe23eb8a9ebac6b64aeb55db88d219177a345a/gcc/cp/mangle.cc#L2978) and [LLVM](https://github.com/llvm/llvm-project/blob/1d4e857acdd93a659387b9799fa262cce2370fb0/clang/lib/AST/ItaniumMangle.cpp#L3551) do not produce this component.

riscv-elf.adoc
kito-cheng: oh yeah, it should apply to non-virtual case as well

riscv-elf.adoc
```diff
@@ -1529,6 +1572,95 @@ attribute is recording for minimal execution environment requirements, so the
...
+
+Additionally, here are a few specific rules:
+
+- `main` funciton is using signature of
+   `(int, pointer to pointer to char) returning int` (`FiiPPcE`).
+- `_dl_runtime_resolve` uses zero for the landing pad.
+- {Cpp} member functions should use the "Pointer-to-member types" mangling rule
+  defined in the Itanium C++ ABI <<itanium-cxx-abi>> with the following
+  additional rules:
+  - Virtual functions should use `v` for `<class type>` rather than the actual
```
mylai-mtk, line 1592: I wonder if you mean replace the \<*class* type\> component with `v` directly, or use a dummy class type named `v` to mangle the class type? I.e. use plain `v` or use an unscoped name (`1v`) for the component?

kito-cheng: M1v, and I realize that's might be ambiguous, let me try to fix that :P

riscv-elf.adoc
kito-cheng: Let me drop that explicitly

riscv-elf.adoc
```diff
@@ -1529,6 +1572,95 @@ attribute is recording for minimal execution environment requirements, so the
...
+    class name.
+  - The return type of a virtual class member function should mangle to
+    `void *` if it is a pointer or reference to a non-primitive type. A pointer
+    to a pointer of a non-primitive type is not included in this rule.
+  - Class destructors should use the signature `void (*)(void*)` (`FvPvE`).
+  - `<exception-sepc>` should be ignored.
+  - Static function is following the rule as non-member function.
+- `wchar_t` should match the type of the target platform. For example, on
+   Linux, it uses `int`, so it mangles to `i` rather than `w` for {Cpp}.
+- Function with an empty parameter list are treated as `void` (`v`).
```
mylai-mtk, line 1602: Wording suggestion:

> - Functions with an empty parameter list are treated as explicitly declaring taking no parameters (having `void` as the parameter list)

riscv-elf.adoc
```diff
@@ -1529,6 +1572,95 @@ attribute is recording for minimal execution environment requirements, so the
...
+- {Cpp} member functions should use the "Pointer-to-member types" mangling rule
+  defined in the Itanium C++ ABI <<itanium-cxx-abi>> with the following
+  additional rules:
+  - Virtual functions should use `v` for `<class type>` rather than the actual
+    class name.
+  - The return type of a virtual class member function should mangle to
+    `void *` if it is a pointer or reference to a non-primitive type. A pointer
+    to a pointer of a non-primitive type is not included in this rule.
+  - Class destructors should use the signature `void (*)(void*)` (`FvPvE`).
+  - `<exception-sepc>` should be ignored.
```
mylai-mtk, line 1598: This rule should apply to all functions (not just the class methods), so it should be moved up a level.

riscv-elf.adoc
```diff
@@ -1374,29 +1543,137 @@ A number of symbols, named mapping symbols, describe the boundaries.
...
+
+- `main` funciton is using signature of
+   `(int, pointer to pointer to char) returning int` (`FiiPPcE`).
+- `_dl_runtime_resolve` uses zero for the landing pad.
+- 'Y' component in the `<function-type>` should be ignored.
+- `<exception-sepc>` should be ignored.
+- {Cpp} member functions should use the "Pointer-to-member types" mangling rule
+  defined in the Itanium C++ ABI <<itanium-cxx-abi>> with the following
+  additional rules:
+  - Member functions should use `v` for `<class type>` rather than the actual
```
mylai-mtk, line 1590: It should clearly state that `v` is an unscoped class to to avoid the misinterpretation of using nested `class v` in place of nested `class foo`.

riscv-elf.adoc
```diff
@@ -1374,29 +1543,137 @@ A number of symbols, named mapping symbols, describe the boundaries.
...
+    class name, e.g. `M1v` rahter than `M3foo` for `class foo`.
+  - The return type of a virtual class member function should mangle to
+    `void *`/`void &` if it is a pointer/reference to a non-primitive
```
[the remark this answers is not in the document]
kito-cheng: Yeah, it's not legal in C++, but it's OK to encoding that to `Rv`, my intention is let use those two situation has different signature.

kito-cheng:
Updated for addressing @mylai-mtk's comment and also minor revise for wording.

riscv-elf.adoc
```diff
@@ -1394,6 +1568,115 @@ is not enough for the disassembler to disassemble the `rv64gcv` version
...
+- {Cpp} member functions should use the "Pointer-to-member types" mangling rule
+  defined in the Itanium C++ ABI <<itanium-cxx-abi>> with the following
+  additional rules:
+  - Member functions should use `v` for `<class type>` rather than the actual
+    class name. For example, use `1v` instead of `3foo` for the `<class type>`
+    in `class foo`.
+  - The return type of a virtual class member function should mangle to
+    `void *`/`void &`(`Pv`/`Rv`) if it is a pointer/reference to a non-primitive
+    type. A pointer to a pointer of a non-primitive type is not included in
+    this rule.
```
mylai-mtk, line 1596: Covariant return types also allow rvalue references. Would you also distinguish the case?

riscv-elf.adoc
kito-cheng: Addressed https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/417/commits/a2ce8205d76edf8cb90476414e9fe0676ca0ad09

riscv-elf.adoc
kito-cheng: I change the rule a little bit, use `class v` rather than `void *` `void &`  or `void &&` :P

riscv-elf.adoc
```diff
@@ -1498,6 +1541,115 @@ is not enough for the disassembler to disassemble the `rv64gcv` version
...
+
+- `main` function uses the signature of
+   `(int, pointer to pointer to char) returning int` (`FiiPPcE`).
+- `_dl_runtime_resolve` uses zero for the landing pad.
+- The 'Y' component in the `<function-type>` should be ignored.
+- `<exception-sepc>` should be ignored.
+- {Cpp} member functions should use the "Pointer-to-member types" mangling rule
+  defined in the Itanium C++ ABI <<itanium-cxx-abi>> with the following
+  additional rules:
+  - Member functions should use `v` for `<class type>` rather than the actual
```
mylai-mtk, line 1563: I wonder if this usage of the dummy `class v` as `<class type>` for all member functions applies to all components of the label? E.g.

+ Consider `void C::foo(int (C::*ptr)())`: Do we encode it as `void v::foo(int (C::*ptr)())` or `void v::foo(int (v::*ptr)())`? 
+ Consider `void C::bar(int C::*ptr)`: Do we encode it as `void v::bar(int C::*ptr)` or `void v::bar(int v::*ptr)`? 

For our purpose of avoiding label conflicts when calling through member function pointers, encoding as `void v::foo(int (C::*ptr)())` and `void v::bar(int C::*ptr)` should already suffice.

riscv-elf.adoc
```diff
@@ -1498,6 +1541,115 @@ is not enough for the disassembler to disassemble the `rv64gcv` version
...
+    class name. For example, use `1v` instead of `3foo` for the `<class type>`
+    in `class foo`.
+  - The return type of a virtual class member function should mangle as
+    `class v` rather than the original type name if it is a pointer or reference
+     to a non-primitive type. A pointer to a pointer of a non-primitive type
+     is not included in this rule.
```
mylai-mtk, line 1569: Wording suggestion:

> The return type of a virtual class member function, if it is a pointer or reference to a class, should have its class type mangled as `class v` rather than the declared class type. Multi-level pointers or references are exempted.


This is to:
+ Be clear that it's just the class type-part of the return type that gets overridden with `class v`, not the whole return type
+ Avoid the use of the ambiguous term "primitive type". This term is actually undefined in the [C++ type system](https://en.cppreference.com/w/cpp/language/type)
+ Be explicit that multi level references do not apply this rule

riscv-elf.adoc
kito-cheng: Ack, generally I agree with you, but I don't have time to revise this until this Thursday :P

riscv-elf.adoc
mylai-mtk, line 1569: We may need to strip the CV-qualifiers of return types when mangling for virtual functions, since they can be different:

> the return type of Derived::f() must be equally or less [cv-qualified](https://en.cppreference.com/w/cpp/language/cv) than the return type of Base::f().

riscv-elf.adoc
```diff
@@ -696,7 +700,26 @@ And occupies three 16 byte entries for the simple landing pad PLT style:
     nop
 ----
 
-Subsequent function entry stubs in the PLT take up 16 bytes.
+The function signature based landing pad PLT style occupies two 32 byte entries:
```
mylai-mtk, line 703: Is it a mistake to say it's "two 32 byte entries"? The following code sequence occupies just 48 bytes.

kito-cheng:
Changes:
- Apply @mylai-mtk 's comment 
  - Update rule for co-variant return type per 
  - Update rule for member pointer
  - Fix typo in PLT stuffs
- Adding rule to address hash result is zero

riscv-elf.adoc
```diff
@@ -1498,6 +1541,122 @@ is not enough for the disassembler to disassemble the `rv64gcv` version
...
+
+The label value for the function signature-based labeling scheme landing pad is
+computed from the hash of the function signature string, which follows the same
+scheme as the "Function types" mangling rule defined in the _Itanium {Cpp} ABI_
+<<itanium-cxx-abi>>. The function signature will also use the "Compression" rule
+defined in the _Itanium {Cpp} ABI_.
+
+The label value is derived from the lower 20 bits of the MD5 hash result of the
+function signature string. If the lower 20 bits are all zeros, the higher 20
+bits are used. If all 32 bits are zeros, the lower 20 bits of the MD5 hash
```
mylai-mtk, line 1554: MD5 results in a 128-bit number, so I guess you mean 'If all 128 bits are zero' here.
But since MD5 gives a 128-bit number, would you consider taking other parts of the number if both `hi20(MD5)` and `low20(MD5)` are zero?

kito-cheng: Oh yeah, let me update the rule, I just thought MD5 is 32 bit, but that should be 128 bits...

riscv-elf.adoc
```diff
@@ -1550,9 +1550,11 @@ scheme as the "Function types" mangling rule defined in the _Itanium {Cpp} ABI_
 defined in the _Itanium {Cpp} ABI_.
 
 The label value is derived from the lower 20 bits of the MD5 hash result of the
-function signature string. If the lower 20 bits are all zeros, the higher 20
-bits are used. If all 32 bits are zeros, the lower 20 bits of the MD5 hash
-result of the string "RISC-V" are used.
+function signature string. If the lower 20 bits are all zeros, use the next
+20 bits, and continue using the next 20 bits until a non-zero value is obtained.
+If less than 20 bits are available in the final segment, the highest 20 bits of
+the MD5 hash result will be used. If all 128 bits are zeros, the lower 20 bits
```
mylai-mtk, line 1556: Would you mind if we use the lowest 8 bits (and zero-extend it to 20 bits) in the final segment in case the lowest 120 bits are all zero? This saves an additional 12-bit logical left shift if the following algorithm is used to implement this paragraph:

```c
uint128_t MD5 = ...;
while (MD5) {
 if (MD5 & 0xFFFFF) return MD5 & 0xFFFFF;
 MD5 >>= 20;
}
```

kito-cheng: Yeah, I think this not make much difference, let me update that later :P

deepak0414:
Kito, I and some other folks have discussed another approach for finer grained forward cfi which would impact PLT generation specifics here. Want to ensure that folks here are also aware.

One major point
- I think we can hide behind BIND_NOW and Full RELRO. This simplifies routing through PLT because PLT relies on readonly func pointer load now and thus can safely use software guarded jump. Thus we can avoid label setup.

More details here
https://docs.google.com/document/d/1_2hI-CptJtbuEmQzYMG_RX1QjW2Hg17nLx5NG0Onhws/edit?tab=t.0#heading=h.ty4ca8pkhh6p

Please take a look and poke holes

PR #436, opened 2024-05-02
Fix rendering of vector type tables

craigblackmore:
Table 2 does not render for me in github.  Delimiting the end of the table resolves this.

Also add newline after Table 2 header row so that it appears in bold like in Table 1.

### DIFF
--- a/gen_vector_type_infos.py
+++ b/gen_vector_type_infos.py
@@ -81,7 +81,7 @@ print (".Type sizes and alignments for vector data types")
 print ("[cols=\"4,3,>3,>2\"]")
 print ("[width=80%]")
 print ("|===")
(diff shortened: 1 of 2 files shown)


kito-cheng:
`vector_type_infos.adoc` is generated file, could you fix that on [gen_vector_type_infos.py](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/gen_vector_type_infos.py)?

Thanks :)

craigblackmore:
New changes:
* Fixed Table 1 header.
* Updated `gen_vector_type_infos.py` and regenerated.

kito-cheng (approved):
Thanks!

Issue #430, opened 2024-03-12
Interpretation of floating-point types


aswaterman:
The ABI indicates that `float` is 4 bytes and `double` is 8 bytes, etc., but we do not actually say that `float` is IEEE 754 binary32, `double` is IEEE 754 binary64, etc.

I think this spec is the right place for that, rather than the C API, because it is at least conceivable that alternate RISC-V C APIs supported by different ABIs would be defined that specify e.g. `long double` as an 8-byte type.  But I'll leave it to the experts to make that decision.

Note that it does _not_ suffice to say that the C implementation conforms to Annex F, because Annex F only specifies `float` and `double`.  It leaves multiple options for `long double`; we want to specify `long double` to be IEEE 754 binary128.

PR #443, opened 2024-06-25
riscv-cc: Reference FP standards for FP types

cmuellner:
As reported in #430, we define the sizes and alignments of all floating-point types, but we don't reference the underlying standards. This change fills this gap.

Fixes: #430.


kito-cheng (approved):
Thanks, LGTM :)

PR #446, opened 2024-07-17
PIC medany: Fix parameters of load instruction

cmuellner:
The explanation of the medium position independent code model
    includes the instruction `l[w|d] a0, a0, %pcrel_lo(.Ltmp3)`
    to calculate the address of a non-local symbol.
    That's not the correct RISC-V syntax for a load instruction.
    Let's fix that.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -122,7 +122,7 @@ This model is similar to the medium any code model, but uses the
 
          # Calculate address of non-local symbol
 .Ltmp3:  auipc  a0, %got_pcrel_hi(symbol)
-         l[w|d] a0, a0, %pcrel_lo(.Ltmp3)
+         l[w|d] a0, %pcrel_lo(.Ltmp3)(a0)
 ----
 
 == Dynamic Linking


riscv-elf.adoc
```diff
@@ -122,7 +122,7 @@ This model is similar to the medium any code model, but uses the
 
          # Calculate address of non-local symbol
 .Ltmp3:  auipc  a0, %got_pcrel_hi(symbol)
-         l[w|d] a0, a0, %pcrel_lo(.Ltmp3)
```
jrtc27, line 125: Uh, a0 is being use as the base address register here. It's just using the ISA manual's syntax rather than the canonical GNU one by putting the immediate after the register as a separate operand rather than parenthesising the register and putting the immediate before it.

jrtc27: i.e.:
```diff
-         l[w|d] a0, a0, %pcrel_lo(.Ltmp3)
+         l[w|d] a0, %pcrel_lo(.Ltmp3)(a0)
```
is the right fix here

cmuellner: Right. Do you prefer to keep it that way or should I adjust to `ld      a0, %pcrel_lo(.Lpcrel_hi0)(a0)`?

jrtc27: > Since all examples in this section focus on RV32, this patch also replaces l[w|d] by lw.

They don't, they are just as valid on RV64 if symbol is an int.

jrtc27: Please keep the `[w|d]`

cmuellner: Done.

I guessed that the sequence focuses on 32-bit loads/stores for simplicity (-> RV32).
But you are right that this only affects the size of the value.
When calculating the address, the `addi` (in the examples above) works for 32-bit and 64-bit addresses.

kito-cheng (approved):
Thanks :)

PR #450, opened 2024-09-06
Fix spelling aupic -> auipc

topperc:


aswaterman (approved):

kito-cheng:
Thanks for catching that :)

PR #452, opened 2024-10-14
Introduce new relocation for landing pad

kito-cheng:
The R_RISCV_LPAD relocation can be used for PLT entry generation and also for
linker relaxation. Additionally, we defined a new mapping symbol type to help
users understand the function signature for the corresponding function.

The addend value is the label value, and it will point to the mapping symbol
placed at the beginning of the function.

e.g.
```asm
foo:         # void foo(void)
$sFvvE:
    lpad 123 # R_RISCV_LPAD $sFvvE + 123
```

We propose two linker relaxations for the landing pad. The first is removing
the entire landing pad, which can be used when symbols have local visibility,
and the address is not taken by any other reference. The second is a landing
pad scheme conversion, designed for backward compatibility (or as a workaround)
for legacy programs that may use functions without declarations.

---

NOTE: This is based on https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/434

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -558,7 +558,7 @@ Description:: Additional information about the relocation
                                             <| S - P
 .2+| 65      .2+| TLSDESC_CALL      .2+| Static  |                   .2+| Annotate call to TLS descriptor resolver function, `%tlsdesc_call(address of %tlsdesc_hi)`, for relaxation purposes only
                                             <|
-.2+| 66      .2+| LPAD              .2+| Static  |                   .2+| Annotates the landing pad instruction inserted at the beginning of the function.
+.2+| 66      .2+| LPAD              .2+| Static  |                   .2+| Annotates the landing pad instruction inserted at this point.
                                             <|
 .2+| 67-190  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
                                             <|
@@ -1593,9 +1593,9 @@ type (`sh_type`) is `SHT_RISCV_LANDING_PAD_INFO`, and the section flags
 
 The landing pad information section is a section that contains the information
 to generate PLT with landing pads. The static linker is required to use the
-landing pad values provided by this section when generating lpad instructions
-for functions listed in this section, unless otherwise specified by the user
-(e.g. through command line options) or the unlabeled CFI scheme is selected.
+landing pad values provided by this section when generating PLT entries, unless
+otherwise specified by the user (e.g. through command line options) or the
+unlabeled CFI scheme is selected.
 
 This section is consist by the entries of the following structure:
 
@@ -1613,8 +1613,9 @@ typedef struct
 } Elf64_Lpadinfo;
 ```
 
-The `lpi_name` field is the index into the string table for the symbol name,
-and the `lpi_value` field is the landing pad value for the symbol.
+The `lpi_sym` field refers to a symbol by indexing into in the symbol table.
+The indexed symbol table is `.dynsym` for shared libraries, and `.symtab` for
+other ELF files.
 
 Every global or weak function symbol, as well as undefined global or weak
 symbols expected to be functions, must have an entry in `.riscv.lpadinfo`.
@@ -1637,12 +1638,7 @@ logic:
     provided by these objects are termed "referencing landing pad values."
   - Static linkers should emit an error if the referencing landing pad values
     for a symbol conflict with each other, i.e., they are different.
-3. Merging Rules for Zero and Non-Zero Values:
-  - The merging logic between zero and non-zero landing pad values does not
-    apply to referencing landing pad values, as these are expected to be
-    generated through C declarations, where only non-zero landing pad values
-    are expected.
-4. Final Landing Pad Value Determination:
+3. Final Landing Pad Value Determination:
   - The final landing pad value is determined by merging definitive and
     referencing landing pad values as follows:
     - If the definitive landing pad value exists and is non-zero, the final
@@ -1705,7 +1701,7 @@ mapping symbols helps the disassembler to disassemble instructions correctly.
 
 The mapping symbol for the landing pad (`$s`) provides additional information
 for the labeled landing pad scheme. This mapping symbol may be generated when
-setting up a landing pad value (e.g., `auipc t2, %lpad_hash("FvvE")` will
+setting up a landing pad value (e.g., `lpad t2, %lpad_hash("FvvE")` will
 generate `$sFvvE`) or when emitting a landing pad instruction (e.g.,
 `lpad %lpad_hash("FvvE")` will generate `$sFvvE`).
 


kito-cheng:
cc @deepak0414 @ved-rivos @mylai-mtk

riscv-elf.adoc
```diff
@@ -2317,6 +2320,85 @@ instructions. It is recommended to initialize `jvt` CSR immediately after
...
 ----
 
+==== Landing Pad Relaxation
+
+  Target Relocation::: R_RISCV_LPAD
+
+  Description:: This relaxation type can relax lpad instruction into a none,
+  which removed the lpad instruciton.
+  This relaxation type can be performe even without `R_RISCV_RELAX`,
+  but the linker should pad nop instruciton to the same length of the original
```
aswaterman, line 2330: instruciton -> instruction

aswaterman (approved):
I think this reloc also provides enough info for the linker to fix up direct jumps to skip over landing pads when the landing pad was not relaxed.

Consider I have `call foo`, and the linker sees that the target of that call still has an `R_RISCV_LPAD` reloc (i.e. not relaxed).  The linker should be able to adjust `call foo` to `call foo+4`.  This should work whether the target is the PLT or the real function.

riscv-elf.adoc
```diff
@@ -2317,6 +2320,85 @@ instructions. It is recommended to initialize `jvt` CSR immediately after
...
+  Condition:: The associated function of this lpad must have local visibility, and
+  it must not be referenced by any relocation other than `R_RISCV_CALL` and
+  `R_RISCV_CALL_PLT`.
+  This relaxation can also be performed when the function has global visibility,
+  if the symbol does not have a corresponding PLT entry and is not referenced by
+  the GOT or by any relocation other than `R_RISCV_CALL` and `R_RISCV_CALL_PLT`.
+
+  Relaxation::
+  - Lpad instruciton associated with `R_RISCV_LPAD` can be removed.
+  - Lpad instruciton associated with `R_RISCV_LPAD` can be replaced with nop
```
aswaterman, line 2342: instruciton -> instruction

riscv-elf.adoc
```diff
@@ -2317,6 +2320,85 @@ instructions. It is recommended to initialize `jvt` CSR immediately after
...
 
+==== Landing Pad Relaxation
+
+  Target Relocation::: R_RISCV_LPAD
+
+  Description:: This relaxation type can relax lpad instruction into a none,
+  which removed the lpad instruciton.
+  This relaxation type can be performe even without `R_RISCV_RELAX`,
+  but the linker should pad nop instruciton to the same length of the original
+  instruction sequence.
```
deepak0414, line 2331: s/instruciton/instruction

riscv-elf.adoc
```diff
@@ -2317,6 +2320,85 @@ instructions. It is recommended to initialize `jvt` CSR immediately after
...
+  Target Relocation::: R_RISCV_LPAD
+
+  Description:: This relaxation type allows an `lpad` instruction to be relaxed
+  into `lpad 0`, which is a universal landing pad that ignores the label value
+  comparison. This relaxation is used when the label value is not computed
+  correctly.
+
+  Condition:: This relaxation can be performed without `R_RISCV_RELAX`, and
+  should not be enabled by default. The user must explicitly enable this
+  relaxation, and it should only be applied during static linking.
```
deepak0414, line 2377: Q: what happens in case of dynamic linking ?

riscv-elf.adoc
```diff
@@ -2317,6 +2320,85 @@ instructions. It is recommended to initialize `jvt` CSR immediately after
...
+  This relaxation type can be performe even without `R_RISCV_RELAX`,
+  but the linker should pad nop instruciton to the same length of the original
+  instruction sequence.
+
+  Condition:: The associated function of this lpad must have local visibility, and
+  it must not be referenced by any relocation other than `R_RISCV_CALL` and
+  `R_RISCV_CALL_PLT`.
+  This relaxation can also be performed when the function has global visibility,
+  if the symbol does not have a corresponding PLT entry and is not referenced by
+  the GOT or by any relocation other than `R_RISCV_CALL` and `R_RISCV_CALL_PLT`.
```
rui314, line 2338: I think we can explain the behavior more clearly if we avoid mentioning symbol visibility. The only important thing here is whether or not a symbol is visible to other ELF modules, i.e. whether or not the symbol is in the dynamic symbol table. Symbol visibility is just one way to control it.

riscv-elf.adoc
```diff
@@ -548,7 +548,9 @@ Description:: Additional information about the relocation
                                             <| S - P
 .2+| 65      .2+| TLSDESC_CALL      .2+| Static  |                   .2+| Annotate call to TLS descriptor resolver function, `%tlsdesc_call(address of %tlsdesc_hi)`, for relaxation purposes only
                                             <|
-.2+| 66-190  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
+.2+| 66      .2+| LPAD              .2+| Static  |                   .2+| Annotates the landing pad instruction inserted at the beginning of the function. The addend indicates the label value of the landing pad, and the symbol value is the address of the mapping symbol for the function signature, which will have the same address as the function.
```
rui314, line 551: What is the "label value of the landing pad?

riscv-elf.adoc
rui314, line 2338: How should we find a function symbol for a given R_RISCV_LPAD relocation? Should we just look for a function symbol at the same location as the relocation refers to?

riscv-elf.adoc
```diff
@@ -2317,6 +2320,85 @@ instructions. It is recommended to initialize `jvt` CSR immediately after
     csrw  jvt, a0
 ----
 
+==== Landing Pad Relaxation
+
+  Target Relocation::: R_RISCV_LPAD
+
+  Description:: This relaxation type can relax lpad instruction into a none,
+  which removed the lpad instruciton.
```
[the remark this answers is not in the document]
ved-rivos: "into a none" -> "into a no-op"

riscv-elf.adoc
```diff
@@ -2317,6 +2320,85 @@ instructions. It is recommended to initialize `jvt` CSR immediately after
...
+--
+
+==== Landing Pad Scheme Relaxation
+
+  Target Relocation::: R_RISCV_LPAD
+
+  Description:: This relaxation type allows an `lpad` instruction to be relaxed
+  into `lpad 0`, which is a universal landing pad that ignores the label value
+  comparison. This relaxation is used when the label value is not computed
+  correctly.
```
ved-rivos, line 2373: what would be the cases where a label may be computed incorrectly.

riscv-elf.adoc
```diff
@@ -1582,6 +1584,7 @@ A number of symbols, named mapping symbols, describe the boundaries.
 | $x.<any>
 | $x<ISA>  .2+| Start of a sequence of instructions with <ISA> extension.
 | $x<ISA>.<any>
+| $s<function-signature-string> | Marker for the landing pad instruction. This should only be used with the function signature-based scheme and should be placed only at the beginning of the function.
```
mylai-mtk, line 1587: I don't quite get the purpose of this mapping symbol: It looks like the only reference to these symbols come from the `LPAD` relocation, but for what? The `LPAD` relocation already have the 20-bit label stored in its addend, and its link to this `%s` mapping symbol provides only the additional information of the signature string, which is not needed (as of now) to link the object files. What's the point of providing the link from function symbol to signature?

mylai-mtk:
Q: Does the `LPAD` relocation and/or the `$s<func-sig>` mapping symbol serve the purpose of providing labels when generating PLT? If so, how is it expected to extract the label for a given function symbol from these new entities?

rui314:
I don't think I fully understand this proposal, but from the linker's perspective, I believe we just need the following:

- A `R_RISCV_LPAD` relocation, whose `r_offset` is at each removable `lpad` instruction location, and `r_sym` refers to a function symbol
- If the function's address was not taken (i.e. the function symbol was not referenced by anything but `R_RISCV_CALL` or  `R_RISCV_CALL_PLT`) and the function symbol is not in the dynamic symbol, the linker removes the `lpad` instruction.

I don't think we need a mapping symbol for the linker to remove a lpad instruction.

riscv-elf.adoc
kito-cheng: I gave it some more thought, and this can actually be applied beyond just static linking. However, dependent shared libraries won’t automatically convert along with it. I’ve removed that limitation and added a NOTE to explain the situation.

riscv-elf.adoc
kito-cheng: Good suggestion, applied https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/452/commits/02546dea2621f2a4850cd9a6920bc6f5bfd32495 :)

riscv-elf.adoc
kito-cheng: https://github.com/riscv/riscv-cfi/issues/151#issuecomment-1918293510

It's kinda debugging propose only, so it safe to strip like all other mapping symbols

riscv-elf.adoc
kito-cheng: Right now, this design requires scanning through the symbol table and relocation table once to figure out which symbols have landing pads. I had previously thought about creating a new section to handle this, but I realized that when dealing with linker relaxation, the best way is still to use relocations to mark them. This approach also avoids introducing a new section format.

kito-cheng:
@rui314 
> I don't think I fully understand this proposal, but from the linker's perspective, I believe we just need the following:
> 
> A R_RISCV_LPAD relocation, whose r_offset is at each removable lpad instruction location, and r_sym refers to a function symbol
> If the function's address was not taken (i.e. the function symbol was not referenced by anything but R_RISCV_CALL or R_RISCV_CALL_PLT) and the function symbol is not in the dynamic symbol, the linker removes the lpad instruction.
> I don't think we need a mapping symbol for the linker to remove a lpad instruction.

 The design of the mapping symbol is meant to make debugging easier for users, so it’s actually optional. This was discussed in the CFI spec issue [1]. If we set aside that purpose, I also think it’s a better approach to have r_sym point to the function symbol.

As for using the addend to record the landing pad value, it helps with PLT generation. The linker can get the value by scanning the instruction at that address, but encoding it directly in the addend avoids needing to read the instruction during relocation.

[1] https://github.com/riscv/riscv-cfi/issues/151#issuecomment-1918293510

riscv-elf.adoc
mylai-mtk: But how is the linker expected to know the incorrectness so it can perform this relaxation?

The Zicfilp mechanism is employed when issuing an indirect call through function pointers, and when calling functions through PLT:

In the first case (indirect calls through pointers), to know that an lpad insn needs to be relaxed to `lpad 0` due to the original label being incorrect, the linker would need to know where the pointer points to, so the caller's label (the "correct" one) can be checked against the callee's label (the lpad insn). I'm not sure if this is the scenario you're targeting, but if it is, I think this (knowing where the pointer points to, or knowing where the call would come from) is an expectation too high for linkers. Besides, in this scenario, I would also wonder how linkers are expected to retrieve the caller's label (the "correct" one) so it can make the comparison?

In the second case (calls through PLT), the indirect call happens in the PLT, which is generated by linkers. The label which linkers use to generate PLT would come from the addend of the `LPAD` relocation, which should contain the same label as the one in its referenced `lpad` insn, so there would be no chance of mismatch and thus incorrectness identifiable by linkers.

The above is my guess and understanding of the intended usage of this relaxation. If we're not on the same page, please do let me know.

kito-cheng: Linker never know (or not always know), and also that's not the right layer to analysis (or guess:P ),  so I expect that relaxation should only enabled when user pass something like `-z force-simple-landing-pad-scheme` to linker.

riscv-elf.adoc
mylai-mtk, line 551: Is this relocation only for the func-sig scheme? Based on its description, it looks like so, but the following `LPAD` relaxation that removes the `lpad` insn seems also applicable to the unlabeled scheme.

riscv-elf.adoc
mylai-mtk: If purpose is to display function signatures when disassembling, this mechanism seems a bit incomplete (?) I suppose since the relocation is a static one, it would not stay in the binary after static linking, thus if a user disassembles a linked ELF, it's still the label numbers instead of signatures that get displayed?

riscv-elf.adoc
mylai-mtk: > Right now, this design requires scanning through the symbol table and relocation table once to figure out which symbols have landing pads. 

Does this mean that `O(R + F)` time complexity and `O(R)` space complexity, where `R` is number of LPAD relocations and `F` is number of global function symbols in the same section, to build a map from function symbols to lpad labels is expected? This map is needed to synthesize PLT.

Note: The complexities come from the following algorithm:
```c++
HashMap<Address, LabelLabel> LpadRelocMap;
HashMap<FunctionSymbol, LpadLabel> LpadLabelMap;
for (auto R: LpadRelocations)
  LpadRelocMap.insert(R.Address, R.addend);

for (auto S: FunctionSymbols)
  if (S.Address in LpadRelocMap) lpadLabelMap.insert(S, LpadRelocMap[S.Address]);
```

mylai-mtk:
@kito-cheng 

> As for using the addend to record the landing pad value, it helps with PLT generation. The linker can get the value by scanning the instruction at that address, but encoding it directly in the addend avoids needing to read the instruction during relocation.

I think this doesn't provide the lpad label we need for PLT generation? For a call target to be generated in the PLT, the target should reside in a shared library or somewhere that the linker cannot determine during static-time linking. When the target resides in a shared library, it does not have the `LPAD` relocation associated with it, since the relocation is a static relocation and should be consumed with by the static linker that created the shared library?

kito-cheng:
@mylai-mtk

> I think this doesn't provide the lpad label we need for PLT generation? For a call target to be generated in the PLT, the target should reside in a shared library or somewhere that the linker cannot determine during static-time linking. When the target resides in a shared library, it does not have the LPAD relocation associated with it, since the relocation is a static relocation and should be consumed by the static linker that created the shared library?

Hmmmmmmmmmmmmmm, yeah, do you have any better idea that creating a new section to recording the label value for those undefined symbols?

riscv-elf.adoc
kito-cheng: That should also work for unlabeled scheme as well, let me think how to make it clearly.

mylai-mtk:
@kito-cheng 

> Hmmmmmmmmmmmmmm, yeah, do you have any better idea that creating a new section to recording the label value for those undefined symbols?

Nope. After skimming through existing sections, I don't have ideas better than creating a new section.

My previous prototype of generating extra symbols for every function may work, but there's too many drawbacks with this approach, including bloated symbol table size and slowed down symbol resolution performance, so I would not recommend this approach seriously.

kito-cheng:
@mylai-mtk added new section. and also new asm directive https://github.com/riscv-non-isa/riscv-asm-manual/pull/113

riscv-elf.adoc
```diff
@@ -1568,6 +1575,51 @@ the `Zicfilp` extension. An executable or shared library with this bit set is
...
+{
+  Elf32_Word    lpi_name;                /* Symbol name (string tbl index) */
+  Elf32_Word    lpi_sig;                 /* Signature for the symbol (string tbl index) */
+  Elf32_Word    lpi_value;               /* Landing pad value for the symbol */
+} Elf32_Lpadinfo;
+
+typedef struct
+{
+  Elf64_Word    lpi_name;                /* Symbol name (string tbl index) */
+  Elf64_Word    lpi_sig;                 /* Signature for the symbol (string tbl index) */
```
mylai-mtk, line 1597: I think we can use `uint32_t` here instead of the less clear `ElfN_Word`? It looks like other string table references also use `uint32_t`, e.g. `st_name` in `struct Elf64_Sym`.

riscv-elf.adoc
```diff
@@ -1568,6 +1575,51 @@ the `Zicfilp` extension. An executable or shared library with this bit set is
...
+  Elf64_Word    lpi_value;               /* Landing pad value for the symbol */
```
mylai-mtk, line 1598: Lpad labels are only 20-bit wide, so we can use `uint32_t`. Same for the 32-bit version above.

riscv-elf.adoc
```diff
@@ -1568,6 +1575,51 @@ the `Zicfilp` extension. An executable or shared library with this bit set is
...
+
+This section is consist by the entries of the following structure:
+
+```
+typedef struct
+{
+  Elf32_Word    lpi_name;                /* Symbol name (string tbl index) */
+  Elf32_Word    lpi_sig;                 /* Signature for the symbol (string tbl index) */
+  Elf32_Word    lpi_value;               /* Landing pad value for the symbol */
+} Elf32_Lpadinfo;
```
mylai-mtk, line 1592: `Elf32_LpadInfo`? (Big `I` instead of small `i`) (Same for 64-bit version below)

riscv-elf.adoc
```diff
@@ -1568,6 +1575,51 @@ the `Zicfilp` extension. An executable or shared library with this bit set is
...
+signature, it can be 0 if the signature string is not present,
+and the `lpi_value` field is the landing pad value for the symbol.
+
+The string hold by `lpi_signature` field is the function signature string, which
+is encoded as same as the mapping symbol of the function signature.
+
+NOTE: Using same encoding as mapping symbol aims to reduce the size of the
+string table
+
+Every symbol with global or weak bind must has a corresponding entry in this
```
mylai-mtk, line 1613: I think you mean "Every **function** symbol with global or weak"?

riscv-elf.adoc
```diff
@@ -1568,6 +1575,51 @@ the `Zicfilp` extension. An executable or shared library with this bit set is
...
+section, the `lpi_name` field must be the same as the symbol name string table
+index.
+
+This section can be discard after static linking stage.
```
mylai-mtk, line 1617: Since you mentioned that "Every symbol with global or weak bind must has a corresponding entry", I think it implies that  the lpad labels are **provided by the object file that defines the function**, right? If this is the case, we can't discard this section after static linking when creating a shared library, since library users would expect to find lpad labels later when linking against this share library.

riscv-elf.adoc
kito-cheng: It's uint32_t, but using ElfNN_* types for the consistency with other ELF structure
```
/* Types for signed and unsigned 32-bit quantities.  */
typedef uint32_t Elf32_Word;
typedef int32_t  Elf32_Sword;
typedef uint32_t Elf64_Word;
typedef int32_t  Elf64_Sword;
```

riscv-elf.adoc
kito-cheng: That's kinda following the naming convention in the ELF stuffs, e.g. `Elf64_Verdef`, `Elf64_Verdaux`, `Elf64_Verneed`, `Elf64_Nhdr`...

riscv-elf.adoc
mylai-mtk: If this is the case (labels can be provided by the object that uses but doesn't define the function), why require labels to be there in the defining object ("Every symbol with global or weak bind must has a corresponding entry")? For the sake of checking if the use-site and define-site agree on the same lpad label?

riscv-elf.adoc
```diff
@@ -1568,6 +1575,51 @@ the `Zicfilp` extension. An executable or shared library with this bit set is
...
+
+Static linker should emit error if objects with same symbol but different
+landing pad value are beging merged, however it may suppress the error if
+linker enable the landing pad schem relaxation.
```
mylai-mtk, line 1621: Is it beneficial that we also implement this check in the dynamic linker for dynamic symbols? E.g. Abort the program if label mismatches are found at program startup or `dlopen()`

riscv-elf.adoc
```diff
@@ -1568,6 +1575,51 @@ the `Zicfilp` extension. An executable or shared library with this bit set is
...
+The `lpi_name` field is the index into the string table for the symbol name,
+the `lpi_signature` field is the index into the string table for the function
+signature, it can be 0 if the signature string is not present,
+and the `lpi_value` field is the landing pad value for the symbol.
+
+The string hold by `lpi_signature` field is the function signature string, which
+is encoded as same as the mapping symbol of the function signature.
+
+NOTE: Using same encoding as mapping symbol aims to reduce the size of the
+string table
```
mylai-mtk, line 1611: This means the signatures stored start with `$` and is in the format of `$x<function-signature-string>`? (Note the additional `x` there)

If the goal is to save bytes in the string table, I think we can always use `Symbol($x<function-signature-string>).st_name + 2` to specify the signature string to keep the referred signature precise without the `$x` prefix?

riscv-elf.adoc
mylai-mtk: If the purpose of the mapping symbol is to provide debug/disassembling info, I think after the introduction of the `.riscv.lpadinfo` section, this purpose can be better served by the new section, since it contains the same information, and has a more size-compact format than the symbol table.

riscv-elf.adoc
mylai-mtk: Also, we should restrict this to local, non-pointer-taken symbols.

riscv-elf.adoc
mylai-mtk, line 551: Does this LPAD relocation intend to serve the "real" purpose of a relocation? That is, ask linkers to fill-in some value (in this case, the label of the lpad instructions) to some offset at link time.

When prototyping this relocation in LLVM, it appears to me that the LLVM backend assumes places to be relocated have a placeholder value 0 encoded, so to emit the LPAD relocations, 0s would have to be encoded at the label locations in relocatable files. This can of course be changed to encode the correct label along with relocation emitted, but I want to ask if this changed is needed, or I can rely on linkers to fill-in the correct labels when relocating at static time.

riscv-elf.adoc
```diff
@@ -1568,6 +1575,51 @@ the `Zicfilp` extension. An executable or shared library with this bit set is
...
+for generating function signature based landing pad PLT, this section also may
+exsiting when the unlabeled landing pad scheme is used.
+
+This section is consist by the entries of the following structure:
+
+```
+typedef struct
+{
+  Elf32_Word    lpi_name;                /* Symbol name (string tbl index) */
+  Elf32_Word    lpi_sig;                 /* Signature for the symbol (string tbl index) */
```
mylai-mtk, line 1590: Proposal: Rename `lpi_sig` to `lpi_info`

This allows this section format to be more versatile.

riscv-elf.adoc
```diff
@@ -1568,6 +1575,51 @@ the `Zicfilp` extension. An executable or shared library with this bit set is
...
+
+The string hold by `lpi_signature` field is the function signature string, which
+is encoded as same as the mapping symbol of the function signature.
+
+NOTE: Using same encoding as mapping symbol aims to reduce the size of the
+string table
+
+Every symbol with global or weak bind must has a corresponding entry in this
+section, the `lpi_name` field must be the same as the symbol name string table
+index.
```
mylai-mtk, line 1615: > the `lpi_name` field must be the same as the symbol name string table index.

Excuse me, I don't understand the necessity of mandating this.

I believe the `lpi_name` field just represents a string which contains the function name, and the "address" of this string doesn't matter? Is there any benefit of this mandate?

riscv-elf.adoc
```diff
@@ -1568,6 +1575,51 @@ the `Zicfilp` extension. An executable or shared library with this bit set is
 required to generate PLTs with the landing pad (`lpad`) instruction, and all
 label are set to a value which hashed from its function signature.
 
+=== Landing Pad Information Section (`.riscv.lpadinfo`)
```
mylai-mtk, line 1578: We need to specify section header info, including the `sh_type`, `sh_flags` field.

riscv-elf.adoc
kito-cheng: Drop mapping symbol so this paragraph will be removed

riscv-elf.adoc
kito-cheng: Good point

riscv-elf.adoc
kito-cheng: I think the constraint can be removed since the index is not really matter, and this constraint may not come with extra benefit even it should already the same index in general.

riscv-elf.adoc
```diff
@@ -1568,6 +1575,51 @@ the `Zicfilp` extension. An executable or shared library with this bit set is
...
+
+Landing pad information section is a section that contains the nessary information
+for generating function signature based landing pad PLT, this section also may
+exsiting when the unlabeled landing pad scheme is used.
```
mylai-mtk, line 1582: Change to: The landing pad information section is a section that contains the information to generate PLT with landing pads. The static linker is required to use the landing pad values provided by this section when generating `lpad` instructions for functions listed in this section, unless otherwise specified by the user (e.g. through command line options).

This change allows us to:
+ Avoid limiting the use of this section to the function-signature scheme
+ Be clear that when lpad labels can be inferred from other information sources contained in the source object (e.g. the `GNU_PROPERTY_RISCV_FEATURE_1_CFI_LP_UNLABELED` flag would imply `0` as the labels), the `riscv.lpadinfo` section would take precedence, unless the linker user commands otherwise clearly (e.g. through the `-z force-zicfilp=unlabeled` command line option)

riscv-elf.adoc
```diff
@@ -1568,6 +1575,51 @@ the `Zicfilp` extension. An executable or shared library with this bit set is
...
+
+This section is consist by the entries of the following structure:
+
+```
+typedef struct
+{
+  Elf32_Word    lpi_name;                /* Symbol name (string tbl index) */
```
mylai-mtk, line 1589: Just curious: Why this isn't the index of symbol in the symbol table?

The symbol table index provides a more definite reference to the symbol, considering the case of repeated symbol names (though this may not really happen with function symbols as its an error to globally define a function more than once.)

riscv-elf.adoc
kito-cheng: I agree with you, so...I just say `can be` not `must be` here, that make it strip-able, but it's also legal to leave here

mylai-mtk: Hmmm, if this is the intention, I would suggest you make it clear that it means the section is strip-able, and does not mean that the section can be discarded under all circumstances after static linkage. I believe if a user strips his binary, he should know the consequence, so marking a section that is only strip-able under certain situations strip-able is acceptable.

My concern is that if this sentence is considered by a linker implementation to be the spec of linker behavior that uniformly allows the section to be dropped after static linkage, the spec is flawed, since when producing shared libraries, the section cannot be dropped or otherwise the library could risk linkage failure to executables.

riscv-elf.adoc
mylai-mtk: I've come up with a case that would cause the current format to fail:

```c
// public_foo.c
void foo() {}

// private_foo_1.c
static int foo(int i) { return i; }
void *get_foo_1() { return foo; }

// private_foo_2.c
static char foo(char c) { return c; }
void *get_foo_2() { return foo; }
```

Compile them to a shared library, and you get 1 GLOBAL FUNC and 2 LOCAL FUNC `foo` symbols, and the signature of these 3 `foo`s are different. With the current format, the lpad info entries of these 3 symbols cannot be distinguished, and this would cause problem for the linker.

To support showing `lpi_info` contents when disassembling, none of the 3 lpad info entries can be omitted, so to distinguish them, symbol table indices should be used in place of string table indices to refer to symbols, and as a consequence, `lpi_name` should be renamed to `lpi_sym` or `lpi_func` (the field does not represent "name" anymore).

kito-cheng: TL;DR: I will update to the symbol index in next update.

And let me tell the story why I use the symbol name before: In the earlier version, we have mapping symbol at the beginning of the function, so local function could use that only, but we gonna to remove the mapping symbol, so that means local symbol may also rely this mechanism to record the signature.

Recording the symbol name is fine before since global symbol has unique symbol name (in theory), and local symbol may have duplicated symbol name, but it's not a problem before.

The concern for using the symbol index is: `.symtab` will gone after strip, so the index is kinda become meaningless, and we have only index for `.dynsym`, but `.dynsym` only contain exported global symbol only, however this issue can be resolve if we say the `.riscv.lpadinfo` should be stripped if `.symtab` got stripped.

So the conclusion is: I think we should use symbol index here, rather than symbol name as your suggestion.

mylai-mtk: Thank you for your explanation, and sorry for ruining your original design about mapping symbol 😥

The main reason I was against the mapping symbol approach was that it's intended usage was obscure. It only had a reference from the `LPAD` relocation that is intended to provide the disassembling string for the lpad label, but considering that LPAD relocations would be gone after static linking, this solution looks incomplete (it only provides the disassembling string for relocatables, but not for executables and shared libraries), so it didn't convince me that we should adopt the mapping symbol approach.

But if how the mapping symbols were associated to the lpad insns is clarified and unified by dropping the reference in the LPAD relocation, and solely rely on the mapping symbols having the `st_value` that points to the lpad insns, I think using mapping symbols to provide the disassembling strings actually has its advantages (and is better than using `.riscv.lpadinfo`):

+ It allows decorating lpad insns not only at the beginning of function symbols, but also in the middle of function bodies
+ As a kind of debug info, it's automatically stripped away when the the binary is stripped
+ Symbol handling in compiler backends and linkers have extensive infrastructure support. Not much work would need to be carried out to have the mapping symbols properly emitted
+ If the approach is adopted, we can remove the `lpi_info` field from the `.riscv.lpadinfo` section, making the section simpler and probably more compact and concise than the current format
  + Also, separating the optional debug info from the necessary `.riscv.lpadinfo` simplifies the treatment of the `.riscv.lpadinfo` section in toolchain (it would be easier to determine if the section needs to kept or can be discarded, and relevant data structures would be easier to maintain)

The only disadvantage (I can come up with) of the updated mapping symbol approach would be a bloated symbol table, but symbol tables are strippable in the final executables/shared libraries, so it does not matter much in production after all. 

Saying so much, would you mind to clarify how the mapping symbol approach was supposed to work, and let's reconsider it (and the format `.riscv.lpadinfo`)?

Thank you 🙇‍♂️

mylai-mtk:
Is it possible that we design this spec in terms of just labeled and unlabeled schemes, instead of the more constraining func-sig and unlabeled schemes? (Same goes for the PLT draft spec)

The motivation is to increase the versatility of these specs. Based on the observation (and "feeling" during my prototyping) that nearly all components of these specs for the func-sig scheme (including the PLT form, relocations, relaxations, `.riscv.lpadinfo` section content, mapping symbol, and the `%lpad_label("func-sig")` assembly operand modifier) are designed with just the assumption that the lpad values can be non-zero, and they come from strings. This implies that the specification of "func-sig" can actually be removed to make these specs more versatile to support all schemes that generates lpad values from strings, which is flexible enough to encode almost all data formats.

mylai-mtk:
Proposal: Should we consider relaxation that removes (at best effort) lpad insns if the resulting GNU AND features does not contain CFI_LP flags?

riscv-elf.adoc
kito-cheng: `Elf64_Word` is `typedef uint32_t Elf64_Word;`, and it's convention to use `ElfXX_*` type here

riscv-elf.adoc
kito-cheng: I would add a NOTE make this non-mandatory suggestion in the spec.

mylai-mtk:
Hi, I have implemented a prototype of this draft in the LLVM toolchain. Since there are some problems if the prototype strictly follows the draft spec, I had modified the spec and here I list my (major) modifications and the reasons:

---

+ Current draft: (None)
+ Modified spec: If the LPAD relocation refers to a symbol, the addend is 0, the func-sig scheme is adopted, and the referred symbol has a non-zero lpad value listed in `.riscv.lpadinfo`, the lpad insn referred to by the LPAD relocation is updated to use the referred symbol's non-zero lpad value from `.riscv.lpadinfo`. The static linker can emit a warning if the LPAD relocation's referred symbol has a zero lpad value listed in `.riscv.lpadinfo` when the LPAD relocation refers to a symbol, the addend is 0, and the func-sig scheme is adopted
+ Reason: This relaxation enables functions implemented in assembly but called from C to conveniently use `lpad 0` as its lpad value, e.g.

```asm
foo:
  lpad 0
  ...
```

```c
// This line causes compiler to emit correct label for `foo` in `.riscv.lpadinfo`,
// which can be used to fix the lpad insn for `foo`.
extern int foo(void);
```

---

+ Current draft: The array element structure of `.riscv.lpadinfo` section is

```c
typedef struct {
  ElfN_Word lpi_name;  // Symbol name (string table index)
  ElfN_Word lpi_sig;   // Signature for the symbol (string table index)
  ElfN_Word lpi_value; // Landing pad value for the symbol
} ElfN_Lpadinfo;
```

+ Modified spec: The array element structure of `.riscv.lpadinfo` section is

```c
typedef struct {
  ElfN_Word lpi_sym;   // Function symbol (`.symtab`/`.dynsym` table index)
  ElfN_Word lpi_value; // Landing pad value for the symbol
} ElfN_Lpadinfo;
```

The `lpi_sym` fields refer to symbols in `.dynsym` for shared libraries; for other ELF files, they refer to `.symtab`

+ Reason:
  + `lpi_name` is changed to `lpi_sym`, since intuitively, it's more precise to refer to a symbol using its symbol table index rather than its name, but perhaps more importantly, this allows static linkers to avoid the complex process of resolving names to symbols, especially in the case of versioned symbols
  + `lpi_sig` is removed, since @kito-cheng provided/explained a better idea to store the information using mapping symbols. By removing `lpi_sig` from `.riscv.lpadinfo`, it's also now clear that the section is needed only before static linkage, and can be discarded afterwards
    + Note that shared libraries are always possible to be fed to static linkers, so strictly speaking it's not recommended to throw away their `.riscv.lpadinfo`
  + `lpi_sym` refers to `.dynsym`/`.symtab` for shared libraries/other files, since it's more natural for the static linker to look at these different symbol tables for these file kinds. For example, llvm-lld does not parse `.symtab` for shared libraries now

---

+ Current draft: Every global/weak function symbol must have an entry in `.riscv.lpadinfo`
+ Modified spec: Every global/weak function symbol, and undefined global/weak symbol that are expected to be a function, must have an entry in `.riscv.lpadinfo`
+ Reason: Undefined global symbols that are expected to be a function may provide a non-zero lpad value that can be used to overwrite a zero lpad value. Undefined weak symbols that are expected to be a function may have a PLT entry, so their lpad values are needed

---

+ Current draft: Static linkers should emit error if objects with same symbol but different lpad values are beging merged
+ Modified spec: Static linkers should merge lpad values from objects using the following logic: If a resolved symbol has a zero and a non-zero lpad value coming from objects that defines, both globally and weakly, the symbol, the non-zero lpad value is adopted as the "definitive lpad value" for the symbol. Otherwise, the definitive lpad value for the symbol is 0. If the symbol defining objects provides more than 1 non-zero lpad values for a resolved symbol, an error should be emitted. Static linkers should also emit an error if for a symbol, the lpad values ("referencing lpad values") coming from objects that just references it (not defining it) conflicts with each other, i.e. they are different. Note that the merging between zero and non-zero lpad values does not apply to the lpad values coming from objects that just references the symbol, since these references are expected to be generated through C declarations, so only non-zero lpad values are expected. The lpad value provided by objects that just references a symbol is termed "referencing lpad values" for the symbol. The final lpad value is merged from definitive and referencing lpad values: If the definitive lpad value exists and is non-zero, the final lpad value is definitive lpad value. Otherwise, if the referencing lpad value exists, the final lpad value is the referencing lpad value. Otherwise, if the definitive lpad values exists but is 0, the final lpad value is 0. Note that the final lpad value is unknown if both definitive and referencing lpad values do not exist.
+ Reason: This lpad value merging logic aims to allow the convenience of `lpad 0` to be supplemented with the robustness of lpad values coming from C declarations, while still guarding against lpad value conflictions. The separation of definitive and referencing lpad values is based on the observation that build system test programs (e.g. simple C program from CMake that tests whether a function is available) often use incorrect function signatures. These incorrect function signatures would trigger lpad value confliction if they were not considered separatively from definitive lpad values. By separating definitive and referencing lpad values and giving definitive lpad values higher priority, we can safely pass these incorrect programs while preventing their incorrectness from contaminating the final lpad values to the largest extent.

---

+ Current draft: The `.riscv.lpadinfo` section may exist when the unlabeled scheme is adopted
+ Modified spec: The `.riscv.lpadinfo` section may not exist, or is ignored, when the unlabeled scheme is adopted
+ Reason: The modified `.riscv.lpadinfo` section is purely for the purpose of providing lpad values for a labeled scheme. It does not make sense to handle this section when the unlabeled scheme is adopted

---

+ Current draft: The LPAD relocation refers to the mapping symbol, which has the `st_value` that points to the lpad insn
+ Modified spec: The LPAD relocation **optionally** refers to a function symbol, which has the `st_value` that points to the lpad insn
+ Reason: This facilitates the new relaxation that tightens `lpad 0` to `lpad non-0` with information from `.riscv.lpadinfo`. The referred to function symbol is used to lookup information in `.riscv.lpadinfo`

mylai-mtk:
I had uploaded my prototype (URLs provided in the spec modification discussion) so more details can be observed. The prototype has passed some C/C++ language test suites.

---

> I consider that before but that mean we need to introduce another attribute or property to record which scheme are used - so I think let keep allocate dedicate bit for function signature - but I do agree we could relax the wording a little bit to make this spec to introduce new labeling scheme without too much modifier in future :)

👍

> I am putting all modifier together and plan to update this tomorrow (if no accident or other urgent thing pop up), also I will put your (@mylai-mtk) name in the commit as coauthor :)

Thank you

kito-cheng:
Changes:
- LPAD relocation no longer require addend and symbol.
- Mapping symbol has extend to used in landing pad setup instruciton as well. e.g. auipc t2, %lpad_hash("FvvE") will having a $sFvvE
- Drop lpi_sig from ElfXX_Lpadinfo
- Rename lpi_name to lpi_sym, and it means symbol index rather than string index now.
- Re-org the commits into 3 pieces: LPAD relocation and relaxation, `.riscv.lpadinfo` section and mapping symbol

NOTE:
I take most comments from @mylai-mtk except this:

> Current draft: (None)
Modified spec: If the LPAD relocation refers to a symbol, the addend is 0, the func-sig scheme is adopted, and the referred symbol has a non-zero lpad value listed in .riscv.lpadinfo, the lpad insn referred to by the LPAD relocation is updated to use the referred symbol's non-zero lpad value from .riscv.lpadinfo. The static linker can emit a warning if the LPAD relocation's referred symbol has a zero lpad value listed in .riscv.lpadinfo when the LPAD relocation refers to a symbol, the addend is 0, and the func-sig scheme is adopted
Reason: This relaxation enables functions implemented in assembly but called from C to conveniently use lpad 0 as its lpad value, but still gets a non-zero lpad value, e.g.

That's too aggressive, so let me leave this along here, I would suggest implement that as a linker extension and should enable by an option.

riscv-elf.adoc
```diff
@@ -1568,6 +1575,79 @@ the `Zicfilp` extension. An executable or shared library with this bit set is
...
+} Elf32_Lpadinfo;
+
+typedef struct
+{
+  Elf64_Word    lpi_sym;                 /* Symbol index */
+  Elf64_Word    lpi_value;               /* Landing pad value for the symbol */
+} Elf64_Lpadinfo;
+```
+
+The `lpi_name` field is the index into the string table for the symbol name,
```
mylai-mtk, line 1606: > The `lpi_name` field is the index into the string table for the symbol name

You forget to update this sentence. My updated wording would be:

> The `lpi_sym` field refers to a symbol by indexing into in the symbol table. The indexed symbol table is `.dynsym` for shared libraries, and `.symtab` for other ELF files.

mylai-mtk (commented):
Thank you for your updates. I've found a few wording problems, but I think the spec is in good shape (at least for now 😂).

riscv-elf.adoc
```diff
@@ -1612,6 +1693,12 @@ is not enough for the disassembler to disassemble the `rv64gcv` version
 correctly. Specifying ISA string appropriately with the two memcpy instruction
 mapping symbols helps the disassembler to disassemble instructions correctly.
 
+The mapping symbol for the landing pad (`$s`) provides additional information
+for the labeled landing pad scheme. This mapping symbol may be generated when
+setting up a landing pad value (e.g., `auipc t2, %lpad_hash("FvvE")` will
```
mylai-mtk, line 1698: I think you mean `lui t2, %lpad_hash(...)`?

BTW, the condition under which this mapping symbol can be emitted could actually be summarized as "if there's a `%lpad_hash(...)`, an lpad mapping symbol is emitted". Would you mind to restructure the spec around the summarized condition so that it's more versatile compared to the current spec that explicitly constrains the emission condition of the symbol to the 2 listed "intentions"?

riscv-elf.adoc
```diff
@@ -1612,6 +1693,12 @@ is not enough for the disassembler to disassemble the `rv64gcv` version
 correctly. Specifying ISA string appropriately with the two memcpy instruction
 mapping symbols helps the disassembler to disassemble instructions correctly.
 
+The mapping symbol for the landing pad (`$s`) provides additional information
```
mylai-mtk, line 1696: > The optional ISA information, when present, will be used until the next instruction mapping symbol.

This existing spec for the interpretation of mapping symbols conflicts with the label mapping symbol.

riscv-elf.adoc
```diff
@@ -1432,6 +1516,138 @@ that a linker or runtime loader needs to check for compatibility.
...
+2. Referencing Landing Pad Values:
+  - For symbols referenced (but not defined) by objects, the landing pad values
+    provided by these objects are termed "referencing landing pad values."
+  - Static linkers should emit an error if the referencing landing pad values
+    for a symbol conflict with each other, i.e., they are different.
+3. Merging Rules for Zero and Non-Zero Values:
+  - The merging logic between zero and non-zero landing pad values does not
+    apply to referencing landing pad values, as these are expected to be
+    generated through C declarations, where only non-zero landing pad values
+    are expected.
```
mylai-mtk, line 1634: I wonder if this assumption about referencing lpad values being only generated through C/C++ declarations and thus must be non-zero is good/necessary. It looks like the assumption may be incorrect, and replacing this limitation with the merge of non-0-valued referencing lpad values over 0-valued lpad values doesn't hurt the security property. 

By saying the assumption may be incorrect, I mean that by observing that many language features require calls into runtime library (e.g. `__cxa_allocate_exception`), these invocations actually represents references generated not from the program source code, but by the compiler. The non-ideal part of the status-quo is that these compiler generated invocations often provides no C/C++ declarations in compiler internals, which means non-zero valued referencing lpad values cannot be obtained for these references. When generated in the form of indirect invocations, these cases introduce a justified scenario for 0-valued referencing lpad values. This scenario may not happen frequently or even exist now (as I did not run into this "compiler-generated indirect invocations into runtime libraries" case), but I think relaxing referencing lpad values to allow 0-valued lpad values is a good practice to reserve the space for possible future program/language forms.

The proposed relax need not to reduce the security property provided by forcing referencing lpad values to be non-zero, since we can introduce merge rules for referencing lpad values (just like what we did for definitive lpad values).

riscv-elf.adoc
kito-cheng: That described under `The mapping symbol for instruction(`$x`)`, so I think it be compatible?

kito-cheng:
Changes:
- Apply @mylai-mtk's comment
- Drop the merge rule for merging zero and non-zero value.

riscv-elf.adoc
kito-cheng: Let me just drop this, I think this leave to special linker relaxation

mylai-mtk:
With the recent updates to force RELRO and BIND_NOW so software-guarded jumps are used for unlabeled scheme LPAD PLTs (riscv-non-isa/riscv-elf-psabi-doc#417), I would like to discuss the possibility of also using `0` as the LPAD labels in LPAD PLTs for the func-sig scheme.

# Intuition

I assume that the [AUIPC, LD, JALR] insn sequences in LPAD PLTs are trusted not to be part of JOP gadgets (after all, it's with this trust that we can use software guarded jumps there), so why not extend this trust to the `LPAD 0` insn?

# Pros

Greatly simplifies the implementation of dynamic linking under the func-sig scheme, since if LPAD labels in LPAD PLTs are fixed, we don't need to pass the labels with the `.riscv.lpadinfo` section and all those policies to handle bad programs (that is, the merging rules of definitive and referencing LPAD labels) can be removed. This means simpler implementation in the toolchain (compilers, linkers and tools), and also less maintenance work on the spec side.

# Cons

+ Allow all functions referenced by an PLT to be indirectly callable without an X7 setup, no matter whether they have an LPAD insn or not.
+ Increase the possibility that func-sig scheme LPAD PLT entries are used as JOP gadgets. (However, since PLT entries are mostly identical, this would only introduce 1 kind of possible JOP gadget and that is which we already assumed is safe.)
+ Programmers now need to make sure they're not redefining symbols with different signatures, since tools can't check for them anymore.
+ Some other cons? (I don't know.)

# My Thoughts

I know that opening up all PLT-referenced functions to be indirectly callable without an X7 setup maybe a bit too much, but the simplification to the func-sig scheme's assembly programming model is, well, very lucrative to me (a toolchain and spec developer). So I decided perhaps the idea is valuable and still throw the issue out.

PR #461, opened 2025-03-26
Relocation Sequencing

lenary:
The rules around paired and marker relocations are reasonably complex, so this adds a note about what could be expected for a single offset.

This change also clarifies a slight difference between "paired" relocations (where you must have both, and they work together), and "associated" relocations which effectively modify or clarify the behaviour of the relocations at a specific offset. I classify R_RISCV_RELAX and R_RISCV_VENDOR as "associated" relocations, as in the case of the former, R_RISCV_RELAX is not required, and in the case of the latter, the same R_RISCV_VENDOR "marker" is used with the 64 nonstandard relocations.

The only new rule here is that if you are using paired relocations, then if one of the relocations is using a specific vendor identifier, then the other must use the same vendor identifier.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -528,7 +528,7 @@ Description:: Additional information about the relocation
                                             <| S + A - P
 .2+| 46-50   .2+| *Reserved*    .2+| -       |                   .2+| Reserved for future standard use
                                             <|
-.2+| 51      .2+| RELAX         .2+| Static  |                   .2+| Instruction can be relaxed, paired with a normal relocation at the same address
+.2+| 51      .2+| RELAX         .2+| Static  |                   .2+| Instruction can be relaxed, associated with a normal relocation at the same address
                                             <|
 .2+| 52      .2+| SUB6          .2+| Static  | _word6_           .2+| Local label subtraction
                                             <| V - S - A
@@ -560,7 +560,7 @@ Description:: Additional information about the relocation
                                             <|
 .2+| 66-190  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
                                             <|
-.2+| 191     .2+| VENDOR        .2+| Static  |                   .2+| Paired with a vendor-specific relocation and must be placed immediately before it, indicates which vendor owns the relocation.
+.2+| 191     .2+| VENDOR        .2+| Static  |                   .2+| Associated with a vendor-specific relocation and must be placed immediately before it, indicates which vendor owns the relocation.
                                             <|
 .2+| 192-255 .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for nonstandard ABI extensions
                                             <|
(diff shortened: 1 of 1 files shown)


Lapshin:
> The only new rule here is that if you are using paired relocations, then if one of the relocations is using a specific vendor identifier, then the other must use the same vendor identifier.

Doesn't it conflict with:

> Vendor identifiers must be unique amongst all vendors providing custom
relocations. Vendor identifiers may be suffixed with a tag to provide extra
relocations for a given vendor.

For example, it may be a case where the two identifiers "vendorX" and "vendorX_tag"

Lapshin:
@lenary , maybe you know somebody who can take a look at GNU binutils patch https://sourceware.org/pipermail/binutils/2025-April/140484.html ?

I'm going to apply the patch for Espressif's toolchain, but it would be great to have the vendor-specific relocations feature in the upstream

lenary:
> > The only new rule here is that if you are using paired relocations, then if one of the relocations is using a specific vendor identifier, then the other must use the same vendor identifier.
> 
> Doesn't it conflict with:
> 
> > Vendor identifiers must be unique amongst all vendors providing custom
> > relocations. Vendor identifiers may be suffixed with a tag to provide extra
> > relocations for a given vendor.
> 
> For example, it may be a case where the two identifiers "vendorX" and "vendorX_tag"

This is absolutely intentional, hence me describing this as a new rule. The problem is if you look for a vendor relocation at a section offset, you cannot tell which one applies to which item in the pair. It's much easier to just require both vendor identifiers are the same.

If you only have one relocation value left in your vendorX identifier (the rest are allocated), and you're adding a paired relocation, you should put both paired relocations into "vendorX_tag".

lenary:
I think @Nelson1225 might be able to look at your binutils patch?

I'm glad to see we're not the only vendor using vendor relocations. Have Espressif published their psABI vendor extensions as a document, or just as a patch?

Lapshin:
@lenary , it's just a patch for now. The extension docs haven't been published yet, but I hope they will be soon

yclwlcy:
Andes has a use case where relocations are used to mark the beginning and end of specific regions. These regions indicate where optimizations, such as zcmt relaxation, are optionally applied depending on linker options (avoiding the need to provide two versions of an object). The relocations used for this purpose are not "paired", but they may appear at the same offset as a paired relocation. 
 
This can lead to scenarios where more than four relocations exist at a single offset, which appears to go beyond the current sequencing rules. I wonder if the limit of four relocations per offset is intended to be strictly enforced, or whether there is room for flexibility—particularly for those relocations that are neither paired nor associated.

lenary:
> Andes has a use case where relocations are used to mark the beginning and end of specific regions. These regions indicate where optimizations, such as zcmt relaxation, are optionally applied depending on linker options (avoiding the need to provide two versions of an object). The relocations used for this purpose are not "paired", but they may appear at the same offset as a paired relocation.
> 
> This can lead to scenarios where more than four relocations exist at a single offset, which appears to go beyond the current sequencing rules. I wonder if the limit of four relocations per offset is intended to be strictly enforced, or whether there is room for flexibility—particularly for those relocations that are neither paired nor associated.

You're right, I overlooked nonstandard marker relocations, which can also apply to a location without changing it. Making the nonstandard marker relocations having to share a vendor identifier with (potentially paired) nonstandard relocations at the same location does seem like a drawback, given there's no limit to how many markers there can be on a location (whereas before I thought there was a limit of 2, for paired relocations).

I think this PR cannot proceed, as the clarifying example is evidently not actually correct.

PR #470, opened 2025-07-01
Add initial specification for Big Endian

djtodoro:
GNU GCC Toolchain already supports big endian for RISC-V target. That support was merged without a change in psABI Document.
Here [0] is the initial PR for adding big endian support in LLVM project, so lets implement documentation part as well.

[0] https://github.com/llvm/llvm-project/pull/146534


rui314:
I think we need to clarify which relocations write data in big-endian order when the output is big-endian. My understanding is as follows:

R_RISCV_{32,64}
R_RISCV_ADD{16,32,64}
R_RISCV_SUB{16,32,64}
R_RISCV_SET{16,32,64}
R_RISCV_32_PCREL
R_RISCV_PLT32

kito-cheng:
@djtodoro do you have plan to update the PR according @rui314's comment?

riscv-cc.adoc
```diff
@@ -167,8 +167,10 @@ callee-saved registers must be restored to what was set on entry, and the
 contents of any fixed registers like `gp` and `tp` never change.
 
 
-NOTE: Calling convention for big-endian is *NOT* included in this specification
-yet, we intend to define that in future version of this specification.
+NOTE: Big-endian calling conventions follow the same rules as little-endian
+calling conventions. The only difference is in the byte ordering of multi-byte
+values in memory and registers. Register usage, argument passing, and return
+value conventions remain the same.
```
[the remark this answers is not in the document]
kito-cheng: It's a draft for that: we already pass 2×XLEN bits scalar in what we expected:
```
calars that are 2×XLEN bits wide are passed in a pair of argument registers,
with the low-order XLEN bits in the lower-numbered register and the high-order
XLEN bits in the higher-numbered register.  If no argument registers are
available, the scalar is passed on the stack by value.  If exactly one
register is available, the low-order XLEN bits are passed in the register and
the high-order XLEN bits are passed on the stack.
```

So I tried to add a paragraph to clarify also give an example for that, also adding a NOTE to describe the rationale.

The other thing I added is for Variadic arguments with 2×XLEN-bit and Aggregates with XLEN < size <= XLEN *2.

I didn't check with GCC implementation yet, but IIRC that's may not match GCC's default big-endian behavior.

```diff
diff --git a/riscv-cc.adoc b/riscv-cc.adoc
index 0768360..037b47f 100644
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -191,6 +191,16 @@ available, the scalar is passed on the stack by value.  If exactly one
 register is available, the low-order XLEN bits are passed in the register and
 the high-order XLEN bits are passed on the stack.
 
+This register-pair ordering is defined in terms of value significance and is
+independent of endianness.  For example, on RV32BE a 64-bit scalar returned
+in a0/a1 places bits [31:0] (the least-significant XLEN bits) in a0 and
+bits [63:32] in a1; memory layout remains big-endian.
+
+NOTE: Defining the register-pair ordering independent of endianness allows
+RV32_Zdinx and Zilsd paired load/store paths to be used directly for argument
+passing and return without extra swaps.  Memory layout remains governed by the
+target endianness.
+
 Scalars wider than 2×XLEN bits are passed by reference and are replaced in the
 argument list with the address.
 
@@ -198,7 +208,10 @@ Aggregates whose total size is no more than XLEN bits are passed in
 a register, with the fields laid out as though they were passed in memory. If
 no register is available, the aggregate is passed on the stack.
 Aggregates whose total size is no more than 2×XLEN bits are passed in a pair
-of registers; if only one register is available, the first XLEN bits are passed
+of registers with the fields laid out as though they were passed in memory:
+the lower-numbered register holds the lower-addressed XLEN-sized chunk of
+the aggregate and the higher-numbered register holds the next chunk;
+if only one register is available, the first XLEN bits are passed
 in a register and the remaining bits are passed on the stack. If no registers are
 available, the aggregate is passed on the stack. Bits unused due to
 padding, and bits past the end of an aggregate whose size in bits is not
@@ -231,7 +244,10 @@ same manner as named arguments, with one exception.  Variadic arguments with
 even-numbered), or on the stack by value if none is available. After a
 variadic argument has been passed on the stack, all future arguments will also
 be passed on the stack (i.e. the last argument register may be left unused due
-to the aligned register pair rule).
+to the aligned register pair rule).  For 2×XLEN scalars placed in an aligned
+register pair, the lower-numbered register holds the least-significant XLEN bits
+and the higher-numbered register holds the most-significant XLEN bits,
+regardless of endianness.
 
 Values are returned in the same manner as a first named argument of the same
 type would be passed.  If such an argument would have been passed by
```

djtodoro:
@kito-cheng Thanks, I agree!

>I didn't check with GCC implementation yet, but IIRC that's may not match GCC's default big-endian behavior.

We will check GCC implementation, and fix it there.

riscv-elf.adoc
```diff
@@ -236,6 +236,49 @@ type-name = identifier-nondigit *identifier-char
...
+* `R_RISCV_ADD32` - 32-bit label addition
+* `R_RISCV_ADD64` - 64-bit label addition
+* `R_RISCV_SUB16` - 16-bit label subtraction
+* `R_RISCV_SUB32` - 32-bit label subtraction
+* `R_RISCV_SUB64` - 64-bit label subtraction
+* `R_RISCV_SET16` - 16-bit local label assignment
+* `R_RISCV_SET32` - 32-bit local label assignment
+* `R_RISCV_SET64` - 64-bit local label assignment
+* `R_RISCV_32_PCREL` - 32-bit PC-relative relocation
+* `R_RISCV_PLT32` - 32-bit relative offset to PLT entry
```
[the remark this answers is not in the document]
djtodoro: It does make sense. I realized that, and already removed it locally... Will be included in the next version of patch. Thanks!

kito-cheng:
@aswaterman could you take a look on the big-endian calling convention part :)

djtodoro:
> @kito-cheng Thanks, I agree!
> 
> > I didn't check with GCC implementation yet, but IIRC that's may not match GCC's default big-endian behavior.
> 
> We will check GCC implementation, and fix it there.

Okay. For this basic test case:

```bash
$ cat test.c
long long test()
{
  return 0x1;
}
```
GCC for LE generates:

```bash
$ riscv64-unknown-linux-gnu-gcc -c test.c -O2 -march=rv32gc -mabi=ilp32
$ riscv64-unknown-linux-gnu-objdump -d test.o

test.o:     file format elf32-littleriscv


Disassembly of section .text:

00000000 <test>:
   0: 4505                 li a0,1
   2: 4581                 li a1,0
   4: 8082                 ret
```

And for BE, it generates:
```bash
$ riscv64-unknown-linux-gnu-gcc -c test.c -O2 -march=rv32gc -mabi=ilp32 -mbig-endian
$ riscv64-unknown-linux-gnu-objdump -d test.o

test.o:     file format elf32-bigriscv


Disassembly of section .text:

00000000 <test>:
   0: 4585                 li a1,1
   2: 4501                 li a0,0
   4: 8082                 ret
```

So, basically it does not follow the proposal here. We managed to come up with a fix, but needs some extra testing (https://github.com/djtodoro/gcc/commit/71a0f9fc4bf9ff1b92ac434e362261ed16ff396b), but with that applied, we now have:

```bash
$ riscv64-unknown-linux-gnu-gcc -c test.c -O2 -march=rv32gc -mabi=ilp32 -mbig-endian
$ riscv64-unknown-linux-gnu-objdump -d test.o

test.o:     file format elf32-bigriscv


Disassembly of section .text:

00000000 <test>:
   0: 4581                 li a1,0
   2: 4505                 li a0,1
   4: 8082                 ret
```

aswaterman:
I haven't had time to think this through yet, but make sure whatever you propose does the right thing for variadic functions.  In particular, you want the argument-register layout to match the memory layout of arguments passed on the stack.  This might encourage you to stick with GCC's current implementation, rather than making the change that @djtodoro mentioned.

djtodoro:
hi @kito-cheng, is there any comment on this?

aswaterman:
@djtodoro were you able to run the GCC test suite with similar failure rate to little endian?  (And if so, with which C library, RISC-V simulator, etc.?)

djtodoro:
> @djtodoro were you able to run the GCC test suite with similar failure rate to little endian? (And if so, with which C library, RISC-V simulator, etc.?)

Hi @aswaterman.

Yes, we've run the GCC test suite and the failure rates for BE vs LE are pretty similar, using `newlib` as the C library. The main noise comes from RVV-related .exp tests that implicitly depend on the V extension, which isn't supported for BE yet — setting those aside, the results are comparable. I will share exact GCC test suite results for both endianess.

 We tested with both qemu and GNU Sim. Some of the fixes/changes we contributed for big-endian support:

  - GCC: https://gcc.gnu.org/pipermail/gcc-patches/2026-February/708243.html
  - GDB (GNU Sim): https://sourceware.org/pipermail/gdb-patches/2025-December/223463.html
  - Toolchain PR: https://github.com/riscv-collab/riscv-gnu-toolchain/pull/1812
  - QEMU patch-set: https://patchew.org/QEMU/20260320150552.169480-1-djordje.todorovic@htecgroup.com/

We're also working on initial support in the LLVM ecosystem — clang and the llvm backend are already merged, and we're currently working on lld, lldb, libc, and compiler-rt.

djtodoro:
cc @MaskRay

aswaterman:
I lost track of where this is.  Does this PR match upstream GCC, and is upstream GCC functionally correct?

Or has the updated variadic function proposal not been integrated into this PR and/or GCC yet?

kito-cheng:
<del>The GCC implementation isn't update the for the register-pair/XLEN * 2 stuff yet, and I don't see any patch so far, I suspect that might be hard to implement, so I consider we may rollback on that part, however I need more feedback on that part.</del>

<del>I thought https://gcc.gnu.org/pipermail/gcc-patches/2026-February/708243.html is the missing part, but I figured out it's not after I review that in weeks ago.</del>

UPDATE: I missed few comment above, let me read and reply later

djtodoro:
> I lost track of where this is. Does this PR match upstream GCC, and is upstream GCC functionally correct?
> 
> Or has the updated variadic function proposal not been integrated into this PR and/or GCC yet?

Hi @andrew, basically upstream GCC does not follow psABI for BE and here is comment why https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/470#issuecomment-3307066504.
I have not proceed with sharing the patches for GCC yet (like this one https://github.com/djtodoro/gcc/commit/71a0f9fc4bf9ff1b92ac434e362261ed16ff396b), because we should agree how are we going to define psABI for that specific case (and for the variadic functions as well) -- basically this https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/470#issuecomment-3461844432 mentions how it should look like.

As soon as we have an agreement, I will go and share concrete patches with GCC community.

aswaterman:
The P extension proposal and the ratified Zilsd extension both have instructions that handle even-odd register-pair operands. Regardless of endianness, they put the LSBs in the even-numbered register and the MSBs in the odd-numbered register. The ABI should match this scheme for named arguments to avoid extra moves. As long as we accomplish that goal, and as long as we get the variadic case correct (regardless of performance), I’d be happy. I believe your proposal checks both boxes?

djtodoro:
> The P extension proposal and the ratified Zilsd extension both have instructions that handle even-odd register-pair operands. Regardless of endianness, they put the LSBs in the even-numbered register and the MSBs in the odd-numbered register. The ABI should match this scheme for named arguments to avoid extra moves. As long as we accomplish that goal, and as long as we get the variadic case correct (regardless of performance), I’d be happy. I believe your proposal checks both boxes?

Yes, that is the idea, I agree completely. I will go once again through psABI content, but the goal is to cover exactly that what you described (plus to cover clearly variadic case).

djtodoro:
Yes, I believe the proposal checks both boxes.

I went through this again and implemented a GCC patch for the RV32BE case here [0].

The intended behavior is:

  - Named 2×XLEN scalar arguments and return values use significance-based ordering, independent of endianness. So for RV32BE, the lower-numbered register holds the low-order XLEN bits and the higher-numbered register holds the high-order XLEN bits. This matches the P/Zilsd even/odd register-pair convention and avoids extra moves.

  - Variadic 2×XLEN scalar arguments keep memory-layout ordering on big-endian targets, so that spilling the argument registers produces the same layout as stack-passed arguments. This keeps `va_arg` correct.

I also added GCC tests covering named scalar arguments, return values, variadic arguments, and aggregates. I will prepare and send this GCC patch to the GCC community once we agree on the psABI wording.


Also, I am changing the psABI PR here, to reflect these changes.

Please note that I have not shared [0] with GCC community yet, but will do.

[0] https://github.com/aleksapaunovic-htec/gcc/commit/57ef1050a1465cc298bc2fbd07c1fe6b0fad13bb

riscv-elf.adoc
```diff
@@ -264,6 +264,56 @@ type-name = identifier-nondigit *identifier-char
...
+* `R_RISCV_64`
+* `R_RISCV_ADD16`
+* `R_RISCV_ADD32`
+* `R_RISCV_ADD64`
+* `R_RISCV_SUB16`
+* `R_RISCV_SUB32`
+* `R_RISCV_SUB64`
+* `R_RISCV_SET16`
+* `R_RISCV_SET32`
+* `R_RISCV_SET64`
```
kito-cheng: We don't have R_RISCV_SET64
```suggestion
```

riscv-elf.adoc
```diff
@@ -264,6 +264,56 @@ type-name = identifier-nondigit *identifier-char
...
+* `R_RISCV_32_PCREL`
+* `R_RISCV_PLT32`
+* `R_RISCV_JUMP_SLOT`
+* `R_RISCV_TLS_DTPMOD32`
+* `R_RISCV_TLS_DTPMOD64`
+* `R_RISCV_TLS_DTPREL64`
+* `R_RISCV_TLS_DTPREL32`
+* `R_RISCV_TLS_TPREL64`
+* `R_RISCV_TLS_TPREL32`
```
kito-cheng, line 919:
```suggestion
* `R_RISCV_TLS_TPREL32`
* `R_RISCV_RELATIVE`
* `R_RISCV_IRELATIVE`
* `R_RISCV_GOT32_PCREL`
* `R_RISCV_TLSDESC`
```

riscv-elf.adoc
```diff
@@ -264,6 +264,56 @@ type-name = identifier-nondigit *identifier-char
...
+
+=== Byte Ordering
+
+RISC-V implementations can support either little-endian or big-endian byte ordering:
+
+* **Little-endian**: The least significant byte is stored at the lowest memory address.
+  This is the default and most common byte ordering for RISC-V.
+* **Big-endian**: The most significant byte is stored at the lowest memory address.
+
+The endianness is fixed at the system level and cannot be changed dynamically.
```
kito-cheng: While changing endianness in a single program doesn't make sense, however this statement is too broad, so I suggest removing it.

Also [1] say: `MBE`, `SBE`, and `UBE` bits in `mstatus` and `mstatush` are `WARL`, so it actually able to change dynamically. 

[1] https://github.com/riscv/riscv-isa-manual/blob/main/src/priv/machine.adoc#endianness-control-in-csrmstatus-and-csrmstatush-registers


```suggestion
```

riscv-cc.adoc
```diff
@@ -254,7 +268,16 @@ same manner as named arguments, with one exception.  Variadic arguments with
...
+to the aligned register pair rule).  For 2×XLEN variadic scalars placed in an
+aligned register pair, the register assignment follows memory layout ordering:
+the lower-numbered register receives the XLEN bits from the lower memory address
+and the higher-numbered register receives the XLEN bits from the higher memory
+address. This ensures correct `va_arg` operation when arguments are spilled to
+the stack.
+
+NOTE: This memory-layout ordering for variadic arguments differs from the
+significance-based ordering used for named arguments and return values on
+big-endian systems.
```
kito-cheng: I would like to extend further here, one more paragraph and one more table rather than a NOTE, the motivation is including `_BitInt`, which is not very intuitive part, also put a table should be make it express more clearly :)

```suggestion
The following table summarizes the register-pair ordering for a value that
occupies an aligned 2×XLEN register pair.  Two orderings are distinguished:
_significance ordering_, in which the lower-numbered register holds the
low-order XLEN bits, and _memory-layout ordering_, in which the lower-numbered
register holds the XLEN bits at the lower memory address.  On little-endian
these two orderings coincide; on big-endian they are opposite.

[cols="4,3,3"]
|===
| Value in an aligned 2×XLEN register pair | Named argument / return value | Variadic argument

| 2×XLEN scalar
| significance ordering
| memory-layout ordering

| `_BitInt(N)` with XLEN < N ≤ 2×XLEN (passed under the 2×XLEN-scalar rules)
| significance ordering
| memory-layout ordering

| Aggregate no larger than 2×XLEN bits
| memory-layout ordering
| memory-layout ordering
|===

Variadic values always use memory-layout ordering so that register-passed and
stack-passed variadic arguments share the same in-memory layout, which is
required for correct `va_arg` operation.

NOTE: On big-endian systems, only scalar-class values (2×XLEN scalars and
`_BitInt` passed under the scalar rules) change ordering between the named and
variadic paths.  Aggregates use memory-layout ordering in both cases.
```

kito-cheng (commented):
I take another round of deep review, and this generally OK, just 2 minor comments and one big suggestion chunk for named argument / return value and variadic argument.

riscv-elf.adoc
djtodoro: I agree.

riscv-cc.adoc
djtodoro: It does make sense! Thanks!

djtodoro:
> I take another round of deep review, and this generally OK, just 2 minor comments and one big suggestion chunk for named argument / return value and variadic argument.

Thanks a lot for the feedback. I addressed the comments.

riscv-elf.adoc
aswaterman: It's fine to remove this sentence, but the ABI should obviously assume that endianness is fixed for a given program.  The fact that endianness is dynamic is a system-level issue, not an ABI issue.

djtodoro: I think that we should clarity this in the document, so I added that into it. Thanks.

djtodoro:
@kito-cheng @aswaterman thanks for your comments! please look into latest changes :)

riscv-elf.adoc
```diff
@@ -2760,4 +2813,4 @@ https://github.com/riscv/riscv-code-size-reduction
 https://github.com/riscv-non-isa/riscv-toolchain-conventions
 
 * [[[rvy-spec]]] "RISC-V Specification for CHERI Extensions"
-https://riscv.github.io/riscv-cheri/
\ No newline at end of file
+https://riscv.github.io/riscv-cheri/
```
kito-cheng: I guess this is changed accidentally?

riscv-cc.adoc
```diff
@@ -1057,4 +1108,3 @@ https://github.com/riscv/riscv-v-spec
 
 * [[[rvv-intrinsic-doc]]] "RISC-V Vector Extension Intrinsic Document"
 https://github.com/riscv-non-isa/rvv-intrinsic-doc
-
```
kito-cheng, line 1060: Keep this unchanged

kito-cheng (commented):
Generally LGTM to me, I will propose move this forward in next psABI meeting to include this as experimental/draft ABI like rv64ilp32 :)

riscv-elf.adoc
djtodoro: ouch, yes

djtodoro:
> Generally LGTM to me, I will propose move this forward in next psABI meeting to include this as experimental/draft ABI like rv64ilp32 :)

Thanks a lot!

djtodoro:
Hey @kito-cheng, just checking if this now can be merged :)

djtodoro:
Ping :)

kito-cheng:
We discussed this on last psABI meeting, the consensus is OK to merge this as experimental ABI, but we would like @aswaterman to give a explicitly LGTM here before we merge this :)

kito-cheng (approved):
And it's my LGTM, @djtodoro thanks for pushing this forward!

djtodoro:
> And it's my LGTM, @djtodoro thanks for pushing this forward!

Thank you!!

djtodoro:
Any additional feedback here? :)

kito-cheng:
@aswaterman did you mind give an explicitly LGTM here? since you and me are give the most review, so I would like to make sure it's ok to you :)

aswaterman (approved):
@kito-cheng Yeah, I'd go forward with it.  I have a suspicion that during implementation we will discover that we got something wrong and will need to revise this spec, but it is marked "draft", after all :)

PR #471, opened 2025-07-14
Account for flexible array members in the integer calling convention.

resistor:
This appears to have been unspecified previously. Example miscompilation in LLVM reported here: https://github.com/llvm/llvm-project/issues/148536

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -209,7 +209,8 @@ type alignment and XLEN bits, but never more than the stack alignment.
 
 Aggregates larger than 2×XLEN bits are passed by reference and are replaced in
 the argument list with the address, as are {Cpp} aggregates with nontrivial copy
-constructors, destructors, or vtables.
+constructors, destructors, or vtables. Aggregates containing C99 flexible array
+members are similarly always passed by reference.
 
 Fixed-length vectors are treated as aggregates.
 


resistor:
Exercising this is tricky, and the particular example I extracted the bug from used an exotic combination of:

- C++17 guaranteed return copy elision
- C99 flexible array members
- A caller written in assembly

Given that flexible array members don't formally exist in C++, this is all outside of the specification. However, the behavior I am proposing here seems like the "least surprising" option.

As evidence that this interpretation is precedented, similar logic appears in the following other LLVM targets:
-  Lanai: https://github.com/llvm/llvm-project/blob/6d4a50272f19c1db65dcf0349b6905f12442f80b/clang/lib/CodeGen/Targets/Lanai.cpp#L105
- ARC: https://github.com/llvm/llvm-project/blob/6d4a50272f19c1db65dcf0349b6905f12442f80b/clang/lib/CodeGen/Targets/ARC.cpp#L115
- ARMv7 (I think?): https://github.com/llvm/llvm-project/blob/6d4a50272f19c1db65dcf0349b6905f12442f80b/clang/lib/CodeGen/Targets/ARM.cpp#L521
- SystemZ: https://github.com/llvm/llvm-project/blob/6d4a50272f19c1db65dcf0349b6905f12442f80b/clang/lib/CodeGen/Targets/SystemZ.cpp#L460
- X86: https://github.com/llvm/llvm-project/blob/6d4a50272f19c1db65dcf0349b6905f12442f80b/clang/lib/CodeGen/Targets/X86.cpp#L510

jrtc27:
I can't see in the equivalent part of AAPCS64 anything that would cover this case. B.2 says:

> If the argument type is a Composite Type whose size cannot be statically determined by both the caller and the callee, the argument is copied to memory and the argument is replaced by a pointer to the copy. (There are no such types in C/C++ but they exist in other languages or in language extensions).

But that claims it only applies to non-C/C++ languages. Probably that's the one that should be covering flexible array members though, i.e. by removing the parenthesised aside.

lenary:
I filed a ticket against aapcs: ARM-software/abi-aa#339 thanks for the references.

riscv-cc.adoc
```diff
@@ -209,7 +209,8 @@ type alignment and XLEN bits, but never more than the stack alignment.
 
 Aggregates larger than 2×XLEN bits are passed by reference and are replaced in
 the argument list with the address, as are {Cpp} aggregates with nontrivial copy
-constructors, destructors, or vtables.
+constructors, destructors, or vtables. Aggregates containing C99 flexible array
+members are similarly always passed by reference.
```
lenary, line 213:
```suggestion
constructors, destructors, or vtables. Aggregates of unknown size, for instance
C99 structs containing flexible array members, are always passed by reference.
```

This might be more future-proof, but I'm not sure "of unknown size" is the best way to explain this.

kito-cheng:
The argument part seems reasonable to apply this rule, but I am a little hesitant about the return value part, because the buffer of the return value is reserved by the caller, and in the current example, the caller cannot specify how much space to reserve for `c`, so the default is that buffer overflow if we passed by reference, and the current behavior of directly removing c is actually safe.

resistor:
> > In the case this was extracted from, the caller and callee agreed on the size of the flexible array member via a compile-time constant. More generally, the caller could pass an integer argument to the callee that specifies the dynamic size of the flexible array member.
> 
> Could you provide testcase to showing that case in C/C++? w/ clang and/or GNU extension is fine to me, I just could not imaging how to write that.

The original code I extracted this from had a caller written in assembly, and a `#define` from the build system was used to ensure that the caller and callee always agreed on the space allocated for the flexible array member.

Here is a somewhat contrived example of doing the same thing in C++ with GNU extensions:

```
#include <cstdlib>

struct FooStruct {
    long a;
    long b;
    long c[];
};

FooStruct foo(int c) {
    FooStruct ret;

    ret.a = 0;
    ret.b = 0;
    for (int i = 0; i < c; ++i)
        ret.c[i] = 0x535;

    return ret;
}

FooStruct *bar() {
    int c = 2;
    FooStruct *s = (FooStruct*)malloc(sizeof(FooStruct) + c*sizeof(long));
    *s = foo(c);
    return s;
}
```

jrtc27:
> > > In the case this was extracted from, the caller and callee agreed on the size of the flexible array member via a compile-time constant. More generally, the caller could pass an integer argument to the callee that specifies the dynamic size of the flexible array member.
> > 
> > 
> > Could you provide testcase to showing that case in C/C++? w/ clang and/or GNU extension is fine to me, I just could not imaging how to write that.
> 
> The original code I extracted this from had a caller written in assembly, and a `#define` from the build system was used to ensure that the caller and callee always agreed on the space allocated for the flexible array member.
> 
> Here is a somewhat contrived example of doing the same thing in C++ with GNU extensions:
> 
> ```
> #include <cstdlib>
> 
> struct FooStruct {
>     long a;
>     long b;
>     long c[];
> };
> 
> FooStruct foo(int c) {
>     FooStruct ret;
> 
>     ret.a = 0;
>     ret.b = 0;
>     for (int i = 0; i < c; ++i)
>         ret.c[i] = 0x535;

^ UB if c > 0

> 
>     return ret;
> }
> 
> FooStruct *bar() {
>     int c = 2;
>     FooStruct *s = (FooStruct*)malloc(sizeof(FooStruct) + c*sizeof(long));
>     *s = foo(c);

I strongly expect the semantics of this to be initialising just the non-flexible portion, regardless of whether copy elision is happening.

>     return s;
> }
> ```

I don't think this example works on AArch64 either. For us to consider making an ABI-breaking change like this we would have to have a statement that GNU C++ defines the semantics to be that, for this edge case, the entire flexible member is assigned. I believe this is not something that is the case, and that the code in question has been operating outside of GNU C++'s semantics, relying on implementation and ABI details that aren't universally true.

resistor:
> I don't think this example works on AArch64 either (https://godbolt.org/z/KnGEscP1c). For us to consider making an ABI-breaking change like this we would have to have a statement that GNU C++ defines the semantics to be that, for this edge case, the entire flexible member is assigned. I believe this is not something that is the case, and that the code in question has been operating outside of GNU C++'s semantics, relying on implementation and ABI details that aren't universally true.

This is somewhat circular. When reported to LLVM, this document was referenced as the source of truth for this on RISC-V (https://github.com/llvm/llvm-project/issues/148536) so I believe it is up to this group to decide. If the decision is that it is working as intended today, I would suggest that clarifying language to that effect be added.

With respect to other targets, I am aware that AArch64 does not work this way. AArch32 (https://godbolt.org/z/vPa9466bq), Mips (https://godbolt.org/z/Ya3hTf7Ws) and X86-64 (https://godbolt.org/z/4fMrr98TW) appear to be implementation-dependent: LLVM preserves the writes to the flexible array member, while GCC does not. I provided links to some other targets in LLVM [here](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/471#issuecomment-3068764469) that appear to preserve the writes.

Personally I find passing and returning the flexible array member structs by reference to be the least-surprising solution to this implementation-dependent situation. However, it's clearly not the end of the world in that the construct is inherently non-portable, is clearly not portable even across compilers on the same architecture in practice, and the current behavior does match one of the common implementation choices across compilers/targets.

AaronBallman:
I think there's two cases: typical FAM use (dynamic allocation is necessary because the FAM is non-empty) and atypical FAM use (empty FAM).

For typical FAM use:
https://eel.is/c++draft/class.copy.elision#1.1

An object with a FAM can never have automatic storage duration unless the FAM is unused. So for typical FAM use, I don't think it should be eligible for guaranteed copy elision. The point to copy elision is that the caller can allocate enough space to hold the object so that the return statement can be constructed in-place. That's not really possible with a FAM unless it's a degenerate case where it has no elements.

For atypical FAM use:
It's not occupying any storage and elision would be possible, but why special case the atypical FAM use?

kito-cheng:
I'm currently leaning toward not accepting this change, and here's why:

- It’s an ABI-incompatible change, and we should really avoid those unless it's absolutely necessary.
- C++17 does guarantee return copy elision, but that guarantee doesn’t exist in C++ versions before 17 — or in C.
- Even in C++17, the guarantee only applies to prvalue returns, not all return cases.
- From a compiler implementation perspective, outside of C++17’s guaranteed prvalue cases, we can’t rely on copy elision always happening. That means we could end up with differences between -O0 and -O[1-3].

References:
[1] https://en.cppreference.com/w/cpp/language/copy_elision
[2] https://timsong-cpp.github.io/cppwp/n4659/class.copy.elision

kito-cheng:
We have discussed this topic several times in the psABI meetings, and the consensus in those meetings was to reject this proposal. My previous comment also included the main reasons, so I am closing this PR. However, I’d still like to sincerely thank @resistor for bringing up this proposal and sparking such a valuable discussion.

PR #474, opened 2025-08-11
Add Tag_RISCV_mop_and_hint_encoding for instruction reinterpretation

kito-cheng:
Various RISC-V extensions reuse existing instruction encoding spaces
for specialized behaviors. CFI extensions use NOP or MOP (Maybe Operation)
instructions to ensure correct program execution even when hardware 
doesn't implement CFI. Similarly, hint extensions redefine specific
encodings for performance optimization purposes.

However, this design conflicts with current toolchain conventions,
where X extension instructions are only usable when the X extension
is enabled: assemblers accept corresponding mnemonics and disassemblers
decode instructions only when the extension is active.

The affected use cases require different behavior:
- Zicfiss instructions should be available when Zimop extension is present
- Zicfilp instructions should be usable even without explicit enabling
  (since auipc is in baseline ISA)
- Hint extensions should reinterpret specific encodings regardless of
  extension enablement

Following conventional toolchain behavior would force compilers to
generate generic instructions ('mop', 'fence w,0', 'auipc x0, <value>')
instead of extension-specific mnemonics ('sspush/sspop', 'pause', 
'lpad <value>'), causing user confusion and reducing code clarity.

While this could be addressed at the ISA specification level, such
changes involve complex and time-consuming procedures. Therefore,
this psABI defines Tag_RISCV_mop_and_hint_encoding to provide toolchain
implementations with clear guidance.

This tag uses a bitmap format where each bit indicates whether specific
instruction encodings should be reinterpreted:
- Bit 0: auipc x0, <value> as lpad <value> (Zicfilp)
- Bit 1: Zimop encodings as Zicfiss instructions
- Bit 2: fence w,0 as pause hints (Zihintpause)
- Bit 3: specific encodings as non-temporal locality hints (Zihintntl)

The tag uses bitwise OR merge policy to allow combining multiple
reinterpretation requirements, with toolchain warnings recommended
when encoding space conflicts occur.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1272,6 +1272,7 @@ non-standard ABI extensions.
 | Tag_RISCV_priv_spec_revision        |       12 | uleb128        | *Deprecated*, indicates the revision version of the privileged specification.
 | Tag_RISCV_atomic_abi                |       14 | uleb128        | Indicates which version of the atomics ABI is being used.
 | Tag_RISCV_x3_reg_usage              |       16 | uleb128        | Indicates the usage definition of the X3 register.
+| Tag_RISCV_mop_and_hint_encoding     |       18 | uleb128        | Indicates the reinterpretation of MOP and hint instruction encodings.
 | Reserved for non-standard attribute | >= 32768 | -              | -
 |===
 
@@ -1436,6 +1437,34 @@ combined. However, an exception exists: the value `0` can merge with `1` or `2`
 value. After the merge, the resulting value will be the non-zero one.
 
 --
+
+===== Tag_RISCV_mop_and_hint_encoding, 18, uleb128=value
+
+Tag_RISCV_mop_and_hint_encoding indicates the reinterpretation of MOP (Maybe Operation) 
+and hint instruction encodings within this object. The value is a bitmap where each bit 
+indicates whether specific instruction encodings should be reinterpreted:
+
+[horizontal]
+Bit 0 (Zicfilp):: When set, `auipc x0, <value>` instructions are interpreted as `lpad <value>`.
+Bit 1 (Zicfiss):: When set, Zimop encoding space instructions are interpreted as Zicfiss instructions.
+Bit 2 (Zihintpause):: When set, `fence w,0` instructions are interpreted as `pause` hints.
+Bit 3 (Zihintntl):: When set, specific instruction encodings are interpreted as non-temporal locality hints.
+
+--
+
+Merge policy:::
+Values are merged using bitwise OR operation.
+
+NOTE: The bitwise OR merge policy may cause disassembler confusion when the same encoding 
+space is redefined by multiple bits. Toolchains may report warnings to alert users 
+when such conflicts occur.
+
+--
+
+NOTE: When specific bits are set, toolchains must disassemble instructions according to 
+the corresponding extensions regardless of whether these extensions are enabled. For example, 
+if bit 0 is set, `auipc x0, <value>` must be disassembled as `lpad <value>`; if bit 1 is 
+set, Zimop encodings must be disassembled as Zicfiss instructions.
 
 === Program Property
 
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1272,7 +1272,7 @@ non-standard ABI extensions.
 | Tag_RISCV_priv_spec_revision        |       12 | uleb128        | *Deprecated*, indicates the revision version of the privileged specification.
 | Tag_RISCV_atomic_abi                |       14 | uleb128        | Indicates which version of the atomics ABI is being used.
 | Tag_RISCV_x3_reg_usage              |       16 | uleb128        | Indicates the usage definition of the X3 register.
-| Tag_RISCV_mop_and_hint_encoding     |       18 | uleb128        | Indicates the reinterpretation of MOP and hint instruction encodings.
+| Tag_RISCV_mop_and_hint_encoding     |       18 | uleb128        | Indicates the reinterpretation of MOP and HINT instruction encodings.
 | Reserved for non-standard attribute | >= 32768 | -              | -
 |===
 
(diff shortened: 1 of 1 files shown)


kito-cheng:
cc @mylai-mtk @deepak0414 @jaidTw

riscv-elf.adoc
```diff
@@ -1437,6 +1438,29 @@ value. After the merge, the resulting value will be the non-zero one.
...
+===== Tag_RISCV_cfi_encoding, 18, uleb128=value
+
+Tag_RISCV_cfi_encoding indicates the reinterpretation of instruction encodings for Control 
+Flow Integrity extensions within this object. This covers both Zicfiss instructions using 
+Zimop encoding space and Zicfilp instructions using baseline ISA encodings 
+(specifically `auipc x0, <value>` as `lpad <value>`).
+
+[horizontal]
+0:: This object's reinterpretation of instruction encodings for CFI extensions is unknown.
+1:: This object contains instructions that must be interpreted according to Zicfiss/Zicfilp CFI extensions.
```
mylai-mtk, line 1450: Why combine the marks of Zicfiss and Zicfilp?

mylai-mtk: But they occupy different encoding spaces and can be enabled independently, so why force them to be "marked" together?

mylai-mtk: Yes, I found that merging them together indeed results in simpler conflict checking rules.

However, I would still like to raise the concern about the fact that Zicfiss occupies the MOP encoding space may one day result in the Zicfilp being selectively enabled without Zicfiss, if some valuable extension is to be redefined in the MOP encoding space and thus conflicts with Zicfiss. Besides, given that shadow stack has a software-based alternative implementation, substituting Zicfiss with the software alternative so MOP could be repurposed for some unique extension doesn't seem like a mad option to me.

kito-cheng:
> Would there be plans to incorporate some kind of framework at the ISA level? Based on the PR description, the current conflicts only arise from toolchain conventions, so I think maybe this psABI place _is_ the right place to solve this issue? Since after all the problems arise from toolchain implementations and decisions, which is technically independent from ISA.
> 
> If this is the right place to regulate the usages of these kind of "always-on" extensions, would it be better to provide a more generic framework than the currently proposed one that serves only the CFI features?

Another case is hint instructions (e.g. Zihintpause and Zihintntl). I originally thought they were different from the CFI situation, but yeah - they’re actually pretty similar. I'm still not like about adding a separate bit for every special case, but given what you said about the potential need to reallocate opcode space, having separate bits does sound like a more future-proof approach.

So let me update and made this PR more generic.

riscv-elf.adoc
kito-cheng: Addressed :)

kito-cheng:
Changes:
- Change to bitmap rather than just 0 or 1
- Included Zihintpause and Zihintntl

deepak0414:
Sounds like a good idea.
Consumer of this would be anything which is decoding instructions (disassembler, instruction/bin walking, etc)

Suggestion1: Instead of `Tag_RISCV_mop_and_hint_encoding`, I think better name would be `Tag_RISCV_Insn_Reinterpret`.

Suggestion2: Keep the highest bit reserved and consume/set it when we run out of 127 bits :-)

riscv-elf.adoc
```diff
@@ -1272,6 +1272,7 @@ non-standard ABI extensions.
 | Tag_RISCV_priv_spec_revision        |       12 | uleb128        | *Deprecated*, indicates the revision version of the privileged specification.
 | Tag_RISCV_atomic_abi                |       14 | uleb128        | Indicates which version of the atomics ABI is being used.
 | Tag_RISCV_x3_reg_usage              |       16 | uleb128        | Indicates the usage definition of the X3 register.
+| Tag_RISCV_mop_and_hint_encoding     |       18 | uleb128        | Indicates the reinterpretation of MOP and hint instruction encodings.
```
jaidTw, line 1275: I suggest to use the uppercase **HINT** in this document to emphasize we are refer to HINT instructions, which is the convention the spec uses

riscv-elf.adoc
```diff
@@ -1437,6 +1438,34 @@ value. After the merge, the resulting value will be the non-zero one.
 
 --
 
+===== Tag_RISCV_mop_and_hint_encoding, 18, uleb128=value
+
+Tag_RISCV_mop_and_hint_encoding indicates the reinterpretation of MOP (Maybe Operation) 
+and hint instruction encodings within this object. The value is a bitmap where each bit
```
jaidTw, line 1444: same here

riscv-elf.adoc
```diff
@@ -1437,6 +1438,34 @@ value. After the merge, the resulting value will be the non-zero one.
 
 --
 
+===== Tag_RISCV_mop_and_hint_encoding, 18, uleb128=value
+
```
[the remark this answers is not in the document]
kito-cheng: The information is expected to be used for assembler and disassembler, we didn't expect OS or loader to load this, the general design for the build attribute/GNU property is:
- build attribute are hold those info which only used for toolchain and that's strip-able
- GNU property is hold those info may used by dynamic linker or OS

One major reason is we didn't have LOAD attribute for build attribute section, which means it's harder to access by dynamic linker or OS.

However I think you raise a valid concern here is we should add more description/NOTE for that :)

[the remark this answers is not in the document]
xypron: Can we add the usage description here, e.g.

"For correct decoding by disassemblers hints are provided describing how maybe instructions and the like have been used."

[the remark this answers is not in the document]
kito-cheng: Thanks for suggestion, I've try to integrate into the description :)

kito-cheng:
Change: 
- Added note that disassemblers rely on provided hints to correctly decode Maybe instructions and related encodings.
- Update wording

kito-cheng:
@mylai-mtk lpad is hint according the spec I think?

https://github.com/riscv/riscv-isa-manual/blob/main/src/rv32.adoc#hint-instructions

mylai-mtk:
@kito-cheng 

> lpad is hint according the spec I think?

If LPAD is a HINT, then I'm confused. LPAD surely modifies architecturally visible state (the ELP bit in CSRs), and implementation that implements it are not allowed to ignore it in that implementation must check reg x7 == uimm20 or raise an exception. I would argue that LPAD does not really qualify as a HINT insn, even though it uses a HINT insn encoding point.

It's a loop-hole in the ISA description about HINTs. I think the problem is: HINT encoding points are not necessarily used as HINT insns (HINT insns do not change any architecturally visible state, except for advancing the pc and any applicable performance counters, and could be ignored by implementation). These HINT encoding points could be used to encode extension insns that have architecturally observable effects and thus could not be ignored if the implementation claims to implement the extension. 

But anyway, at least the `auipc x0, uimm20` encoding point is a listed HINT encoding point. The name "mop_and_hint_encoding" fits Zicfilp in this sense, so I'll drop my opinion about name changing.

topperc:
I will soon land a patch to LLVM to always allow Zicbop pretch instructions which also use hint encodings. https://github.com/llvm/llvm-project/pull/191027

PR #475, opened 2025-08-12
TLSDESC Relax Relocation Clarification

lenary:
Unlike all other relaxation sequences, TLSDESC relaxations only require a R_RISCV_RELAX on their first relocation. Clarify this fact in the text.


kito-cheng (commented):
Could you split the part of dropping symbol from R_RISCV_RELAX to a separate PR, I believe this part could be fast approved since it's apparently document fix, and for the TLSDESC part I would like moving forward after discussion within psABI meeting (although I think this should be no controversial)

kito-cheng:
For TLS relocations it’s indeed a bit special, unlike other reloc relaxations, they’ll still get relaxed even if linker relaxation is turned off. But I still prefer tagging any instructions that might be changed or removed with `R_RISCV_RELAX` to keep the design consistent.

kito-cheng (approved):
Give my explicit LGTM here

PR #476, opened 2025-08-12
RELAX doesn't take the symbol in examples

lenary:
This clarifies that only the base relocation will reference the relocated symbol, and the R_RISCV_RELAX relocation does not point at the same symbol.


jrtc27 (approved):
Thanks

PR #479, opened 2025-10-17
[cc] Document [u]intmax_t

lenary:
These have some effect on the ABI of code that use them, and need to be kept stable long-term, so we should document the sizes they are.

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -577,6 +577,8 @@ alignments (based on the ILP32 convention):
 | int                  |  4            |  4                |
 | long                 |  4            |  4                |
 | long long            |  8            |  8                |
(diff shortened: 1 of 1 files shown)


lenary:
Here is more background on how this leaks into the ABI:
- https://thephd.dev/intmax_t-hell-c++-c
- https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2425.pdf

I have actually opened https://github.com/riscv-non-isa/riscv-c-api-doc/pull/129 to not document the other `int_fastN_t` and `int_leastN_t` in the ABI as they currently don't always match between toolchains, and I don't think we want to make them ABI stable. Sorry for not cross-linking to that change originally.

PR #483, opened 2025-11-25
Reserve relocations for the RISC-V Y base ISA.

resistor:
Per the defined criteria for reserving relocation numbers ahead of standardization, we would like to reserve 10 relocation numbers for the RISCV Y base ISA. Our current best estimate (based on existing CHERI ABI implementations and outstanding Y specification work items) is that the Y ISA will require 7 relocations. We prefer to round that up to 10 relocations for this reservation to reduce the risk of conflicts, with the intention of releasing any reserved relocation numbers that are unused once the Y base ABI is standardized.


kito-cheng (approved):
LGTM, thanks for reserving 66 for landing pad, I probably should create another PR after this PR merge to update the text a little bit.

For #392...that's just inactive for long time so it's OK to skip that IMO :P

Will merge once @jrtc27 also approve since this is kinda doc change only not require any PoC or approve from any open source place holder.

PR #485, opened 2026-01-29
Add Relax Feature String for fine-grained linker relaxation control

kito-cheng:
This proposal serves as an alternative to PR #393, introducing Relax Feature String to specify constraints on linker relaxation for individual instructions.

When both `.option arch, +c` and `.option arch, -c` code regions exist in the same object file, the linker cannot properly handle this mixed situation because `EF_RISCV_RVC` is an object-level flag.

The Zcmt extension attempts to relax `call`/`tail`/`jal` instructions into `cm.jalt`/`cm.jt`. In baremetal toolchain multilib scenarios, using ISA strings leads to only two choices:

1. Add a Zcmt-specific multilib (increases toolchain release size, and the multilib is identical to non-Zcmt version except ISA string)
2. Violate the arch string semantics

When Zicfilp is enabled, LPAD instructions must be 4-byte aligned. Linker relaxation may convert call to `c.jal` or `cm.jalt`, causing LPAD misalignment. The current `.option exact` solution completely disables relaxation, which is overly conservative.

1. **Long strings**: ISA strings like RVA23 profile can be considerable
2. **Poor extensibility**: Cannot express "allow RVC but prohibit Zcmt"
3. **Zcmt semantic issues**: Cannot elegantly handle Zcmt opt-out
4. **Linker performance**: Complex ISA string parsing

Extend `R_RISCV_RELAX` to optionally reference a dummy symbol with prefix `$r` containing comma-separated feature flags.

- **Name**: starts with `$r` prefix (e.g., `$rnorvc`)
- **Binding**: `STB_LOCAL`
- **Type**: `STT_NOTYPE`
- **Size**: 0

Features MUST be sorted in lexicographical order. This simplifies linker parsing overhead.

| Feature | Description |
|---------|-------------|
| `norvc` | Prohibit relaxation into compressed instructions (C/Zc*). Implies `nozcmt` since `cm.jt`/`cm.jalt` are also compressed. | | `nozcmt` | Prohibit relaxation into Zcmt table jump instructions. Useful for performance-critical regions since Zcmt requires table lookup. |

Old linkers ignore the symbol referenced by `R_RISCV_RELAX`, so they use default relaxation behavior.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -528,7 +528,7 @@ Description:: Additional information about the relocation
                                             <| S + A - P
 .2+| 46-50   .2+| *Reserved*    .2+| -       |                   .2+| Reserved for future standard use
                                             <|
-.2+| 51      .2+| RELAX         .2+| Static  |                   .2+| Instruction can be relaxed, paired with a normal relocation at the same address
+.2+| 51      .2+| RELAX         .2+| Static  |                   .2+| Instruction can be relaxed, paired with a normal relocation at the same address. This relocation might point to an optional dummy symbol containing a Relax Feature String (see <<relax-feature-string>>), which specifies constraints on the relaxation process.
                                             <|
 .2+| 52      .2+| SUB6          .2+| Static  | _word6_           .2+| Local label subtraction
                                             <| V - S - A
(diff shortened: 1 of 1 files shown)


lenary:
The problem with using ISA strings, as I understand it, is that the libraries may not have been compiled with `zcmt` in the arch string, and we probably don't want to recompile libraries just to include `zcmt` in the arch string for each `R_RISCV_RELAX`. This is more fun when we realise that if we do recompile the libraries, `zcmt` probably needs special handling in anything dealing with multilib matching.

The counterargument is "how can we be sure that a static library compiled without zcmt is compatible with zcmt everywhere it has a `R_RISCV_RELAX` on a jump/call?" and we... cannot? For instance in baremetal applications, there can be calls before the jump table is set up (the latter is a similar problem to disabling relaxation before initializing `gp`).

I do prefer ISA strings to this proposal.

For reference, ELD now has several supplemental flags for enabling or disabling groups of relaxations because of the architecture: `--no-relax-c` and `--{no-}relax-xqci`. Both end up being for groups of extensions, rather than one at a time.

jrtc27:
Yes, you have to honour the set of extensions used when compiling that specific code fragment, rather than use the merged set, because you never know when people are doing manual dynamic dispatch. Same as how we ended up with the ISA string in .riscv.attributes not being invalid if it has conflicting extensions between object files.

kito-cheng:
My main concern was how to toggle zcmt per region, but still disable only zcmt without disabling c.

If we group zcmt with general c (norvc region), then ISA strings plus a `--[no-]relax-zcmt` linker option can solve the problem.

After reviewing our internal zcmt cases, I found that cases needing "nozcmt but c" are very rare. So ISA strings is the better choice.

Although I don't strongly support this proposal anymore, I prefer to keep this PR open until the ISA string approach lands.

NOTE: Nelson and I discussed this offline and reached this conclusion. Since I'm going to reply, he said he's too lazy to reply himself.


PS: Sorry to call copilot too early, I am highly rely on those tool to fix my English issue, however I think I should do that on local first next time :P

kito-cheng:
Close this since I start to implement on LLVM side

PR #487, opened 2026-03-06
Update GitHub Actions to requested versions

rpsene:
This updates GitHub Actions workflow references to the requested versions.

Targets:
- `actions/checkout` -> `v6`
- `actions/upload-artifact` -> `v6`
- `softprops/action-gh-release` -> `v2`


jrtc27 (approved):
Sure, if there are no breaking changes we need to care about in our simple use here

kito-cheng:
Thanks :)

Issue #486, opened 2026-01-29
Linker Relaxation Marker Relocations


lenary:
RISC-V uses relocations to mark relaxable instructions and other byte ranges.

Conventionally, Linkers do not relax instructions that are not marked with a
relocation. This is why relocations are coupled to the width of the instruction
they are relocating.

Right now, RISC-V has two main relocations that mark relaxable byte ranges:
- `R_RISCV_RELAX` (paired with another relocation)
- `R_RISCV_ALIGN` (which notes a byte range of nops that can be shrunk)

However, I want to discuss a different relocation: `R_RISCV_TPREL_ADD`.

This is used in the TLS "Local Exec Sequence":

```
  lui  a5, %tprel_hi(symbol)           # R_RISCV_TPREL_HI20 (symbol), R_RISCV_RELAX
  add  a5, a5, tp,%tprel_add(symbol)   # R_RISCV_TPREL_ADD (symbol), R_RISCV_RELAX
  lw   t0, %tprel_lo(symbol)(a5)       # R_RISCV_TPREL_LO12_I (symbol), R_RISCV_RELAX
```

This relaxation has no effect on the `add` instruction it is applied to. It is
present so linkers can change the above sequence into the following:

```
  lw t0, <tp-offset-for-symbol>(tp)
```

How do we feel about adding more relocations like this to the ABI? (I'm not
proposing deprecating, removing, or changing `R_RISCV_TPREL_ADD`.)

Here is one reason we might want to define a new marker relocation, in a similar
way: Assembler Relaxation.

If a user writes `beq a0, a1, symbol`, and `symbol` is not defined in the same
section, then this is converted to a long branch sequence:

```
  bne a0, a1, .+4
  jal x0, symbol  # R_RISCV_JAL (symbol), R_RISCV_RELAX
```

Once the assembler makes this change, the linker cannot undo it even if it
discovers that `symbol` is within range of the original `beq`. Why not? Because
there is no relocation on the `bne`, and we don't relax things without a
relocation.

One way we could change this is to define a `R_RISCV_LONG_BRANCH` relocation,
which the assembler would add to the `bne`, and define a linker relaxation which
can undo the long branch sequence back to `beq a0, a1, symbol`.

- Do we think this kind of use of relocations is a good idea?
- Do we think long branches would be worth investigating further?

Why am I asking?

I am thinking of defining similar marker relocations in the QC psABI Extensions,
[...]

PR #489, opened 2026-03-19
Add new relocation type for global array accesses with non-constant indices.

LiqinWeng:
Relocation Type Includes: R_RISCV_REGREL_ADD/R_RISCV_REGREL_LO12_I/R_RISCV_REGREL_LO12_S

The RISC-V architecture addresses non-constant subscript elements using lui + addi + add + ld/st. The low-address calculation process of addi instruction is offloaded th the ld/st, thereby eliminating low-address calculation and reducing addressing instructions.

The scenario is as follows: same as sh1add/sh2add/sh3add
```
lui     vr1, %hi(sym)
addi    vr1, vr1, %lo(sym)
add     vr2, vrx, vr1
lbu     vr3, off(vr2)
```
 After Compiler Transformation
```
lui     vr1, %hi(sym+off)
add     vr2, vrx, vr1, %regrel_add(sym+off)
lbu     vr3, %regrel_lo(sym+off)(vr2)
```
After Linker Relaxation:
```
add vr2, vrx, gp
lbu vr3, <gp-relative-offset>(vr2)
```
Need update the ld/st's offset and add/sh1add/sh2add/sh3add's src2 be replace by gp, so need add new relocation deal with this scenario.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -588,17 +588,17 @@ Description:: Additional information about the relocation
                                             <|
 .2+| 66      .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
                                             <|
-.2+| 67      .2+| REGREL_LO12_I .2+| Static  | _I-Type_          .2+| Low 12 bits of 32-bit Global Array Offset, `%regrel_lo(symbol)`
+.2+| 67-76   .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for Y base ISA
                                             <|
-.2+| 68      .2+| REGREL_LO12_S .2+| Static  | _S-Type_          .2+| Low 12 bits of 32-bit Global Array Offset, `%regrel_lo(symbol)`
+.2+| 77      .2+| REGREL_LO12_I .2+| Static  | _I-Type_          .2+| Low 12 bits of 32-bit Global Array Offset, `%regrel_lo(symbol)`
                                             <|
-.2+| 69      .2+| REGREL_ADD    .2+| Static  |                   .2+| Global Array usage, `%regrel_add(symbol)`
+.2+| 78      .2+| REGREL_LO12_S .2+| Static  | _S-Type_          .2+| Low 12 bits of 32-bit Global Array Offset, `%regrel_lo(symbol)`
                                             <|
-.2+| 70      .2+| REGREL_SHXADD .2+| Static  |                   .2+| Global Array usage, `%regrel_shxadd(symbol)`
+.2+| 79      .2+| REGREL_ADD    .2+| Static  |                   .2+| Global Array usage, `%regrel_add(symbol)`
                                             <|
-.2+| 71-76   .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for Y base ISA
+.2+| 80      .2+| REGREL_SHXADD .2+| Static  |                   .2+| Global Array usage, `%regrel_shxadd(symbol)`
                                             <|
-.2+| 77-190  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
+.2+| 81-190  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
                                             <|
 .2+| 191     .2+| VENDOR        .2+| Static  |                   .2+| Paired with a vendor-specific relocation and must be placed immediately before it, indicates which vendor owns the relocation.
                                             <|
@@ -1898,7 +1898,8 @@ Relaxation result:
   instruction sequence if possible.
 
   Condition:: Global-pointer relaxation requires that Tag_RISCV_x3_reg_usage
-  must be 0 or 1, and offset between global-pointer and symbol is within +-2KiB
+  must be 0 or 1, and offset between global-pointer and symbol is within +-2KiB.
+  `R_RISCV_REGREL_SHXADD` the Zba feature is required.
 
   Relaxation::
   - Instruction associated  with `R_RISCV_HI20` can be removed.
@@ -1910,7 +1911,7 @@ Relaxation result:
   Example::
 +
 --
-Relaxation candidate:
+Relaxation candidate 1:
 [,asm]
 ----
     lui tX, 0       # R_RISCV_HI20 (symbol), R_RISCV_RELAX
@@ -1921,7 +1922,22 @@ Relaxation result:
 [,asm]
 ----
     add tY, tY, gp
-    lw t1, <address-of-symbol>(tY)
+    lw t1, <gp-offset-for-symbol>(tY)
+----
++
+--
+Relaxation candidate 2: same as sh2add/sh3add
+[,asm]
+----
+    lui tX, 0          # R_RISCV_HI20 (symbol), R_RISCV_RELAX
+    sh1add tY, tY, tX  # R_RISCV_REGREL_SHXADD (symbol), R_RISCV_RELAX
+    lw t1, 0(tY)       # R_RISCV_REGREL_LO12_I (symbol), R_RISCV_RELAX
+----
+Relaxation result:
+[,asm]
+----
+    sh1add tY, tY, gp
+    lw t1, <gp-offset-for-symbol>(tY)
 ----
 --
 
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -588,7 +588,15 @@ Description:: Additional information about the relocation
                                             <|
 .2+| 66      .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
                                             <|
-.2+| 67-76   .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for Y base ISA
+.2+| 67      .2+| REGREL_LO12_I .2+| Static  | _I-Type_          .2+| Low 12 bits of 32-bit Global Array Offset, `%regrel_lo(symbol)`
+                                            <|
+.2+| 68      .2+| REGREL_LO12_S .2+| Static  | _S-Type_          .2+| Low 12 bits of 32-bit Global Array Offset, `%regrel_lo(symbol)`
+                                            <|
+.2+| 69      .2+| REGREL_ADD    .2+| Static  |                   .2+| Global Array usage, `%regrel_add(symbol)`
+                                            <|
+.2+| 70      .2+| REGREL_SHXADD .2+| Static  |                   .2+| Global Array usage, `%regrel_shxadd(symbol)`
+                                            <|
+.2+| 71-76   .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for Y base ISA
                                             <|
 .2+| 77-190  .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for future standard use
                                             <|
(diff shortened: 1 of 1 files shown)


LiqinWeng:
cc @lenary

riscv-elf.adoc
```diff
@@ -588,7 +588,15 @@ Description:: Additional information about the relocation
...
-.2+| 67-76   .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for Y base ISA
+.2+| 67      .2+| REGREL_LO12_I .2+| Static  | _I-Type_          .2+| Low 12 bits of 32-bit Global Array Offset, `%regrel_lo(symbol)`
+                                            <|
+.2+| 68      .2+| REGREL_LO12_S .2+| Static  | _S-Type_          .2+| Low 12 bits of 32-bit Global Array Offset, `%regrel_lo(symbol)`
+                                            <|
+.2+| 69      .2+| REGREL_ADD    .2+| Static  |                   .2+| Global Array usage, `%regrel_add(symbol)`
+                                            <|
+.2+| 70      .2+| REGREL_SHXADD .2+| Static  |                   .2+| Global Array usage, `%regrel_shxadd(symbol)`
+                                            <|
+.2+| 71-76   .2+| *Reserved*                          .2+| -       |                   .2+| Reserved for Y base ISA
```
lenary: 67-76 are reserved for a specific use - these relocations do not relate to Y base ISA. Please use values from 77 onwards.

riscv-elf.adoc
```diff
@@ -1880,6 +1888,43 @@ Relaxation result:
...
+----
+    lui tX, 0       # R_RISCV_HI20 (symbol), R_RISCV_RELAX
+    add tY, tY, tX  # R_RISCV_REGREL_ADD (symbol), R_RISCV_RELAX
+    lw t1, 0(tY)    # R_RISCV_REGREL_LO12_I (symbol), R_RISCV_RELAX
+----
+Relaxation result:
+[,asm]
+----
+    add tY, tY, gp
+    lw t1, <address-of-symbol>(tY)
```
lenary:
```suggestion
    lw t1, <gp-offset-for-symbol>(tY)
```

riscv-elf.adoc
```diff
@@ -1880,6 +1888,43 @@ Relaxation result:
 ----
 --
 
+==== Global Array Offset Relaxation
+
+  Target Relocation:: R_RISCV_HI20, R_RISCV_REGREL_LO12_I, R_RISCV_REGREL_LO12_S,
+  R_RISCV_REGREL_ADD, R_RISCV_REGREL_SHXADD
```
lenary: Please split off the `R_RISCV_REGREL_SHXADD` relaxation into a separate description, or add a shxadd example below (either way, an example is needed, as that relaxation is more complex).

lenary (commented):
Initial comments, to facilitate reviews.

These seem reasonable improvements to me, if we're continuing down the route of adding more relaxations and are willing to do [marker relocations](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/486).

I think it is reasonable that the `REGREL_*` relocations take the global symbol they relate to, but this does prevent them being implicitly relax (as we want to use the symbol in R_RISCV_RELAX for an arch string), which will cause a blow-up in the static relocation tables. That's the price of more linker relaxations, I feel.

riscv-elf.adoc
LiqinWeng: Could you take a look : https://github.com/llvm/llvm-project/issues/185586#issue-4049740735.I'd like to know if adding the handling logic for the inst scenarios below introduces any potential , as: add.uw/sh1add.uw/sh2add.uw/sh3add.uw

riscv-elf.adoc
LiqinWeng: done~

riscv-elf.adoc
LiqinWeng: done~

LiqinWeng:
cc @topperc @MaskRay

kito-cheng:
Hi @LiqinWeng, thanks for this proposal!

There is few review comment:

I think `REGREL_ADD` and `REGREL_SHXADD` can be merged into one relocation. Both `add` and `shXadd` just rewrite rs2 to gp. The linker can check the instruction encoding to know whether it is `add` or `shXadd` if needed, so there is no need to use different relocation types for them.

Also, I suggest adding a set of `PCREL_REGREL_*` relocations for medany code model support. The current `REGREL_*` only works with medlow (using `lui`). With `PCREL_REGREL_*`, medany (using `auipc`) can also benefit from this indexed array access relaxation. I was think we could reuse the same relocations for both models — the difference is just whether the relocation points to a symbol or a local label — but that would make the linker implementation more complex. Using separate relocation types is simpler and makes the semantics clearer.

jrtc27:
REGREL is a bit of a strange name. I imagine you've done it because of TPREL, but that's a bit different. With TPREL for TLS, TP is a pointer, and you materialise the offset of the TLS symbol within the static TLS block in a register that you add to TP (and since that's an offset not an address, you're always using lui+addi, never auipc+addi). With your proposed REGREL, the GPR in question isn't a pointer, it's some additional offset, and it's the lui+addi that's the pointer (and hence why auipc+addi would be needed for medany/PIE/PIC). This naming therefore seems backwards.

lenary:
> I think `REGREL_ADD` and `REGREL_SHXADD` can be merged into one relocation. Both `add` and `shXadd` just rewrite rs2 to gp. The linker can check the instruction encoding to know whether it is `add` or `shXadd` if needed, so there is no need to use different relocation types for them.

One question for the case of `add` is whether we also want to support rewriting `rs1` instead (we cannot for the shxadd). Maybe this is something we derive from the encoding though.

kito-cheng:
> > I think `REGREL_ADD` and `REGREL_SHXADD` can be merged into one relocation. Both `add` and `shXadd` just rewrite rs2 to gp. The linker can check the instruction encoding to know whether it is `add` or `shXadd` if needed, so there is no need to use different relocation types for them.
> 
> One question for the case of `add` is whether we also want to support rewriting `rs1` instead (we cannot for the shxadd). Maybe this is something we derive from the encoding though.

I'm inclined to require that `REGREL_ADD` always rewrites `rs2` for `add`. Since `add` is commutative, the compiler can choose the operand order without cost, so constraining the `lui` result to `rs2` has zero cost on the compiler side. The alternative would be either introducing two relocation types (one for `rs1`, one for `rs2`), or having the linker figure out which operand to rewrite — which would require pairing with `R_RISCV_HI20` or some register tracking analysis that we don't currently do.

kito-cheng:
> REGREL is a bit of a strange name. I imagine you've done it because of TPREL, but that's a bit different. With TPREL for TLS, TP is a pointer, and you materialise the offset of the TLS symbol within the static TLS block in a register that you add to TP (and since that's an offset not an address, you're always using lui+addi, never auipc+addi). With your proposed REGREL, the GPR in question isn't a pointer, it's some additional offset, and it's the lui+addi that's the pointer (and hence why auipc+addi would be needed for medany/PIE/PIC). This naming therefore seems backwards.

Yeah, I agree the naming is a bit confusing. I didn't have a good idea when I first replied, but after thinking more, how about `R_RISCV_BASE_IDX_ADD` and  `R_RISCV_BASE_IDX_LO12_[I|S]`? This better describes what the relocation actually is - a base + index addressing pattern.

I considered `GPREL` at first, but it would be even more confusing since the pattern has nothing to do with `GP` when relaxation doesn't happen.

lenary:
> I considered GPREL at first, but it would be even more confusing since the pattern has nothing to do with GP when relaxation doesn't happen.

Yes, the original LLVM proposal is GPREL but I asked for it not to be that.

lenary:
Looking at the sequences again, one thing this proposal needs is a separate relocation for `c.addi` - our relocation implicitly encodes the instruction size, so we would need a different relocation for the 16-bit instructions vs the 32-bit ones.

The alternative is to say that the compiler has to emit a 32-bit add, and that it's up to the linker to compress the add to a `c.add`.

There is no standard 16-bit encoding of shXadd, but Xqci can turn shXadd into `qc.c.muliadd`.

LiqinWeng:
> Also, I suggest adding a set of `PCREL_REGREL_*` relocations for medany code model support. The current `REGREL_*` only works with medlow (using `lui`). With `PCREL_REGREL_*`, medany (using `auipc`) can also benefit from this indexed array access relaxation. I was think we could reuse the same relocations for both models — the difference is just whether the relocation points to a symbol or a local label — but that would make the linker implementation more complex. Using separate relocation types is simpler and makes the semantics clearer.

Yes, this functionality will be extended in the future to cover `aupic + addi + add + ld/st.

LiqinWeng:
> I think `REGREL_ADD` and `REGREL_SHXADD` can be merged into one relocation. Both `add` and `shXadd` just rewrite rs2 to gp. The linker can check the instruction encoding to know whether it is `add` or `shXadd` if needed, so there is no need to use different relocation types for them.

Yes, during the compilation phase, I have unified the implementations of `add` and `shxadd`, ensuring that only `src2` is replaced by `gp`.  LLVM implement: https://github.com/llvm/llvm-project/pull/185353

LiqinWeng:
> REGREL is a bit of a strange name. I imagine you've done it because of TPREL, but that's a bit different. With TPREL for TLS, TP is a pointer, and you materialise the offset of the TLS symbol within the static TLS block in a register that you add to TP (and since that's an offset not an address, you're always using lui+addi, never auipc+addi). With your proposed REGREL, the GPR in question isn't a pointer, it's some additional offset, and it's the lui+addi that's the pointer (and hence why auipc+addi would be needed for medany/PIE/PIC). This naming therefore seems backwards.

Do you have any ideas for name?

kito-cheng:
> Looking at the sequences again, one thing this proposal needs is a separate relocation for `c.addi` - our relocation implicitly encodes the instruction size, so we would need a different relocation for the 16-bit instructions vs the 32-bit ones.
> 
> The alternative is to say that the compiler has to emit a 32-bit add, and that it's up to the linker to compress the add to a `c.add`.
> 
> There is no standard 16-bit encoding of shXadd, but Xqci can turn shXadd into `qc.c.muliadd`.

I would prefer let compiler always emit 32 bits add, then relax later, otherwise the b is 2 new relocations, and one potential new relocation for `R_RISCV_TPREL_ADD`, and maybe one more `R_RISCV_GPREL_ADD` for future.

kito-cheng:
> Do you have any ideas for name?

@lenary @jrtc27 do you think `R_RISCV_BASE_IDX_ADD` and `R_RISCV_BASE_IDX_LO12_[I|S]` is better? or other suggestion?

kito-cheng:
> > Also, I suggest adding a set of `PCREL_REGREL_*` relocations for medany code model support. The current `REGREL_*` only works with medlow (using `lui`). With `PCREL_REGREL_*`, medany (using `auipc`) can also benefit from this indexed array access relaxation. I was think we could reuse the same relocations for both models — the difference is just whether the relocation points to a symbol or a local label — but that would make the linker implementation more complex. Using separate relocation types is simpler and makes the semantics clearer.
> 
> Yes, this functionality will be extended in the future to cover `aupic + addi + add + ld/st.

I would incline include that within this PR rather than separate one.

kito-cheng:
@LiqinWeng I've build<del>vibe</del> a PoC for LLVM and binutils, and the commit message is little bit mess, but it should work.

LLVM: https://github.com/kito-cheng/llvm-project/tree/pr185353-review
binutils: https://github.com/kito-cheng/riscv-binutils-gdb/tree/kitoc/base_idx_reloc
https://github.com/kito-cheng/riscv-gnu-toolchain/tree/base-idx-reloc

kito-cheng:
@LiqinWeng discussed with me offlist about the medany part, and she is happy to extend the spec to cover that part.

Also she has update the name to `R_RISCV_BASE_IDX_*` in https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/489/changes/47d7b9545debba3b09d8fc252a79966744266302

LiqinWeng:
@jrtc27 @lenary @asb pls review again. The implementation related to AUIPC will be handled by @kito-cheng

LiqinWeng:
ping^^

PR #490, opened 2026-03-19
Antora refactor

wmat:
Antora merged to add the requirements for Antora so that this spec appears on the docs.riscv.org library.

PR #491, opened 2026-03-19
Revert "Antora refactor"

jrtc27:
Reverts riscv-non-isa/riscv-elf-psabi-doc#490


jrtc27:
For full transparency / documentation: after discussion with @kito-cheng, we have force-pushed master back to how it was prior to #490:
```
Total 0 (delta 0), reused 0 (delta 0), pack-reused 0 (from 0)
To github.com:riscv-non-isa/riscv-elf-psabi-doc
 + edc8342...9d36878 9d36878 -> master (forced update)
```
This was to avoid having the .pdf binary artifact forever in the git history that everyone must fetch. We hope that few, if any, people have fetched from the repo between the merging of the PR and this force-push, and apologise to those that have for the inconvenience.

PR #492, opened 2026-03-25
Adds the Antora scaffolding

wmat:
This adds the requirements that Antora needs to be able to present the 1.0 version of the spec on docs.riscv.org.  This does not touch the master branch but rather is a branch of master at the v1.0 git sha.  This will never be merged to master but will exist along side it in order to include the historical release in html in the docs library.


wmat:
@jrtc27 given that this isn't a PR against Master, do you want to review it at all?

jrtc27:
It is currently a PR against master… but yes I would like to, though currently tied up with the CHERI Blossoms conference here in Cambridge so might be a few days

wmat:
Thanks, no rush.  These changes do indeed target master.  They allow future releases to be able to be presented via Antora on docs.riscv.org, so I would appreciate your review when you have time.

wmat:
Let's defer this for now as we don't need it yet.

PR #495, opened 2026-04-15
Give _BitInt(N) the same section level as Fixed-Length Vector.

topperc:
It was one level lower making it appear to be part of Fixed-Length Vector.


lenary (approved):
LGTM

kito-cheng (approved):
LGTM :)

Issue #488, opened 2026-03-13
TLSDESC resolver calling convention has vector state callee-saved


jrtc27:
Currently, TLSDESC states the following:

> * Any other registers are callee-saved. This includes any vector registers when the vector extension is supported.

This is problematic, as it means the run-time linker must save and restore the entire vector register state if it uses vectors internally, even though in most cases the caller does not need the state preserving. This is similarly awkward to lazy binding, where the approach taken was to avoid the problem entirely via DT/STO_RISCV_VARIANT_CC, like AArch64 does for its SVE state.

On AArch64 things are a bit of a mess. GCC will treat TLSDESC resolvers as clobbering the SVE state (see https://github.com/gcc-mirror/gcc/commit/80c13ac5ec1a584561a2c474a5d61e7881ef4f82), whereas LLVM currently does not (see https://github.com/llvm/llvm-project/issues/186500), but should, as glibc only preserves the Q registers. We should adopt the same behaviour, especially considering that V's register size is allowed to be even bigger than SVE's, and so the problem is even worse for us.

jrtc27:
Cc @MaskRay @ishitatsuyuki

LukeZhuang:
Adhemerval mentioned another solution of “multiple ifunc variants” [here](https://inbox.sourceware.org/libc-alpha/09A14C77-9EEA-4707-9E36-D3AAF8A98B24@gmail.com/T/#mf838de7b89da0f82e878fd0dcebb28b0ff458ed7), of which my understanding is creating multiple different `dl_tlsdesc_dynamic` versions (with/without V-ext etc.), and let the dynamic loader decide which one to use.

If so, I think it does not resolve the whole issue, because more and more modern high-performance RISCV machines with Linux/Android (TLS only happens and people cares about speed on them I guess) are using V-ext. It is more like hiding the issue instead of solving it.

So, looks like changing the ABI by assuming "All other registers EXCEPT VREGS are callee-saved" is the currently best solution IMO, given that saving-restoring them in `dl_tlsdesc_dynamic` is insanely expensive, and I guess we would also have issue with the frame size because of different VLEN. I can volunteer to push this

PR #496, opened 2026-04-17
Change the v-regs (and CSRs) to be caller-saved in the TLSDESC func

LukeZhuang:
This clarifies that the TLSDESC func is no longer responsible for save-restore the vector registers and vector CSRS, so it will be known to clobber them by default in the compiler.

### Background
Here is some background of TLSDESC: (maybe not written for most of the folks here :) )

TLSDESC is an alternative implementation of the traditional TLS global dynamic model. The key idea is leaving a call with an undetermined function pointer at link time. Then at the runtime, the dynamic loader looks at which category the symbol belongs to and choose the best suitable handling function for it. There are 4 cases:
1. `dl_tlsdesc_undefweak` [extremely simple]: when the symbol is undefweak, just return 0. (usually not happen)
2. `dl_tlsdesc_return` [simple]: when the variable is in the static TLS block, just return the offset to TP (almost the same as LE+IE, covers most of the cases in the practice, maybe ~90%)
3. `dl_tlsdesc_dynamic` fast path [little bit complex]: when the module generation count is not changes, which happens between `dlopen`s occur. Need to lookup the DTV. (most of the cases for the modules loaded by `dlopen`, maybe 7%)
4. **`dl_tlsdesc_dynamic` slow path** [complex]: usually when new `dlopen` happens. Fall back to calling `__tls_get_addr` (rare, maybe 3%)

### The issue
There is an issue with the dynamic slow path above. Currently we define that the TLSDESC call is using a custom calling convention: 

> Any other registers are callee-saved. This includes any vector registers when the vector extension is supported.

That means `dl_tlsdesc_dynamic` itself is responsible to save-restore all the registers it modifies. The problem is that it uses `call __tls_get_addr` who is with a standard calling convention. It pessimistically assuming that `__tls_get_addr` may clobber all non-preserved registers, including vector registers. And actually it does since it uses `malloc`! So **we have to manually save/restore all the vector registers in `dl_tlsdesc_dynamic`**, which causes wasting a lot of time on the things that may not be actually used (even if it uses v-regs, it does not use much). 

This issue is the main blocker that preventing [the GLIBC patch](https://inbox.sourceware.org/libc-alpha/09A14C77-9EEA-4707-9E36-D3AAF8A98B24@gmail.com/T/#u) from being checked-in. We cannot get that ready until we resolved this issue.

### The proposals
There are two possible solutions:
1. As Jessica described here: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/488, AArch64 changes SVE as clobbered and we can also adopt this behavior, let the compiler deal with v-regs. **I vote for it and created this pull request**. The regression should be tiny IMO.
2. Adhemerval mentioned about "multiple ifunc variants" in the GLIBC discussion, which my understanding is creating different function versions on different extensions and let the dynamic loader decide which one to use. I think this does not resolve the whole issue though.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1142,7 +1142,7 @@ The TLS descriptor `entry` is called with a special calling convention, specifie
 
 - `a0` is used to pass the argument and return value.
 - `t0` is used as the link register.
-- Any other registers are callee-saved. This includes any vector registers when the vector extension is supported.
+- Any other registers are callee-saved excepting the vector registers and vector CSRs when the vector extension is supported, which should be caller-saved.
 
 Example assembler load and store of a thread local variable `i` using the `%tlsdesc_hi`, `%tlsdesc_load_lo`, `%tlsdesc_add_lo` and `%tlsdesc_call`
 assembler functions. The emitted relocations are in the comments.


LukeZhuang:
@jrtc27 @MaskRay @rui314 @aswaterman @kito-cheng @topperc

kito-cheng:
This is an ABI breakage, but after thinking about the real impact, I am okay to take this change. Here is how I think about it:

- **Vector regs only need save/restore on the slow path, so maybe the cost is fine?**
  It may be fine on machines with small VLEN, but on machines with large VLEN the overhead is still very big. In the worst case (VLEN = 65536 bits = 8 KB), saving all 32 v-regs costs **256 KB**. Even with the largest real VLEN today, VLEN = 16 Kb ([Vitruvius](https://dl.acm.org/doi/10.1145/3575861)), it is still **64 KB** — much bigger than the default `SIGSTKSZ` (8 KB). In a signal handler (especially when the program uses `sigaltstack`) or a deep call chain, this can cause stack overflow.

- **Can we force `__tls_get_addr` to be built without any vector instructions?**
  In theory yes, but in practice it is almost impossible. It is not only about `malloc`; helpers like `memcpy` and `memset` are also often auto-vectorized by the compiler, or they directly call libc versions that use the V extension. To de-vectorize the whole slow path call graph is too costly.

- **Is the cost of caller-saved high?**
  - For programs that do not use vector, the cost is 0. But under the current design, even a program without vector still pays cost, as long as glibc is built with V extension.
  - For programs that use vector, TLS access usually does not appear in the middle of vectorized code, so the cost is also close to 0.

- **Compare to traditional GD model**
  - We did not give it a special calling convention, so it follows the standard ABI, which means vector regs are caller-saved. This PR makes the two models behave the same way.

- **AArch64 SVE already did this!**
  - So this change has a precedent, and in practice it did not cause problems.

- **Is it an ABI breakage?**
  - Strictly speaking, yes. But luckily (or unluckily) the glibc patch has not landed yet, so in theory no binary depends on the current vector callee-saved behavior.
  - Honestly, I personally treat content on the master branch as already standardized (psABI 2.0 is not ratified yet, but it is already on master), so this argument is partly me convincing myself. But since no affected implementation has shipped, I think this window is still acceptable for the change.


For the PR itself, my main ask is to also make vector CSRs explicit (e.g. "excepting the vector registers and vector CSRs"), to avoid ambiguity for the vsetvli insertion pass in the implementation. At the same time, I prefer using a general term like "vector CSRs" instead of listing each one, so that if a new vector CSR is added in the future, we do not need to update this paragraph again.

Finally, there may be a matrix extension in the future, but it is still a draft, so I do not think we need to say anything about matrix regs now. If a matrix extension comes later, it should follow this caller-saved rule when its spec is shaped, so that we do not need to change it again after deployment.

jrtc27:
This is a duplicate of #488 …

jrtc27:
Well, that’s the issue and this is the PR, but it’s a bit sad to be splitting the discussion of the problem between the two

LukeZhuang:
> Well, that’s the issue and this is the PR, but it’s a bit sad to be splitting the discussion of the problem between the two

@jrtc27 Sorry Jessica, my initial thought is to push the TLSDESC GLIBC support, which is long path: resolves this issue in psABI, then make changes in LLVM/GCC, finally we can support in GLIBC then.
Maybe I'm too aggressive. If I caused any offense, I can close this PR immediately.

jrtc27:
No, all that's fine, I just felt it would have been better to have the discussion on the existing issue rather than fork a new thread in this PR. But that ship has now sailed.

(488 came out of me starting to implement TLSDESC for RISC-V in FreeBSD)

LukeZhuang:
@kito-cheng Thanks Kito for the opinions. The frame size and the matrix-extension are good points, we may also need to keep track with such extension in the future, if we make this change for vectors today.
Besides that, I also just think of musl/Android. What is the status of TLSDESC support on them? Do they need changes as well? (I guess that may not be so important since saving-restoring all registers should always be fine)

@jrtc27 Thanks for the idea, I should have asking for your permission explicitly first :) Would you mind if I add you to the co-author in the commit then?

jrtc27:
The contents of the PR is all your work so Co-authored-by isn't warranted

kito-cheng:
@LukeZhuang I am less concern about musl/Android: musl didn't have TLSDESC support yet, and RISC-V Android also not formal released yet, so we still have time to fix this.

LukeZhuang:
Updated with additional clarifications about vector CSRs

LukeZhuang:
I've thought of one question these days when trying to implement it on the compilers: some of the vregs (v1-v7, v24-v31) are already callee-saved, should we also exclude them from TLSDESC clobber list? That's equivalent to this question: "should TLSDESCCALL have the exact same vector calling convention as __tls_get_addr"? IMO, I guess we don't have to.

Maybe not so important because conservatively treating it as clobber ALL vregs is always the safe-most plan, it does not affect correctness at all. But I guess we'd better reach a consensus on this :)

aswaterman:
> some of the vregs are already callee-saved

This is only true in the vector calling convention variant.  In the standard calling convention, the vector registers are all caller-saved.

> Maybe not so important because conservatively treating it as clobber ALL vregs is always the safe-most plan

This seems like a reasonable choice to me; as pointed out in #488, the caller usually doesn't need to preserve any vregs anyway.

LukeZhuang:
> > some of the vregs are already callee-saved
> 
> This is only true in the vector calling convention variant. In the standard calling convention, the vector registers are all caller-saved.

Oh, I just read the pdf carefully and yes, that's my misunderstanding, thank you for correcting me!

LukeZhuang:
Hi @kito-cheng, I remember that you said we may want some opinion from Android guys in the last meeting, just curious may I ask is there any updates? Thanks!

LukeZhuang:
> Hi @kito-cheng, I remember that you said we may want some opinion from Android guys in the last meeting, just curious may I ask is there any updates? Thanks!

gentle ping :)

kito-cheng:
> Hi @kito-cheng, I remember that you said we may want some opinion from Android guys in the last meeting, just curious may I ask is there any updates? Thanks!

Confirmed from Elliott Hughes @ google:

1) No official riscv64 android and no official NDK ABI yet.
2) They are OK for this change

kito-cheng (approved):
LGTM :)

LukeZhuang:
> > Hi @kito-cheng, I remember that you said we may want some opinion from Android guys in the last meeting, just curious may I ask is there any updates? Thanks!
> 
> Confirmed from Elliott Hughes @ google:
> 
> 1. No official riscv64 android and no official NDK ABI yet.
> 2. They are OK for this change

@kito-cheng Thank you very much for confirming! Looks like I don't have way to check-in this change, could you please help me do that?

kito-cheng:
It's sort of ABI breakage change so we need @jrtc27's explicitly approve as well, also I would defer merge until then next psABI meeting (which is happen in next week) :)

jrtc27 (commented):
(So there's a record here, verbally agreed in this month's psABI meeting)

LukeZhuang:
Thanks @kito-cheng and @jrtc27! I want to volunteer to push the landing of it and have already prepared several toolchain patches (gcc/llvm/glibc/binutils), they will be sailed one-by-one soon :)

PR #498, opened 2026-04-23
Allocate an EF_RISCV_RVY e_flag for RVY

arichardson:
The toolchain should set this flag to indicate that the ELF file uses the RVY base ISA and an appropriate pure-capability ABI. Future pull requests will define the pure-capability ABIs for RVY as well as the associated calling conventions.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -355,6 +355,13 @@ below.
 
 NOTE: RV64ILP32* ABIs are experimental.
 
+[[EF_RISCV_RVY]]
+EF_RISCV_RVY (0x00000040)::: This bit is set when the binary targets the
+<<rvy-spec,RVY>> base ISA using a pure-capability ABI.
+
+NOTE: The pure-capability RVY calling conventions are not yet defined in this specification.
+
+
 Until such a time that the *Reserved* bits are allocated by future
 versions of this specification, they shall not be set by standard software.
 Non-standard extensions are free to use bits 24-31 for any purpose. This may
@@ -2144,3 +2151,6 @@ https://github.com/riscv/riscv-code-size-reduction
 
 * [[[rv-toolchain-conventions]]] "RISC-V Toolchain Conventions"
 https://github.com/riscv-non-isa/riscv-toolchain-conventions
+
+* [[[rvy-spec]]] "RISC-V Specification for CHERI Extensions"
+https://riscv.github.io/riscv-cheri/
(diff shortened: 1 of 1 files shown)


riscv-elf.adoc
```diff
@@ -355,7 +355,15 @@ below.
...
-Until such a time that the *Reserved* bits (0x00ffffe0) are allocated by future
+// NOTE: Previous versions of CHERI used the custom value 0x00010000 (EF_RISCV_CHERIABI)
+[[EF_RISCV_RVY]]
+EF_RISCV_RVY (0x00000040)::: This bit is set when the binary targets the
+<<rvy-spec,RVY>> base ISA using a pure capability ABI.
+
+NOTE: The pure-capability RVY calling conventions are not yet defined in this specification.
+
+
+Until such a time that the *Reserved* bits (0x00ffff80) are allocated by future
```
arichardson: The reserved mask is changing by two bits since I believe the update was omitted for EF_RISCV_RV64ILP32

lenary: Can we just remove the value here so we don't need to update it in two places?

arichardson: Yes that's probably a better solution. I wonder if the reason that it's spelled out here is that the top byte is not part of the mask?

lenary: The top byte is not `*Reserved*`, it's `*Non-standard extensions*`, so this should be conveyed by the table named "Layout of e_flags" above.

riscv-elf.adoc
```diff
@@ -355,7 +355,15 @@ below.
 
 NOTE: RV64ILP32* ABIs are experimental.
 
-Until such a time that the *Reserved* bits (0x00ffffe0) are allocated by future
+// NOTE: Previous versions of CHERI used the custom value 0x00010000 (EF_RISCV_CHERIABI)
```
kito-cheng: I would suggest to drop this 
```suggestion
```

riscv-elf.adoc
lenary: Oh kito just did it https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/500

kito-cheng (approved):
LGTM, and I think that should be necessary for RVY, and defer the merge until next psABI meeting in case there is different opinion

arichardson:
> LGTM, and I think that should be necessary for RVY, and defer the merge until next psABI meeting in case there is different opinion

Thanks for the review! I've rebased to fix the merge conflict.

PR #499, opened 2026-04-23
DRAFT: Initial description of the pure-capability calling conventions

arichardson:
In general, RVY ABIs try to follow the base RVI as closely as possible.
One difference are variadic functions and ones with on-stack arguments,
where the calling convention is slightly different to ensure that the
argument block can be tightly bounded to catch overflows.


arichardson:
CC: @jrtc27 @davidchisnall @resistor @chrehrhardt

riscv-cc.adoc
```diff
@@ -43,20 +43,20 @@ pointer relaxation is disabled in the toolchain.
...
-frame, thereby starting the linked list. By convention, the lowest XLEN value
-shall point to the previous frame, while the next XLEN value shall be the
-return address. The end of the frame record chain is indicated by the address
-zero appearing as the next link in the chain.
+Code that uses a frame pointer will construct a linked list of stack frames, where each frame links to its caller using a "frame record".
+A frame record consists of two pointer-sized values on the stack;
+the return address and the link to the next frame record.
+The frame pointer register will point to the innermost frame, thereby starting the linked list.
+By convention, the lowest pointer value shall point to the previous frame, while the next pointer value shall be the return address.
+The end of the frame record chain is indicated by a `NULL` pointer appearing as the next link in the chain.
```
kito-cheng: Could you keep this less change as possible, I figured out it only change XLEN to pointer size and `address zero` to `NULL`, and this part could be a separated patch since this also fix for `RV64ILP32` which is not RUY ABI only.

riscv-cc.adoc
```diff
@@ -43,20 +43,20 @@ pointer relaxation is disabled in the toolchain.
...
 
 After the prologue, the frame pointer register will point to the Canonical
 Frame Address or CFA, which is the stack pointer value on entry to the current
 procedure. The previous frame pointer and return address pair will reside just
-prior to the current stack address held in `fp`. This puts the return address
-at `fp - XLEN/8`, and the previous frame pointer at `fp - 2 * XLEN/8`.
+prior to the current stack address held in `fp`.
+// NOTE: Using sizeof(void *) instead of XLEN/8 here is an ABI change for RV64ILP32 since fp is now 4 instead of 8 byte aligned.
+// If this is a problem rather than a bugfix then we can use max(XLEN/8,sizeof(void *)) instead.
+This puts the return address at `fp - sizeof(void *)` and the previous frame pointer at `fp - 2 * sizeof(void *)`.
```
kito-cheng: We could define a PTRSZ or PTRLEN in `Terms and Abbreviations` and the use here and above I think.

riscv-cc.adoc
```diff
@@ -97,6 +97,14 @@ The Floating-Point Control and Status Register (fcsr) must have thread storage
 duration in accordance with C11 section 7.6 "Floating-point environment
 <fenv.h>".
 
+=== Capability Register Convention
+
+Since the RVY capability registers extend the integer register file, the same conventions and names are used.
+The only difference is that they are `YLEN=2*XLEN` bits wide and must be aligned to `YLEN` bits as well.
```
kito-cheng: Need to define YLEN in `Terms and Abbreviations` at introduction.adoc

riscv-cc.adoc
```diff
@@ -97,6 +97,14 @@ The Floating-Point Control and Status Register (fcsr) must have thread storage
 duration in accordance with C11 section 7.6 "Floating-point environment
 <fenv.h>".
 
+=== Capability Register Convention
```
kito-cheng: All other place use `Pure-Capability`?
```suggestion
=== Pure-Capability Register Convention
```

riscv-cc.adoc
```diff
@@ -675,6 +683,72 @@ The ILP32E calling convention is not compatible with ISAs that have registers
...
+These calling conventions are used when the base ISA is RVY and a pure-capability ABI is in use.
+In these ABIs, all pointers (including the stack pointer, frame pointer, and return address) are represented as capabilities (using the extended GPRs).
+
+The register names and usage remain the same as the <<integer-cc,integer calling convention>>, but the registers `x0`-`x31` (and their ABI aliases) now refer to the `YLEN`-wide extended GPRs.
+Any integer operation (e.g., add/multiply/etc.) writing to the `x1`-`x31` registers sets the address bits and clears the metadata.
+
+The pure-capability ABIs closely follow the same calling conventions as the base integer and hardware floating-point calling conventions, but since the GPRs are now `YLEN` bits wide the following differences exist:
+
+* Variadic and on-stack arguments use a different register convention (see next subsection).
+* `2*XLEN` (i.e. `YLEN`) wide scalars are passed in a single argument register if they are of pointer type (including `(u)intptr_t`).
```
kito-cheng: Curious about cheri LLVM implementation, how it how to distinguish between `(u)int64_t` and `(u)intptr_t` on RV32? it just model as new type other than `i64`?

riscv-cc.adoc
```diff
@@ -855,6 +965,45 @@ alignments (based on the LP64 convention):
...
+alignment (based on the IL32PC64 convention):
++
+.C/{Cpp} type sizes and alignments for IL32PC64
+[cols="4,>2,>3"]
+[width=60%]
+|===
+| Type                   | Size (Bytes)    | Alignment (Bytes)
+| void *                 | 8               | 8
+| `intptr_t`/`uintptr_t` | 8               | 8
+| All other types        | See ILP32* ABIs | See LP64* ABIs
```
kito-cheng:
```suggestion
| All other types        | See ILP32* ABIs | See ILP32* ABIs
```

riscv-cc.adoc
```diff
@@ -675,6 +683,72 @@ The ILP32E calling convention is not compatible with ISAs that have registers
...
++
+NOTE: For non-pointer scalar types the maximum size passed in a single argument register is still `XLEN` rather than `YLEN`.
+This avoids the implementation difficulty of passing `int64_t` in a single register when `YLEN` is 64 (RV32).
+// It also reflects that the high bits of capability registers cannot be used for computation directly as they need to be copied to another register using `YHIR` first.
+// A future calling convention variant may allow for packing these types in a single register.
+
+* Aggregates up to `2*YLEN` bits are passed in `YLEN` argument register(s) (using the memory layout of the aggregate) if they are at least `YLEN`-bit aligned.
++
+NOTE: This rule is required to ensure that aggregates that could potentially be used to store capabilities, but do not declare capability members (such as `alignas(16) char bytes[16]`) are passed in capability registers instead of being split into two registers (which would invalidate the capability tag).
+// This does mean that `int64_t` and `struct { int64_t }` have different calling conventions but this behaviour matches the base integer calling convention.
```
kito-cheng, line 714: I incline to make it consistently, I mean scalar part also use whole YLEN to pass for GPRs and not limited to pointer only, but I guess here might be some story behind the design? could you share more here?

riscv-cc.adoc
```diff
@@ -748,8 +822,41 @@ RV64ILP32Q:: RV64ILP32 with hardware floating-point calling
 convention for ABI_FLEN=128 (i.e. <<ELFCLASS32,ELFCLASS32>> and
 <<EF_RISCV_FLOAT_ABI_QUAD,EF_RISCV_FLOAT_ABI_QUAD>>).
 
-The LP64* ABIs are only compatible with RV64* ISAs. The ILP32* are compatible
-with RV32* and RV64* ISAs.
+[[abi-il32pc64]]
+IL32PC64:: Integer and capability calling convention only, hardware floating-point calling convention is not used
```
kito-cheng: `pure-` prefix?
```suggestion
IL32PC64:: Integer and pure-capability calling convention only, hardware floating-point calling convention is not used
```

riscv-cc.adoc
```diff
@@ -675,6 +683,72 @@ The ILP32E calling convention is not compatible with ISAs that have registers
...
+NOTE: This rule is required to ensure that aggregates that could potentially be used to store capabilities, but do not declare capability members (such as `alignas(16) char bytes[16]`) are passed in capability registers instead of being split into two registers (which would invalidate the capability tag).
+// This does mean that `int64_t` and `struct { int64_t }` have different calling conventions but this behaviour matches the base integer calling convention.
+
+* For aggregates that can be passed in pairs of registers,   they are never split into one register part and one on-stack part.
+  Instead of splitting the entire aggregate is passed on the stack.
+* Aggregates larger than `2*YLEN` or with an alignment larger than the stack alignment are always passed by reference.
+* All other aggregates follow the base calling convention.
+* Unions follow the same rules as aggregates: if they are `YLEN`-bit aligned and less than or equal to `2*YLEN` bits in size,   they are passed in memory layout in `YLEN` registers.
+  Otherwise, the base calling convention rules apply.
+
```
kito-cheng, line 722: Explicitly repeat this again to prevent ambiguous on return value.

```suggestion

Values are returned in the same manner as a first named argument of the same
type would be passed.
```

riscv-cc.adoc
```diff
@@ -675,6 +683,72 @@ The ILP32E calling convention is not compatible with ISAs that have registers
...
+* Variadic and on-stack arguments use a different register convention (see next subsection).
+* `2*XLEN` (i.e. `YLEN`) wide scalars are passed in a single argument register if they are of pointer type (including `(u)intptr_t`).
++
+NOTE: For non-pointer scalar types the maximum size passed in a single argument register is still `XLEN` rather than `YLEN`.
+This avoids the implementation difficulty of passing `int64_t` in a single register when `YLEN` is 64 (RV32).
+// It also reflects that the high bits of capability registers cannot be used for computation directly as they need to be copied to another register using `YHIR` first.
+// A future calling convention variant may allow for packing these types in a single register.
+
+* Aggregates up to `2*YLEN` bits are passed in `YLEN` argument register(s) (using the memory layout of the aggregate) if they are at least `YLEN`-bit aligned.
```
kito-cheng, line 710: Adding few wording to say it will fallback to base calling convention rules if less than `YLEN`-bit aligned

riscv-cc.adoc
```diff
@@ -97,6 +97,14 @@ The Floating-Point Control and Status Register (fcsr) must have thread storage
 duration in accordance with C11 section 7.6 "Floating-point environment
 <fenv.h>".
 
+=== Capability Register Convention
+
+Since the RVY capability registers extend the integer register file, the same conventions and names are used.
+The only difference is that they are `YLEN=2*XLEN` bits wide and must be aligned to `YLEN` bits as well.
+
+IMPORTANT: Capability register usage in the ILP32* and LP64* ABIs is not specified yet, only IL32PC64* and L64PC128* ABIs have a well-defined calling convention for capability registers.
+
```
kito-cheng, line 107: Add explicitly description for x5 is used for on-stack-arg capability here

riscv-cc.adoc
```diff
@@ -675,6 +683,72 @@ The ILP32E calling convention is not compatible with ISAs that have registers
...
+Measurements on other architectures (AArch64) have shown that most variadic functions are simply wrappers around a function taking a `va_list`.
+In standard ABIs that pass variadic arguments in registers, the callee must spill these registers to the stack to create a contiguous `va_list` structure, which adds overhead and complicates the implementation.
+Passing all variadic arguments on the stack from the start simplifies these wrappers and avoids complex register-to-stack spilling logic.
+Using `YLEN`-sized slots for all arguments, including those smaller than `YLEN`, further avoids the need for complex logic to determine the offset of each argument.
+This approach has been used successfully in other CHERI-enabled architectures such as Morello and was also adopted by Apple in the variadic calling convention for AArch64 on iOS and macOS.
+
+2. A capability that covers all on-stack arguments on the stack is passed in the `x5` (`t0`) register.
+3. For variadic functions, this capability has its `Store` and `Execute` permissions cleared.
+4. If no on-stack arguments are passed to a variadic function, the `x5` register is set to null.
```
kito-cheng:
```suggestion
4. If no on-stack arguments are passed to a variadic function, the `x5` register is set to `NULL` capability (tagged invalid).
```

riscv-cc.adoc
```diff
@@ -675,6 +683,72 @@ The ILP32E calling convention is not compatible with ISAs that have registers
 that require load and store alignments of more than 32 bits. In particular, this
 calling convention must not be used with the D ISA extension.
 
+[[cheri-cc]]
```
kito-cheng: nit
```suggestion
[[pure-cap-cc]]
```

riscv-cc.adoc
```diff
@@ -981,6 +1130,17 @@ alignment used will be those of a pointer, and the fetched pointer will be used
 as the address of the actual argument. The `va_copy` macro is a single pointer
 copy and the `va_end` macro performs no operation.
 
+==== va_list, va_start, and va_arg in pure-capability ABIs
+NOTE: In the pure-capability ABIs, the handling of variadic arguments is different.
+All variadic arguments are passed on the stack, and a capability to them is passed in the `x5` (`t0`) register.
+See <<cheri-cc>> for details.
```
kito-cheng: nit
```suggestion
See <<pure-cap-cc>> for details.
```

riscv-cc.adoc
```diff
@@ -1054,3 +1214,7 @@ https://github.com/riscv/riscv-v-spec
 * [[[rvv-intrinsic-doc]]] "RISC-V Vector Extension Intrinsic Document"
 https://github.com/riscv-non-isa/rvv-intrinsic-doc
 
+* [[[riscv-cheri-spec]]] "RISC-V Specification for CHERI Extensions "
```
kito-cheng: nit
```suggestion
* [[[riscv-cheri-spec]]] "RISC-V Specification for CHERI Extensions"
```

kito-cheng (commented):
@arichardson thanks for the PR, I believe that's important piece for RVY spec, so I am happy to helping this moving forward :)

One major concern to me is we have set a rule for psABI: require both LLVM and GNU toolchain implementation PoC, I know there is mature LLVM implementation, however I am not aware about the GNU toolchain part especially for GCC? waiver is discussable, but I want to make sure my understanding is correct.

riscv-cc.adoc
arichardson: Yes, in LLVM IR it is modeled as `ptr addrspace(200)`.

riscv-cc.adoc
arichardson: I'll defer this one to @resistor, but there are at least two problems we want to avoid (primarily on RV32 systems, but similar for i128 on RV64). We don't want to pass i64 in the YLEN GPRs because:
a) if you pass i64/u64 in two registers you can directly perform computations on them using standard integer instructions, if you put the second half in the metadata of the register, the caller has to pack the two registers into one and the callee has to unpack it again.
b) It causes challenges for compiler implementation

resistor: On the implementation challenges front, we have implementation experience doing f64 in YLEN GPRs on CHERIoT. This causes modeling challenges for the LLVM backend, as you end up with a type that is legal (as in can be stored in a register), but on which almost no operations are legal. This was workable for an FP type, but required a lot of whackamoling of issues.

I expect doing the same for an integer type to be 
much more challenging, as LLVM bakes in a lot more assumptions of what operations should be legal on a legal integer type.

This is maybe not a disqualifying concern (it's ultimately software and can probably be worked around _somehow_), but in combination with the fact that we believe overall code quality will be better with split i64/i128, it seems clear to us that splitting is the correct choice.

riscv-cc.adoc
arichardson: Since we are not defining an ABI for the non-purecap uses of RVY that sounds good to me.

riscv-cc.adoc
arichardson:
```suggestion
4. If no on-stack arguments are passed to a variadic function, the `x5` register is set to a `NULL` capability (marked as invalid).
```

arichardson:
I've tried to address all the open comments. I pushed those as separate commits to make it more obvious what has changed since the initial review but I'll squash them all together once we are happy for this PR to move out of the draft state.

riscv-cc.adoc
```diff
@@ -97,6 +97,15 @@ The Floating-Point Control and Status Register (fcsr) must have thread storage
 duration in accordance with C11 section 7.6 "Floating-point environment
 <fenv.h>".
 
+=== Pure-Capability Register Convention
+
+Since the RVY capability registers extend the integer register file, the same conventions and names are used.
+The only difference is that they are `YLEN` bits wide and must be aligned to `YLEN` bits as well.
+Register usage follows the underlying ILP32*/LP64* ABIs with the only difference that `x5` is used for on-stack arguments (see <<pure-cap-on-stack-args>>).
```
kito-cheng, line 104: `x5` will clobber during lazy binding[1], so I think we have few options here:
1. Change to another register, also specify `_dl_runtime_resolve` should save/restore that.
2. Change the sequence of first entry of PLT, also specify `_dl_runtime_resolve` should save/restore that.
3. Disallow lazy binding for Pure-Capability Register Convention, and do nothing to first entry of PLT, but this may lose the some capability like glibc's `LD_AUDIT`.
 

[1] https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#procedure-linkage-table

riscv-cc.adoc
```diff
@@ -981,6 +1136,17 @@ alignment used will be those of a pointer, and the fetched pointer will be used
...
 
+==== va_list, va_start, and va_arg in pure-capability ABIs
+NOTE: In the pure-capability ABIs, the handling of variadic arguments is different.
+All variadic arguments are passed on the stack, and a capability to them is passed in the `x5` (`t0`) register.
+See <<pure-cap-cc>> for details.
+
+As in other ABIs, the `va_list` type is represented as `void *` (which is a capability now).
+The `va_start` macro initializes the `va_list` using the capability provided in `x5`.
+Since all variadic arguments are on the stack, the callee does not need to construct a varargs save area for registers.
+Since the `va_list` now includes bounds information, attempting to use a non-existant argument via `va_arg` will result in a trap.
```
kito-cheng:
```suggestion
Since the `va_list` now includes bounds information, attempting to use a non-existent argument via `va_arg` will result in a trap.
```

riscv-cc.adoc
```diff
@@ -675,6 +684,77 @@ The ILP32E calling convention is not compatible with ISAs that have registers
...
+// It also reflects that the high bits of capability registers cannot be used for computation directly as they need to be copied to another register using `YHIR` first.
+// A future calling convention variant may allow for packing these types in a single register.
+
+* Aggregates up to `2*YLEN` bits are passed in `YLEN` argument register(s) (using the memory layout of the aggregate) if they are at least `YLEN`-bit aligned.
+All less-aligned aggregates follow rules of the underlying base integer or float calling convention (ILP32*/LP64* ABIs).
++
+NOTE: This rule is required to ensure that aggregates that could potentially be used to store capabilities, but do not declare capability members (such as `alignas(16) char bytes[16]`) are passed in capability registers instead of being split into two registers (which would invalidate the capability tag).
+// This does mean that `int64_t` and `struct { int64_t }` have different calling conventions but this behaviour matches the base integer calling convention.
+
+* For aggregates that can be passed in pairs of registers,   they are never split into one register part and one on-stack part.
```
kito-cheng:
```suggestion
* For aggregates that can be passed in pairs of registers, they are never split into one register part and one on-stack part.
```

riscv-cc.adoc
```diff
@@ -675,6 +684,77 @@ The ILP32E calling convention is not compatible with ISAs that have registers
...
+  Instead of splitting the entire aggregate is passed on the stack.
+* Aggregates larger than `2*YLEN` or with an alignment larger than the stack alignment are always passed by reference.
+* All other aggregates follow the base calling convention.
+* Unions follow the same rules as aggregates: if they are `YLEN`-bit aligned and less than or equal to `2*YLEN` bits in size,   they are passed in memory layout in `YLEN` registers.
```
kito-cheng:
```suggestion
* Unions follow the same rules as aggregates: if they are `YLEN`-bit aligned and less than or equal to `2*YLEN` bits in size, they are passed in memory layout in `YLEN` registers.
```

introduction.adoc
kito-cheng, line 60: Need list IL32PC64, IL32PC64F, IL32PC64D, IL32PC64E, L64PC128, L64PC128F, L64PC128D, and L64PC128Q here

riscv-cc.adoc
```diff
@@ -675,6 +684,77 @@ The ILP32E calling convention is not compatible with ISAs that have registers
...
+Measurements on other architectures (AArch64) have shown that most variadic functions are simply wrappers around a function taking a `va_list`.
+In standard ABIs that pass variadic arguments in registers, the callee must spill these registers to the stack to create a contiguous `va_list` structure, which adds overhead and complicates the implementation.
+Passing all variadic arguments on the stack from the start simplifies these wrappers and avoids complex register-to-stack spilling logic.
+Using `YLEN`-sized slots for all arguments, including those smaller than `YLEN`, further avoids the need for complex logic to determine the offset of each argument.
+This approach has been used successfully in other CHERI-enabled architectures such as Morello and was also adopted by Apple in the variadic calling convention for AArch64 on iOS and macOS.
+
+2. A capability that covers all on-stack arguments on the stack is passed in the `x5` (`t0`) register.
+3. For variadic functions, this capability has its `Store` and `Execute` permissions cleared.
+4. If no on-stack arguments are passed to a variadic function, the `x5` register is set to a `NULL` capability (marked as invalid).
+
```
kito-cheng, line 745: I think we should add a new subsection, "Hardware Floating-Point Calling
Convention for Pure-Capability ABIs" here

The reason: the Hardware Floating-Point Calling Convention has special rules for 
small structs that hold only one or two floating-point values (the "flattening"
rules). With the new pure-capability rules for how aggregates are passed, it is
not clear how these two sets of rules work together.

These are the main open questions I see:

1. How is `struct { double d; void *p; }` passed? In `fa0` and `a0` (as a
   capability)? Is the tag kept?
2. The "one floating-point real and one integer" rule limits the integer part to
   `XLEN` bits. If that member is a pointer or capability, is the whole struct
   passed by the integer calling convention, or can the member go in a
   `YLEN`-wide GPR?
3. Do the rules for a struct with one or two floating-point reals change at all?
4. Is "a capability is never passed in a floating-point register" written down
   anywhere? I could not find it. 
5. When a struct matches both the new pure-capability aggregate rule
   (`YLEN`-aligned -> `YLEN` register) and the existing floating-point
   flattening rule, which one wins?

There is also a case that mixes these. On IL32PC64F, `ABI_FLEN` is 32, so a
`double` (64 bits) does not fit in a floating-point register. So how is
`struct { double d; void *p; }` passed on IL32PC64F? I think it should be passed
in two `YLEN` registers (`a0` + `a1`), not flattened, but the current text does
not make this clear.

Here is a draft for the new subsection. Please feel free to mix it in and change
anything. :)

```asciidoc
[[pure-cap-hardware-fp-cc]]
==== Hardware Floating-Point Calling Convention for Pure-Capability ABIs

The pure-capability hardware floating-point ABIs (IL32PC64F, IL32PC64D,
L64PC128F, L64PC128D, and L64PC128Q) use the same hardware floating-point
calling convention as the corresponding base ABI, with the differences described
in this subsection.
A capability is never passed in a floating-point register: floating-point
registers are `ABI_FLEN` bits wide and carry no capability tag, so they cannot
represent a capability. Capabilities and pointers are always passed in the
`YLEN`-wide general-purpose (capability) registers.

For a struct argument, the following rules are applied in order:

1. the hardware floating-point flattening rules in this subsection;
2. if no flattening rule applies, the <<pure-cap-cc,pure-capability aggregate
   rule>>;
3. if the pure-capability aggregate rule does not apply either, the integer
   calling convention.

Routing a non-flattened struct through the pure-capability aggregate rule
(step 2) before the integer calling convention (step 3) ensures that a struct
which is not flattened but contains a capability is passed in whole `YLEN`
registers, and is not split across `XLEN` units, preserving its tag.

The rules for a struct containing one or two floating-point reals (and the
corresponding complex-number rules) are unchanged from the base hardware
floating-point calling convention, as they do not involve capabilities.

A struct containing one floating-point real and one integer member, in either
order, is flattened and passed in a floating-point register and a
general-purpose register, provided the floating-point real is no more than
`ABI_FLEN` bits wide and the integer member is either:

* a non-pointer integer (or bitfield) no more than `XLEN` bits wide, passed in
  the low `XLEN` bits of an integer argument register (if the member is narrower
  than `XLEN` bits, the remaining bits are unspecified); or
* a pointer or `(u)intptr_t`, passed as a capability in a `YLEN`-wide argument
  register with its tag preserved.

If the floating-point real is wider than `ABI_FLEN` bits, or the integer member
is a non-pointer scalar wider than `XLEN` bits (for example `int64_t` on the
IL32PC64* ABIs), the struct is not flattened and is handled by step 2 or 3
above.

For example, given `struct { double d; void *p; }`:

* On an ABI where `ABI_FLEN` is at least 64 (e.g. IL32PC64D, L64PC128D), the
  `double` fits in a floating-point register, so the struct is flattened and
  passed in `fa0` and `a0` (as a capability), preserving the tag of `p`.
* On IL32PC64F, where `ABI_FLEN` is 32, the `double` exceeds `ABI_FLEN` and the
  struct is not flattened. It is `2*YLEN` bits and `YLEN`-aligned, so the
  pure-capability aggregate rule (step 2) passes it in two `YLEN` registers,
  `a0` (the `double`) and `a1` (the capability `p`, with its tag preserved).

NOTE: Routing non-flattened structs through the pure-capability aggregate rule
does not invalidate capability tags. Both the flattening rules and the aggregate
rule keep each capability whole in a single `YLEN` register and never split it
between registers or between a register and the stack, so its tag is preserved.

type would be passed.
```

riscv-cc.adoc
arichardson: @jrtc27 I don't quite remember why we used x5. Also you're probably the best person to suggest what we should do here 🙂

riscv-cc.adoc
arichardson: Thank you for this suggested wording! That definitely sounds like what we want: follow the base conventions as closely as possible except for pointer arguments which need to go in YLEN registers to preserve tags. 

I'll go over this in more detail later and will respond to the questions as well.

kito-cheng:
After the discussion with Jessica on the GOT/PLT topics last time, I realized I should also take a closer look at `riscv-elf.adoc` to see what needs to change there. 
Here is what I found that needs RVY-specific updates:

- GOT/PLT:
  - GOT entry should hold YLEN instead of pointer-size?
  - PLT code seq should use `ly` instead of `ld` or `lw`, and PLT0 entry should be revised as well.
  - Need more detail review later.

- TLS:
  - TLS descriptor structure needs adjustment: the `entry` field of `tls_descriptor` is a function pointer, which is a YLEN capability under the pure-capability ABIs, while `arg` is still an XLEN `unsigned long`.
  - Instruction substitutions in the TLS code sequences need to review, e.g. `add a5, a5, tp` may adjust to `yadd a5, tp, a5`, and all load might be replace with `ly`
  - Thread pointer: what capability properties should be set for `tp`?
  - Whole TLS might need to review more carefully.

- Relaxations: We may revise all linker relaxation, some of them may not work with RVY, one example is Global-Pointer relaxation: it's hard to determine a right permission ofr gp since that may cover different region, data, read only data, or even text, so disalow this might be right solution for RVY.

- Dynamic relocations: wordclass relocations such as `R_RISCV_JUMP_SLOT` and `R_RISCV_32/64` cannot produce a tag. RVY needs capability-filling dynamic relocations.

jrtc27:
GOT/PLT: Yes. The same is true of the PLT header, plus the fifth instruction is ADDIY not ADDI, and the shift is reduced by one (which makes it zero on RV64Y, and so we instead remove the instruction and add a nop at the end).

TLS: We have a whole new TLS model for CHERI that adds a level of indirection (think per-thread GOT for pointers to that thread's thread-local variables), for various reasons (adding bounds to them efficiently, plus permitting compartmentalisation).

Relaxation: Yes, some are no longer appropriate.

Dynamic relocations: R_RISCV_JUMP_SLOT is kept, but stores a capability instead of an integer address. We add new relative and absolute relocations for initialising capabilities where on non-CHERI you'd use R_RISCV_RELATIVE or R_RISCV_32/64.

riscv-cc.adoc
arichardson: I've used your suggested wording here since I believe it matches exactly what we want.

arichardson: But I'll defer to @jrtc27 for final confirmation.

PR #500, opened 2026-04-23
Drop the Reserved bits mask value text description for e_flags since …

kito-cheng:
…we have table for that

This prevent us to forgot update this in future, we already forgot this when adding RV64ILP32


lenary (approved):
LGTM

arichardson (approved):

PR #501, opened 2026-04-24
Use pointer size in the frame layout instead of XLEN

arichardson:
This is motivated by RVY where the frame layout uses YLEN (2*XLEN) values to store the return address/frame pointer.

It is also technically a change for the RV64ILP32 ABI, but my understanding is that this change is actually a bug fix, since we are storing pointers.

Split out from https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/499

### DIFF
--- a/introduction.adoc
+++ b/introduction.adoc
@@ -34,6 +34,7 @@ This specification uses the following terms and abbreviations:
 | NTBS              | Null-Terminated Byte String
 | XLEN              | The width of an integer register in bits
 | FLEN              | The width of a floating-point register in bits
+| PTRSZ             | The width of a pointer type in bits (`8*sizeof(void *)`)
 | Linker relaxation | A mechanism for optimizing programs at link-time, see <<Linker Relaxation>> for more detail.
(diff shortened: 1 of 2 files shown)


lenary (approved):
LGTM

introduction.adoc
```diff
@@ -34,6 +34,7 @@ This specification uses the following terms and abbreviations:
 | NTBS              | Null-Terminated Byte String
 | XLEN              | The width of an integer register in bits
 | FLEN              | The width of a floating-point register in bits
+| PTRSZ             | The width of a pointer type in bits (`8*sizeof(void *)`)
```
arichardson, line 37: Wasn't sure if I should use bytes or bits here. Right now the only use has to divide by 8 again, so maybe bytes is better?

kito-cheng: It's okay to keep it in bits to consistent with XLEN and FLEN :)

kito-cheng (approved):
LGTM :)

PR #503, opened 2026-06-09
[elf] Fix Section Heading

lenary:
In asciidoc, you cannot put a `====` header (of any number of `=`) inside an open (`--`) block. If you do, the text is passed through directly.

I noticed there was a problem with the "Policy for Merge Objects with Different File Headers" header, which was inside an open block. The block is associated with the `e_flags` list, but the valid `e_flags` values are ended shortly before, and this is a new section describing how to merge not just `e_flags` but also the other top-level elf data.

The fix is to end the `--` block before the heading, so this heading shows up as a correctly numbered heading in the PDF and the HTML Document.

From my scans, this was the only place this has occurred.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -355,6 +355,8 @@ below.
 
 NOTE: RV64ILP32* ABIs are experimental.
 
+--
+
 Until such a time that the *Reserved* bits are allocated by future
 versions of this specification, they shall not be set by standard software.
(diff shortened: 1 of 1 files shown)


kito-cheng:
Thanks for catching this :)

kito-cheng (approved):

lenary:
No worries. Quick to fix in the end :)

PR #504, opened 2026-06-29
Fix LMUL value for the eight-register fixed-length vector case

kito-cheng:


kito-cheng:
This is obviously a typo so merge ahead

PR #505, opened 2026-06-29
Define ABI for riscv_rvv_vector_bits fixed-length vector types

kito-cheng:
The RISC-V C API provides the `riscv_rvv_vector_bits(N)` type attribute, which turns a sizeless RVV data or mask type into a fixed-length type. Describe how such types are passed and returned.

A `riscv_rvv_vector_bits` type reuses the Standard Fixed-length Vector Calling Convention Variant. Its ABI_VLEN is derived from the sizeless RVV type the fixed-length type was created from, not from the total size of the fixed-length type:

- For a data type with vector register group multiplier LMUL, ABI_VLEN is N divided by LMUL, that is the size normalized to m1. An integer LMUL=n type occupies n vector argument registers; a fractional LMUL type occupies a single one.

- For a mask type vbool<RATIO>_t, ABI_VLEN is the number of mask elements multiplied by RATIO. The mask occupies a single vector argument register. The number of mask elements alone is not enough, because the same count can come from different RATIO and VLEN combinations, so ABI_VLEN follows the originating mask type.

With this ABI_VLEN the fixed-length type is passed and returned in the same way as the scalable RVV type it was derived from. The Fixed-Length Vector section is also updated to register these types and point to the calling convention variant for their ABI_VLEN.

### DIFF
--- a/riscv-cc.adoc
+++ b/riscv-cc.adoc
@@ -493,6 +493,30 @@ The ABI_VLEN is a parameter of this calling convention variant. It could be set
 by a command line option for the compiler or specified by a function
 attribute in the source code.
 
+For a fixed-length vector created by the `riscv_rvv_vector_bits(N)` type
+attribute (see <<Fixed-Length Vector>> and the RISC-V C API), the ABI_VLEN is
+derived from the sizeless RVV type that the fixed-length type was created from,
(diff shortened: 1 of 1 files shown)


kito-cheng:
The reason I created this as draft is we don't really has spec on RISC-V C API side yet...but most content I think is reasonable.

PR #506, opened 2026-07-01
Fix typo of vector function mangling rule examples

zengdage:
A  'v' character was mistakenly added in the examples for Name Mangling of Vector Functions. This patch removes the redundant 'v'.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -451,10 +451,10 @@ is mangled as
     vint32m8_t _ZGVr8N64l_foo (int i);  // LMUL=8
 
     // VLA Mode
-    vint32m1_t _ZGVr1Nxvl_foo (int i);  // LMUL=1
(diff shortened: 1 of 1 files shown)


riscv-elf.adoc
```diff
@@ -2760,4 +2760,4 @@ https://github.com/riscv/riscv-code-size-reduction
 https://github.com/riscv-non-isa/riscv-toolchain-conventions
 
 * [[[rvy-spec]]] "RISC-V Specification for CHERI Extensions"
-https://riscv.github.io/riscv-cheri/
\ No newline at end of file
+https://riscv.github.io/riscv-cheri/
```
kito-cheng: Keep this unchanged

kito-cheng (commented):
Otherwise LGTM

riscv-elf.adoc
zengdage: Done.

PR #507, opened 2026-07-01
Fix calculation for indirect low-part relocations

### LINKED ISSUE #502: Document `R_RISCV_PCREL_LO12_[I,S]` behavior when paired with `R_RISCV_GOT_HI20`
Current RISC-V toolchains generate `GOT_HI20` relocations paired with `PCREL_LO12[I,S]` when indexing a preemptible symbol in PIC way,

```
$ clang --target=riscv64-unknown-linux-musl test.c -c -fPIC
$ objdump -dr test.o

test.o: file format elf64-littleriscv

Disassembly of section .text:

0000000000000000 <foo>:
       0: 1141          addi    sp, sp, -0x10
       2: e406          sd      ra, 0x8(sp)
       4: e022          sd      s0, 0x0(sp)
       6: 0800          addi    s0, sp, 0x10

0000000000000008 <.Lpcrel_hi0>:
       8: 00000517      auipc   a0, 0x0
                0000000000000008:  R_RISCV_GOT_HI20     a
       c: 00053503      ld      a0, 0x0(a0)
                000000000000000c:  R_RISCV_PCREL_LO12_I .Lpcrel_hi0
                000000000000000c:  R_RISCV_RELAX        *ABS*
      10: 4108          lw      a0, 0x0(a0)
      12: 60a2          ld      ra, 0x8(sp)
      14: 6402          ld      s0, 0x0(sp)
      16: 0141          addi    sp, sp, 0x10
      18: 8082          ret
$ cat test.c 
int a;

int foo(void)
{
        return a;
}
```

However, `PCREL_LO12[I,S]` is defined to be calculated through `S - P`[1], which seems to assume they're paired with `PCREL_HI20`, and their behavior is only mentioned when they're paired in this manner[2]. Though current linkers process `GOT_HI20 + PCREL_LO12[I,S]` pairs in a similar way, I think we should have it documented clearly.

[1]: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/e47ab3bfcf8028b4d465e2eb6bf5f57eb88de496/riscv-elf.adoc?plain=1#L513-L515
[2]: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/e47ab3bfcf8028b4d465e2eb6bf5f57eb88de496/riscv-elf.adoc?plain=1#L856-L888

kito-cheng:
`R_RISCV_PCREL_LO12_I`, `R_RISCV_PCREL_LO12_S`, `R_RISCV_TLSDESC_LOAD_LO12`
and `R_RISCV_TLSDESC_ADD_LO12` do not compute a value on their own; each
fills the low 12 bits of the value computed by its paired high-part
relocation. Their `S - P` calculation was misleading, since `S`, `A` and
`P` all come from the high-part relocation, not the low-part label, and it
hid that the low-part may also pair with `R_RISCV_GOT_HI20`,
`R_RISCV_TLS_GOT_HI20` or `R_RISCV_TLS_GD_HI20`.

Replace `S - P` with "Same as the paired high-part relocation" in the
relocation table, and generalize the PC-relative text to refer to the
high-part relocation instead of `R_RISCV_PCREL_HI20` specifically.

Fix #502

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -1125,9 +1125,9 @@ Description:: Additional information about the relocation
 .2+| 23      .2+| PCREL_HI20    .2+| Static  | _U-Type_          .2+| High 20 bits of 32-bit PC-relative reference, `%pcrel_hi(symbol)`
                                             <| S + A - P
 .2+| 24      .2+| PCREL_LO12_I  .2+| Static  | _I-type_          .2+| Low 12 bits of a 32-bit PC-relative, `%pcrel_lo(address of %pcrel_hi)`, the addend must be 0
-                                            <| S - P
+                                            <| Same as the paired high-part relocation
 .2+| 25      .2+| PCREL_LO12_S  .2+| Static  | _S-Type_          .2+| Low 12 bits of a 32-bit PC-relative, `%pcrel_lo(address of %pcrel_hi)`, the addend must be 0
-                                            <| S - P
+                                            <| Same as the paired high-part relocation
 .2+| 26      .2+| HI20          .2+| Static  | _U-Type_          .2+| High 20 bits of 32-bit absolute address, `%hi(symbol)`
                                             <| S + A
 .2+| 27      .2+| LO12_I        .2+| Static  | _I-Type_          .2+| Low 12 bits of 32-bit absolute address, `%lo(symbol)`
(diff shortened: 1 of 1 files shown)


jrtc27:
I think we should just drop the formula, and maybe replace it with something like "Same as referenced relocation" (don't love that wording, others may have better ideas). Leaving it as S-P in the table is unhelpful and adding notes many sections later doesn't really make it better.

kito-cheng:
> I think we should just drop the formula, and maybe replace it with something like "Same as referenced relocation" (don't love that wording, others may have better ideas). Leaving it as S-P in the table is unhelpful and adding notes many sections later doesn't really make it better.

Yeah, let me simplify the change :)

riscv-elf.adoc
```diff
@@ -1116,18 +1116,18 @@ Description:: Additional information about the relocation
...
+                                            <| G + GOT + A - P
 .2+| 21      .2+| TLS_GOT_HI20  .2+| Static  | _U-Type_          .2+| High 20 bits of 32-bit PC-relative TLS IE GOT access, macro `la.tls.ie`
                                             <|
 .2+| 22      .2+| TLS_GD_HI20   .2+| Static  | _U-Type_          .2+| High 20 bits of 32-bit PC-relative TLS GD GOT reference, macro `la.tls.gd`
                                             <|
 .2+| 23      .2+| PCREL_HI20    .2+| Static  | _U-Type_          .2+| High 20 bits of 32-bit PC-relative reference, `%pcrel_hi(symbol)`
                                             <| S + A - P
 .2+| 24      .2+| PCREL_LO12_I  .2+| Static  | _I-type_          .2+| Low 12 bits of a 32-bit PC-relative, `%pcrel_lo(address of %pcrel_hi)`, the addend must be 0
-                                            <| S - P
+                                            <| Value of the paired high-part relocation
```
jrtc27: I did consider using "value" in my wording but that _might_ get confusing with the fact that the paired relocation's value is the high bits, whereas we want the low bits of the "entire value" i.e. calculation?

kito-cheng: Good point, the value might confusing, maybe just `Same as the paired high-part relocation`?

(I updated what that, that is better than `value` anyway)

lenary (approved):
This feels like an improvement to me. Thanks!

jrtc27 (approved):
Sure, this new wording seems ok to me, thanks

PR #509, opened 2026-07-10
Remove 'r' char from isa for openmp vector function mangling and fix some typos

zengdage:
Remove 'r' char from isa for vector function mangling. The char  `1, 2, 4, 8, h, i,  j` are not yet occupied in GCC.

```
    isa := <lmul>

    LMUL=1   --> '1'
    LMUL=2   --> '2'
    LMUL=4   --> '4'
    LMUL=8   --> '8'
    LMUL=1/2 --> 'h'
    LMUL=1/4 --> 'i'
    LMUL=1/8 --> 'j'
```

LMUL 1/2, 1/4 and 1/8 will likely not be implemented, though dedicated identifiers are assigned to them for specification completeness.

### DIFF
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -467,7 +467,7 @@ is mangled as
 
 [,c]
 ----
-    // The type of `i` is pointer type and 's' equals 1, so 'number' also equals 4 (i.e., 1 * sizeof(int *)).
+    // The type of `i` is pointer type and 's' equals 1, so 'number' also equals 4 (i.e., 1 * sizeof(*i)).
 
     // VLS Mode, VLEN=128
     vint32m1_t _ZGVr1N4l4_foo  (int *i);  // LMUL=1
@@ -509,7 +509,7 @@ is mangled as
 
 [,c]
 ----
-    // 's' equals 1, and 'number' equals 4 (i.e., 1 * sizeof(int *)).
+    // 's' equals 1, and 'number' equals 4 (i.e., 1 * sizeof(*i)).
 
     // VLS Mode, VLEN=128
     vint32m1_t _ZGVr1N4R4_foo  (int *i);  // LMUL=1
--- a/riscv-elf.adoc
+++ b/riscv-elf.adoc
@@ -278,7 +278,7 @@ Name mangling rule for vector functions.
 ----
 mangled-vector-name := "_ZGV" <isa> <mask> <len> <parameters> "_" <func-name>
 
-<isa> := "r" <lmul>
+<isa> := <lmul>
 
 <lmul> := "1" | "2" | "4" | "8" (LMUL in turn is 1, 2, 4, 8)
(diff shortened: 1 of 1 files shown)


zengdage:
@kito-cheng HI kito, I remove 'r' for openmp vector function mangling in here. Please help to review, thanks.

riscv-elf.adoc
```diff
@@ -278,10 +278,10 @@ Name mangling rule for vector functions.
 ----
 mangled-vector-name := "_ZGV" <isa> <mask> <len> <parameters> "_" <func-name>
 
-<isa> := "r" <lmul>
+<isa> := <lmul>
 
 <lmul> := "1" | "2" | "4" | "8" (LMUL in turn is 1, 2, 4, 8)
-        | "h" | "q" | "e"       (LMUL in turn is 1/2, 1/4, 1/8)
+        | "h" | "i" | "j"       (LMUL in turn is 1/2, 1/4, 1/8)
```
kito-cheng: I would prefer this keep unchanged, this is ISA specific, so we only need to care about the letter is not allocated yet in RISC-V is fine.

kito-cheng (commented):
Give some background around this PR, one of OpenMP GCC maintainer suggest us should use the mangled named as compact as possible, the human readability should not be first priority for that, so he suggest just using single letter is enough.

Although this is an ABI breakage change, it is still early enough to make a change.

riscv-elf.adoc
zengdage: The 'e' is already used for x86-64 AVX512F. Though GCC treats it as ISA-specific, the LLVM demangler identifies ISAs via such markers (see [https://github.com/llvm/llvm-project/blob/main/llvm/lib/IR/VFABIDemangler.cpp#L51](https://github.com/llvm/llvm-project/blob/main/llvm/lib/IR/VFABIDemangler.cpp#L51)) . Keeping 'e' may lead to conflicts and require some significant changes to the LLVM demangler.

kito-cheng: I think the change is still within an acceptable scope:
https://github.com/kito-cheng/llvm-project/commit/cf2dd786a1e91e08098a5fe794d530d023dddcab
  
And if our justification for changing the letters is that another arch already uses them... that actually sounds more like an argument for sticking with the two-letter scheme.

zengdage: Done. keep h,q,e unchanged.

kito-cheng (approved):
LGTM :)

kito-cheng:
We discussed that on psABI meeting, it's really unfortunately to changes that right after we merge, but it's early enough, so let moving forward to unblock the OpenMP support :)

zengdage:
> We discussed that on psABI meeting, it's really unfortunately to changes that right after we merge, but it's early enough, so let moving forward to unblock the OpenMP support :)

Okay, thank you for your patient help.

PR #512, opened 2026-08-13
[elf] Add Zijf Reloc + Relaxation

lenary:
Zijf "Jump Far" proposal: riscv/riscv-isa-manual#3269

This adds:
- A relocation for `JF`, for jumps within functions.
- A relaxation to turn `AUIPC`+`JALR ra` into `JFAL`, which can cover ±16MiB.

The current expectation is that compilers would:
- Emit `JF` similarly to how `JAL x0` is currently used (for unconditional jumps within functions, or in the long conditional pseudos)
- Not emit `JFAL`, as this is covered by the current call sequence, but have this as a relaxation result.

There are likely other relaxations I have missed (tail to `jf`, and `jf`/`jfal` to compressed jumps), but I think these are the two key items to start with.

The exact encoding of the `JFAL` offset is not yet set (there are some other suggestions in the Zijf isa manual thread), so this will need to stay draft until those questions are more settled.


pz9115:
A possible workaround for the `JFAL` alignment/convergence problem might be a fixed-size 6-byte relaxed form.

For an 8-byte `AUIPC+JALR`, the linker could choose between:

```asm
c.nop
jfal target
```

and

```asm
jfal target
c.nop
```

Both are 6 bytes, but put `JFAL` at different PC phases modulo 4. This could let the linker satisfy a stronger displacement-alignment requirement without changing the total relaxed size, so switching between the two forms would not move subsequent code.

That seems potentially different from ordinary 2-byte relaxation, where changing instruction sizes can move other `JFAL`s in and out of alignment.

Would this be worth exploring if the extra displacement bit from 4-byte alignment is considered valuable?

lenary:
> Would this be worth exploring if the extra displacement bit from 4-byte alignment is considered valuable?

I'm not sure the real value of this, given Zijf is to increase code density, if we need the `c.nop`, a better approach would be to just have a 48-bit instruction which would surely have enough space for a 16-byte aligned offset rather than a 32-byte aligned one, while covering the required range.

PR #514, opened 2026-09-23
[elf] Require zero addends for more GOT relocations

### LINKED ISSUE #510: TLS GOT Relocations + Addends
Following on from #188 

There are some more `HI20` relocations that are used to get the address of a specific GOT slot, where I'm not sure an addend is logical:
- `R_RISCV_TLS_GOT_HI20` used by TLS IE
- `R_RISCV_TLS_GD_HI20` used by TLS GD
- `R_RISCV_TLSDESC_HI20` used by TLSDESC

I think none of these should take into account the Addend, as I understand.

~~I am less sure about `R_RISCV_GOT32_PCREL` but it could be another one that must have a zero addend.~~ `R_RISCV_GOT32_PCREL` needs an addend (see comment below: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/510#issuecomment-5802785155)

lenary:
The updated relocations all may use a GOT slot. HI20 with a GOT slot does not make sense in the ABI, so the addend must be zero.

Fixes #510


aswaterman (approved):

kito-cheng (approved):
LGTM

kito-cheng:
And I believe this change is uncontroversial, so merge ahead.

PR #515, opened 2026-09-24
[elf] Update Zcmt Initialization Guidance

lenary:
- Use weak reference for `__jvt_base$` to avoid issues
- Update the assembly to use `lla` to ensure we only use the local image definition.


kito-cheng (approved):
LGTM, oh, and seems like we already did that way in newlib

